PR #13, opened 2024-08-07
feat: abstract messaging

brian-baugher:
This change reduces what we export to solely what is needed. Stops exporting TARGET_ORIGINS and TemplateMetadata. Stops exporting useMessage hooks. Adds a new hook that uses a useMessage hook but makes it really easy to just grab the document.
Also updates docs to reflect this and Editor props. New usage:
```
const Edit: () => JSX.Element = () => {
  const entityDocument = useDocumentProvider();

  return (
    <DocumentProvider value={entityDocument}>
      <Editor document={entityDocument} puckConfigs={puckConfigs} />
    </DocumentProvider>
  );
};
```


vite.config.ts
```diff
@@ -13,7 +13,7 @@ export default defineConfig(() => ({
       formats: ["es", "cjs"] as LibraryFormats[], // typescript is unhappy without this forced type definition
     },
     rollupOptions: {
-      external: ["react", "react-dom", "@measured/puck"],
+      external: ["react", "react-dom", "@measured/puck", "@yext/pages"],
```
mkilpatrick, line 16: Is yext/pages still necessary?

brian-baugher: rm'd

mkilpatrick (approved):

PR #16, opened 2024-08-09
feat: remove view page button from header

tsaikatie:
Screenshot 2024-08-09 at 10 09 12 AM

### DIFF
--- a/src/components/Editor.tsx
+++ b/src/components/Editor.tsx
@@ -1,7 +1,6 @@
 import { InternalEditor } from "../internal/components/InternalEditor.js";
 import React from "react";
 import "./index.css";
-import { DocumentProvider } from "@yext/pages/util";
 import { useEffect, useState, useCallback, useRef } from "react";
 import { LoadingScreen } from "../internal/puck/components/LoadingScreen.tsx";
 import { Toaster } from "../internal/puck/ui/Toaster.tsx";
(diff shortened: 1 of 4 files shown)


src/internal/puck/components/Header.tsx
```diff
@@ -31,18 +30,13 @@ import {
...
-};
-
 export const customHeader = (
   handleClearLocalChanges: () => void,
   handleHistoryChange: (histories: History[], index: number) => void,
   data: Data,
   handleSaveData: (data: Data) => Promise<void>,
   isDevMode: boolean
 ) => {
-  const entityDocument = useDocument<any>();
```
brian-baugher, line 45: Is useDocument used anywhere else? If not I think we can remove @yext/pages as a dependency. @mkilpatrick this might solve our double document provider situation. We can just use the DocumentProvider in the starter and not worry about passing the document to the Editor component from ve-lib.

brian-baugher: That would give us this usage case: 
```tsx
const Edit: () => JSX.Element = () => {
  const entityDocument = useDocumentProvider();

  return (
      <DocumentProvider value={entityDocument}>
        <Editor puckConfigs={puckConfigs} />
      </DocumentProvider>
  );
};
```

brian-baugher: I can do this in a separate PR @tsaikatie, kinda out of scope for this one, this looks good

brian-baugher (approved):

src/internal/puck/components/Header.tsx
brian-baugher: Tried this and it didn't work 😞 seems that getting rid of the double DocumentProvider won't be quite that simple

src/components/Editor.tsx
```diff
@@ -349,22 +348,20 @@ export const Editor = ({ document, puckConfigs }: EditorProps) => {
   return (
     <>
       {!isLoading ? (
-        <DocumentProvider value={document}>
```
mkilpatrick, line 352: Double check that everything works when removing this. I thought Brian was saying it was necessary.

asanehisa: Yeah it all works on my end. I tested by using the Dev Mode feature in YSS. Did Brian mention what broke? I don't see any problems while playing around

mkilpatrick (approved):

PR #18, opened 2024-08-09
fix:  update Puck after clearing local changes and fix undo state

ehallerYext:
[SUMO-6143](https://yexttest.atlassian.net/browse/SUMO-6143?atlOrigin=eyJpIjoiZGI5NTk1ZmUzOTE2NGZlNGEzYTdmN2UwZjc2N2EzYzEiLCJwIjoiaiJ9)

This fixes the Clear Local Changes button to:
1. Close the modal on confirmation
2. Reset the Puck data state without a page refresh

Additionally, this fixes the undo button to properly undo to the original state before there were any changes instead of getting stuck at the first change.

### DIFF
--- a/src/components/Editor.tsx
+++ b/src/components/Editor.tsx
@@ -1,13 +1,12 @@
 import { InternalEditor } from "../internal/components/InternalEditor.js";
-import React from "react";
 import "./index.css";
 import { DocumentProvider } from "@yext/pages/util";
-import { useEffect, useState, useCallback, useRef } from "react";
+import React, { useEffect, useState, useCallback, useRef } from "react";
 import { LoadingScreen } from "../internal/puck/components/LoadingScreen.tsx";
 import { Toaster } from "../internal/puck/ui/Toaster.tsx";
 import { getLocalStorageKey } from "../internal/utils/localStorageHelper.ts";
 import { TemplateMetadata } from "../internal/types/templateMetadata.ts";
-import { type History, type Data, type Config } from "@measured/puck";
+import { type History, type Config } from "@measured/puck";
 import {
   useReceiveMessage,
   useSendMessageToParent,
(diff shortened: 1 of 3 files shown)


colton-demetriou (approved):

brian-baugher (approved):

mkilpatrick:
> [SUMO-6143](https://yexttest.atlassian.net/browse/SUMO-6143?atlOrigin=eyJpIjoiZGI5NTk1ZmUzOTE2NGZlNGEzYTdmN2UwZjc2N2EzYzEiLCJwIjoiaiJ9)
> 
> This change is implemented to update the header after the user confirms that they would like to clear their local changes.
> 
> Tested by packing VE and using the locally packed version when running the VE starter in dev mode.
> 
> The second part of this item is to refresh the header after the user publishes their changes. It seems this may require changes to Puck. Per the [Puck Docs](https://puckeditor.com/v/canary/docs/api-reference/components/puck#data), the initial data passed to the component cannot be updated without remounting the component. Doing so causes some glitchy UI and is not a great experience. Another potential solution for this is to set a "initial history index" to mark as the new beginning of history after changes have been published. The issue with this is that every time a change is made in the puck editor, the header is being remounted, clearing the "initial history index" state. Would love to hear any other ideas of how we can get around this, but thought I would leave this here in case I can't get to it before I leave.

Setting initialHistory won't work as that only gets used on Puck mount. It might work by setting the puck history. If that doesn't seem to work then we'll need to have the page refresh.

asanehisa (approved):

src/components/Editor.tsx
```diff
@@ -329,7 +349,6 @@ export const Editor = ({ document, puckConfigs }: EditorProps) => {
   );
 
   const isLoading =
-    !puckData ||
```
jwartofsky-yext, line 332: Should this be replaced with visualConfigurationData?

jwartofsky-yext (commented):
lgtm assuming all is tested and working as expected

PR #25, opened 2024-08-26
fix: make quick find keyboard shortcut work

benlife5:
Adds event listeners for cmd+k and ctrl+k. When triggered, sends a message to the parent (storm).

Adds support for the shortcut when the header or sidebars are selected. Still does not work on the puck preview iframe because we cannot add a listener to that iframe from outside the puck package.

### DIFF
--- a/src/components/Editor.tsx
+++ b/src/components/Editor.tsx
@@ -355,6 +355,31 @@ export const Editor = ({ document, puckConfigs }: EditorProps) => {
     TARGET_ORIGINS
   );
 
+  const { sendToParent: openQuickFind } = useSendMessageToParent(
+    "openQuickFind",
+    TARGET_ORIGINS
+  );
+
+  const keyboardHandler = (event: KeyboardEvent) => {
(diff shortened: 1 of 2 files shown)


jwartofsky-yext (approved):
lgtm

src/components/Editor.tsx
```diff
@@ -355,6 +355,31 @@ export const Editor = ({ document, puckConfigs }: EditorProps) => {
...
 
+  const { sendToParent: openQuickFind } = useSendMessageToParent(
+    "openQuickFind",
+    TARGET_ORIGINS
+  );
+
+  const keyboardHandler = (event: KeyboardEvent) => {
+    if ((event.ctrlKey || event.metaKey) && event.key === "k") {
+      openQuickFind({
+        payload: {
```
mkilpatrick, line 366: If you use `useReceiveMessage` on the other side you don't even need to send a payload. None of the things you're sending are used on the other side are they?

mkilpatrick (approved):

briantstephan (approved):

jwartofsky-yext (approved):

PR #27, opened 2024-08-26
fix: redirecting on localhost

mkilpatrick:
We were auto redirecting to 404 on localhost which made local development not work in Yext dev.


colton-demetriou (approved):
did you verify that localhost is actually in the ancestor?

briantstephan (approved):

mkilpatrick:
> did you verify that localhost is actually in the ancestor?

yes, localhost works now

PR #28, opened 2024-08-26
feat: add debug logging with xYextDebug

mkilpatrick:
When the `xYextDebug` query param is added, debug logging is enabled which logs many function calls and data being passed around through useMessage. This will make it easier for us to debug issues without having to add debugging locally.

Screenshot 2024-08-26 at 6 19 24 PM


briantstephan (approved):
This looks really useful

mkilpatrick:
> This looks really useful

It'll also helpful for determining if we've introduced some effect that causes infinite requests.

jwartofsky-yext (approved):
🔥

asanehisa (approved):

PR #44, opened 2024-09-17
feat(util): add renderEntityFields function

briantstephan:
This adds a `renderEntityFields` function which adds a selector dropdown field to a component. The selector includes any mappable entity fields, based on a provided filter.

Manually tested by calling the function from the VE starter repo and saw the selector appear in the right panel and work as expected.

Demo:
https://github.com/user-attachments/assets/d27b98a7-434c-444a-b46d-4fa61946cbae

### DIFF
--- a/src/utils/getFilteredEntityFields.ts
+++ b/src/utils/getFilteredEntityFields.ts
@@ -7,16 +7,23 @@ type Only<T, U> = {
 };
 type Either<T, U> = Only<T, U> | Only<U, T>;
 
-type AllowList = {
-  allowList: string[];
+type ConfigFields<T extends Record<string, any>> = T["stream"]["fields"];
+
+type AllowList<T extends Record<string, any>> = {
+  allowList: ConfigFields<T>[number][];
 };
-type DisallowList = {
-  disallowList: string[];
+
+type DisallowList<T extends Record<string, any>> = {
+  disallowList: ConfigFields<T>[number][];
 };
-type EntityFieldTypesFilter = { types: EntityFieldTypes[] };
-export type RenderEntityFieldFilter = Either<
+
+type EntityFieldTypesFilter = {
+  types: EntityFieldTypes[];
+};
+
+export type RenderEntityFieldFilter<T extends Record<string, any>> = Either<
   EntityFieldTypesFilter,
-  Either<AllowList, DisallowList>
+  Either<AllowList<T>, DisallowList<T>>
 >;
 
 type EntityFieldTypes = "type.string" | "type.image" | `c_${string}`;
@@ -104,6 +111,7 @@ const getEntityTypeToFieldNames = (
       if (fieldNameToSchemas.length === 0) {
         return prev;
       }
+
       for (const fieldNameToSchema of fieldNameToSchemas) {
         const typeName = getTypeFromSchemaField(fieldNameToSchema.schemaField);
         if (!typeName) {
@@ -133,13 +141,14 @@ const getEntityTypeToFieldNames = (
   }, new Map<string, string[]>());
 };
 
-export const getFilteredEntityFields = (filter?: RenderEntityFieldFilter) => {
+export const getFilteredEntityFields = <T extends Record<string, any>>(
+  filter?: RenderEntityFieldFilter<T>
+) => {
   const entityFields = useEntityFields();
 
-  let filteredEntityFields = entityFields.stream.expression.fields
-    // filter to top level fields for now, though this is only based on the dot in the stream field, not the true schema
-    .filter((field) => field.children === undefined)
-    .filter((field) => !DEFAULT_DISALLOWED_ENTITY_FIELDS.includes(field.name));
+  let filteredEntityFields = entityFields.stream.expression.fields.filter(
+    (field) => !DEFAULT_DISALLOWED_ENTITY_FIELDS.includes(field.name)
+  );
 
   if (filter?.allowList) {
     filteredEntityFields = filteredEntityFields.filter((field) =>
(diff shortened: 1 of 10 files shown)


jwartofsky-yext (commented):
lgtm

Could you add a video?

src/utils/getFilteredEntityFields.ts
```diff
@@ -0,0 +1,172 @@
...
+  [P in keyof T]: T[P];
+} & {
+  [P in keyof U]?: never;
+};
+type Either<T, U> = Only<T, U> | Only<U, T>;
+
+type AllowList = {
+  allowList: string[];
+};
+type DisallowList = {
```
mkilpatrick, line 13: add empty lines between things throughout

src/utils/getFilteredEntityFields.ts
```diff
@@ -0,0 +1,172 @@
...
+      } else {
+        prev.get(typeName)?.push(field.definition.name);
+      }
+    }
+
+    return prev;
+  }, new Map<string, string[]>());
+};
+
+export const getFilteredEntityFields = (filter?: RenderEntityFieldFilter) => {
```
mkilpatrick, line 136: Move the filtering stuff into getEntityTypeToFieldNames directly and remove this function. It was my first pass version that getEntityTypeToFieldNames supersedes now and I didn't remove it from my PR.

src/utils/renderEntityFields.tsx
```diff
@@ -0,0 +1,56 @@
...
+  ObjectField,
+} from "@measured/puck";
+import {
+  getFilteredEntityFields,
+  RenderEntityFieldFilter,
+} from "./getFilteredEntityFields.ts";
+
+type RenderProps = Parameters<CustomField<any>["render"]>[0];
+
+type RenderEntityFields<T extends RenderProps> = {
```
mkilpatrick, line 15: Talk to Nik/Yen how to make autocomplete work for AllowFields. Change this to 
```
type RenderEntityFields<T extends RenderProps, U extends Record<string, any>> = {
  renderProps: RenderProps;
  fieldName: keyof T["value"];
  objectFields?: ObjectField<any>["objectFields"];
  filter?: RenderEntityFieldFilter<U>;
};
```

src/utils/renderEntityFields.tsx
```diff
@@ -0,0 +1,56 @@
...
+  renderProps: RenderProps;
+  fieldName: keyof T["value"];
+  objectFields?: ObjectField<any>["objectFields"];
+  filter?: RenderEntityFieldFilter;
+};
+
+export const renderEntityFields = <T extends RenderProps>({
```
mkilpatrick, line 22: Change to 
```
const renderEntityFields = <T extends RenderProps, U extends Record<string, any>>(
  props: RenderEntityFields<T, U>
) => {
```

src/utils/getFilteredEntityFields.ts
```diff
@@ -0,0 +1,172 @@
+import { useEntityFields, YextSchemaField } from "../hooks/index.ts";
+
+type Only<T, U> = {
+  [P in keyof T]: T[P];
+} & {
+  [P in keyof U]?: never;
+};
+type Either<T, U> = Only<T, U> | Only<U, T>;
+
+type AllowList = {
```
mkilpatrick, line 10: Change these types to
```
type ConfigFields<T extends Record<string, any>> = T["stream"]["fields"];

type AllowList<T extends Record<string, any>> = {
  allowList: ConfigFields<T>[number][];
};

type DisallowList<T extends Record<string, any>> = {
  disallowList: ConfigFields<T>[number][];
};

type EntityFieldTypesFilter = { types: EntityFieldTypes[] };

type RenderEntityFieldFilter<T extends Record<string, any>> = Either<
  EntityFieldTypesFilter,
  Either<AllowList<T>, DisallowList<T>>
>;
```

src/utils/getFilteredEntityFields.ts
briantstephan: Talked offline, we just want to remove the `.filter((field) => field.children === undefined)` filter on line 141.

src/utils/getFilteredEntityFields.ts
```diff
@@ -0,0 +1,181 @@
...
+  name: string;
+  schemaField: YextSchemaField;
+};
+
+/**
+ * Returns a list of {@link EntityFieldNameAndSchema} given a {@link YextSchemaField}.
+ * @param schemaField The schema field for which to get the field names.
+ * @returns Entity field names and their corresponding schema fields.
+ */
+const getEntityFieldNames = (
```
mkilpatrick, line 50: We should have this take in the `filter { subfields }` so that we can use it wherever needed. Then getEntityTypeToFieldNames can be simplified. 

I'm wondering if we should also simplify or merge getFilteredEntityFields with this.

getFilteredEntityFields does:

1. getEntityFields (top level only)
2. filters disallowed
3. filters by allow/disallow list
4. gets all fields by type
5. which calls getEntityFieldNames for each top level field
6. gets the schema for each of those fields, makes a map
7. filter to the types

I realize I wrote this but I'll admit it's a little confusing. After talking with Aaron the allow/deny list will remain top level, so steps 1-3 are fine as is. But how could we easily return filteredEntityFields that includes the subfields? Most of that work already lives in getEntityFieldNames just without filtering. 

Btw, filtering by type would work at both the top and sub-levels while allow/deny is top-level only.

See what you can come up with. Another option could be to have getEntityFieldNames return more info about each field that indicates whether it's top level or not so that we could call that for step 1 and filter for steps 2-3. But it also feels weird to call getEntityFieldNames again within getEntityTypeToFieldNames.

briantstephan: Updated so that it should handle subfields properly

mkilpatrick (approved):

mkilpatrick (approved):

PR #45, opened 2024-09-17
feat(util): add resolveProp function

jwartofsky-yext:
Tested locally using hardcoded document

The field value is returned if it exists

Otherwise the field name is returned


package.json
```diff
@@ -43,6 +43,7 @@
     "@radix-ui/react-switch": "^1.1.0",
     "@radix-ui/react-tooltip": "^1.1.2",
     "@yext/eslint-config": "^1.0.0",
+    "@yext/pages": "^1.2.0-beta.3",
```
mkilpatrick, line 46: Thinking about it more, I think we need document to be passed to resolveProp to avoid this lib from depending on pages.

src/utils/resolveProp.ts
```diff
@@ -0,0 +1,24 @@
...
+  let current = document;
+  for (let i = 0; i < steps.length; i++) {
+    if (current?.[steps[i]]) {
+      current = current[steps[i]];
+    } else {
+      missedStep = true;
+      break;
+    }
+  }
+  if (!!current && !missedStep) {
```
mkilpatrick, line 18: This would skip if current is the empty string which we don't want.

src/utils/resolveProp.ts
```diff
@@ -1,21 +1,17 @@
-import { useDocument } from "@yext/pages/util";
-
-export const resolveProp = <T>(entityField: string): T => {
-  const document = useDocument() as any;
-
+export const resolveProp = <T>(entityField: string, document: any): T => {
```
mkilpatrick, line 1: I'd make document first

src/utils/resolveProp.ts
```diff
@@ -1,21 +1,17 @@
...
   // check for the entity field in the document
   const steps: string[] = entityField.split(".");
   let missedStep = false;
   let current = document;
   for (let i = 0; i < steps.length; i++) {
-    if (current?.[steps[i]]) {
+    if (current?.[steps[i]] !== undefined) {
```
mkilpatrick, line 7: undefined check is unnecessary

test/resolveProp.test.ts
```diff
@@ -0,0 +1,37 @@
+import { test, assert } from "vitest";
+import { resolveProp } from "../src/utils/resolveProp.ts";
+
+test("resolve prop data found", async () => {
+  assert.equal(
+    resolveProp("location.address.city", {
+      location: {
```
mkilpatrick, line 7: The actual stream fields aren't `location.address` it's just `address`.

test/resolveProp.test.ts
```diff
@@ -0,0 +1,37 @@
+import { test, assert } from "vitest";
+import { resolveProp } from "../src/utils/resolveProp.ts";
+
+test("resolve prop data found", async () => {
```
mkilpatrick, line 4: resolveProp returns value when field found in document

test/resolveProp.test.ts
```diff
@@ -0,0 +1,37 @@
...
+        address: {
+          city: "potato land",
+        },
+      },
+    }),
+    "potato land"
+  );
+});
+
+test("resolve prop data empty", async () => {
```
mkilpatrick, line 17: resolveProp returns empty value when field found in document

test/resolveProp.test.ts
```diff
@@ -0,0 +1,37 @@
...
+        address: {
+          city: "",
+        },
+      },
+    }),
+    ""
+  );
+});
+
+test("resolve prop data not found", async () => {
```
mkilpatrick, line 30: resolveProp returns constant value when field not found in document

src/utils/resolveProp.ts
jwartofsky-yext: "" is falsey, so this will not pass if we want to allow empty strings.

mkilpatrick (approved):

PR #47, opened 2024-09-18
feat(util): use built-in fields

jwartofsky-yext:
This will use the new built-in fields for Visual Editor

Note that we are currently waiting on some changes to the built-in fields so this may change and still needs to be tested on a dev site.


jwartofsky-yext:
Going to recreate this since a bunch of stuff was renamed and moved around

PR #48, opened 2024-09-19
feat: add siteId

asanehisa:


jwartofsky-yext (approved):
lgtm

Will also want to pass this through YSS and VES once the built-in field is updated

I don't think this will hurt anything if merged before built-in fields are added, but you may want to test

asanehisa:
Oops actually decided we don't need this item.

PR #60, opened 2024-09-26
feat: hide static value field for complex entity types

benlife5:
Only display the Static Value field if it's type.string or type.phone. Probably more basic types to do this for in the future.


jwartofsky-yext (commented):
Why do we not want people to set a static value for image url, but we allow it for phone?

jwartofsky-yext (approved):

benlife5:
> Why do we not want people to set a static value for image url, but we allow it for phone?

I think we probably do want static image support, but ComplexImage has a number of fields so it's not as simple as just making the one string field available. It looked to me that Phone is just made available in the document as a string

mkilpatrick (approved):

PR #65, opened 2024-10-01
feat: toggle theme editing mode

benlife5:
https://github.com/user-attachments/assets/c3f306be-1688-4ed0-a665-9c02b1b420e5


src/internal/components/InternalEditor.tsx
```diff
@@ -6,19 +6,19 @@ import {
...
 import { TemplateMetadata } from "../types/templateMetadata.ts";
 import { EntityFieldProvider } from "../../components/EntityField.tsx";
 import { SaveState } from "../types/saveState.ts";
 import { DevLogger } from "../../utils/devLogger.ts";
+import ThemeSidebar from "../puck/components/ThemeSidebar.tsx";
 
 interface InternalEditorProps {
   puckConfig: Config;
   puckInitialHistory: InitialHistory | undefined;
-  isLoading: boolean;
```
benlife5, line 21: As far as I could tell, `isLoading`, `canEdit`, and `change()` weren't doing anything -- `canEdit` was only read when updating `canEdit`

src/internal/components/InternalEditor.tsx
```diff
@@ -97,41 +96,62 @@ export const InternalEditor = ({
...
               appState.data,
               puckInitialHistory?.histories[0].state.data, // used for clearing local changes - reset to first puck history
               handleSave,
-              templateMetadata.isDevMode && !templateMetadata.devOverride
+              templateMetadata.isDevMode && !templateMetadata.devOverride,
+              themeModeActive,
+              toggleThemeModeActive
             );
           },
+          actionBar: themeModeActive ? () => <></> : undefined,
```
benlife5, line 152: This isn't 100% done because I've got an open question in the Puck discord about how to fully hide these components

src/internal/components/InternalEditor.tsx
```diff
@@ -97,41 +96,62 @@ export const InternalEditor = ({
...
-    if (!canEdit) {
-      setCanEdit(true);
-      return;
-    }
+  const toggleThemeModeActive = () => {
+    setThemeModeActive((prev) => !prev);
   };
 
+  // set permissions on the component level to allow for dynamic updating
+  Object.values(puckConfig.components).forEach((component) => {
```
mkilpatrick, line 115: Does this need to be in a useEffect? If themeModeActive changes does this update properly? Or does refreshPermissions somehow just rerender this component?

src/internal/components/InternalEditor.tsx
```diff
@@ -97,41 +96,62 @@ export const InternalEditor = ({
   };
 
   const handleSave = async (data: Data) => {
+    if (themeModeActive) {
+      devLogger.logFunc("saveStyles");
+      // save styles
```
asanehisa, line 101: is this commenting indicating the styles are being saved?

src/internal/components/InternalEditor.tsx
```diff
@@ -97,41 +96,62 @@ export const InternalEditor = ({
...
+      return;
+    }
     devLogger.logFunc("saveVisualConfigData");
     saveVisualConfigData({
       payload: { visualConfigurationData: JSON.stringify(data) },
     });
   };
 
-  const change = async () => {
```
mkilpatrick, line 106: I think I had added this because there were issues without it. If thinks are working fine then great, just be on the lookout.

src/internal/components/InternalEditor.tsx
```diff
@@ -97,41 +96,62 @@ export const InternalEditor = ({
...
               handleSave,
-              templateMetadata.isDevMode && !templateMetadata.devOverride
+              templateMetadata.isDevMode && !templateMetadata.devOverride,
+              themeModeActive,
+              toggleThemeModeActive
             );
           },
+          actionBar: themeModeActive ? () => <></> : undefined,
+          components: themeModeActive ? () => <></> : undefined,
+          fields: themeModeActive ? () => <ThemeSidebar /> : undefined,
```
mkilpatrick, line 154: Does the override such that the fields aren't loaded at all now or does undefined mean "load it normally?"

jwartofsky-yext (commented):
Could you include a video?

src/internal/components/InternalEditor.tsx
benlife5: Meant it as a stub for where we'll hook up things later -- updated to make that clearer!

src/internal/components/InternalEditor.tsx
benlife5: I'll do this in another PR and check that it's ok

src/internal/components/InternalEditor.tsx
benlife5: undefined means load it normally

That's what my open question is discord is about -- is there some way to say don't load it at all?

mkilpatrick: I was just wondering. I would have assumed returning nothing renders nothing.

src/internal/components/InternalEditor.tsx
benlife5: This works because the themeModeActive state is defined in this component, so it rerenders when that updates.

If I put it in a useEffect with [themeModeActive], it unfortunately doesn't work because the header useEffect that calls refreshPermissions fires first.

But I'm not sure -- is it better to have this in a useEffect and have it set another state variable when it completes, which then triggers the header useEffect?

mkilpatrick: If it's working and not causing a bunch of rerenders that fire other things off needlessly then it's fine. You can use the xYextDebug=true param to see if we're now making more calls to things before/after.

benlife5: It didn't seem to be making more calls, but I realized it works if I moved it to the header useEffect so that seems better

jwartofsky-yext (approved):
lgtm

mkilpatrick (approved):

src/internal/components/InternalEditor.tsx
mkilpatrick: nice

asanehisa (approved):

PR #68, opened 2024-10-02
feat(components): add react-color color picker

jwartofsky-yext:
Verified that the color picker works and is interactable in the theme sidebar.

Tested in dev mode

ColorSelector can be added to props in components and will then appear in the editor. Colors can be edited and the state persists through reloading. Color is reset when clearing local data.

Example addition to Hero.tsx in `github.com/yextSolutions/pages-visual-editor-starter`
```
  myColor: ColorSelector({
    label: "My Color",
    type: "custom",
  }),
```

Which appears in the editor as:

image


src/internal/puck/components/ColorPicker.tsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import { SketchPicker, RGBColor } from "react-color";
+import { PresetColor } from "react-color/lib/components/sketch/Sketch";
+
+interface ColorPickerProps {
```
mkilpatrick, line 5: Almost always prefer types over interfaces

src/internal/puck/components/ColorPicker.tsx
```diff
@@ -0,0 +1,25 @@
...
+  disableAlpha: boolean;
+  color: RGBColor;
+  setColor: (color: RGBColor) => void;
+  presetColors?: PresetColor[] | undefined;
+}
+
+const ColorPicker = (props: ColorPickerProps) => {
```
mkilpatrick, line 12: Destructure the props

mkilpatrick, line 12: Idk if @benlife5 or @briantstephan have any thoughts on if we'll ultimately want this to work through Puck props or not. What do you guys think? After talking with Brian today it seems like we'll want the theme stuff to be a document field that ultimately gets saved to the site entity, so it would be separate from the Puck config. This is also good because theme is at the site level and not per layout which puckconfig is.

benlife5: I agree that it makes sense that the theme is saved/handled separately than the Puck config

That being said, I think the easiest thing to do for the UI right now is to convert the theme config into Puck fields so that we can use <AutoField> out of the box

for example, this https://github.com/yext/visual-editor/blob/6a6e956d5bb734aa5b80e8870dd5f808015cb8b9/src/internal/puck/components/ThemeSidebar.tsx#L77C1-L118C1
outputs this with no additional styling needed
Screenshot 2024-10-02 at 2 17 50 PM

jwartofsky-yext: For colors, I think we could use the color picker with custom field type:
https://puckeditor.com/docs/api-reference/fields/custom

I don't think there is anything to change with the color picker itself for us to use it in that way.

mkilpatrick: For it to work with puck fields we should expose the ColorPicker AS a 'custom' Puck field. That way it's easy to use. Given Ben's example:

```
case "color":
              field.objectFields[styleKey] = ColorPicker(label)
              break;
```
Also see how YextEntityFieldSelector works - https://github.com/YextSolutions/pages-visual-editor-starter/blob/mappable-fields/src/components/Hero.tsx#L597

benlife5:
I think we want an additional function like RenderYextEntityFieldSelector that returns 
`{type: 'custom', label: <label>, render: (props : RenderProps) => {}` with that render function hooked up to your ColorPicker component

src/internal/puck/components/ColorPicker.tsx
```diff
@@ -0,0 +1,26 @@
...
+  disableAlpha: boolean;
+  color: RGBColor;
+  setColor: (color: RGBColor) => void;
+  presetColors?: PresetColor[] | undefined;
+};
+
+const ColorPicker = (props: ColorPickerProps) => {
+  const { disableAlpha, presetColors, color, setColor } = props;
+  return (
+    <SketchPicker
```
mkilpatrick, line 15: I feel like we can just inline this in ColorSelector instead of creating a wrapper that doesn't really do anything.

jwartofsky-yext: Done

mkilpatrick (approved):

PR #70, opened 2024-10-02
feat: add toggle for "Use Entity Field" vs "Use Constant Value"

asanehisa:
toggle only available on valid types still.  "type.string" and "type.phone" 

now using radio

https://github.com/user-attachments/assets/f88fe607-daa6-41f6-8536-3a7abcfd0851

### DIFF
--- a/src/components/YextEntityFieldSelector.tsx
+++ b/src/components/YextEntityFieldSelector.tsx
@@ -127,18 +127,18 @@ const ToggleMode = ({
         <div className="ve-flex ve-items-center ve-space-x-2">
           <RadioGroupItem
             value="false"
-            id="r1"
+            id="false"
             onClick={() => toggleConstantValueEnabled(false)}
           />
-          <Label htmlFor="r1">Use Entity Value</Label>
+          <Label htmlFor="false">Use Entity Value</Label>
         </div>
         <div className="ve-flex ve-items-center ve-space-x-2">
           <RadioGroupItem
             value="true"
-            id="r2"
+            id="true"
             onClick={() => toggleConstantValueEnabled(true)}
           />
-          <Label htmlFor="r2">Use Constant Value</Label>
+          <Label htmlFor="true">Use Constant Value</Label>
         </div>
       </RadioGroup>
     </div>
--- a/src/components/YextEntityFieldSelector.tsx
+++ b/src/components/YextEntityFieldSelector.tsx
@@ -48,8 +48,9 @@ export const YextEntityFieldSelector = <T extends Record<string, any>>(
       const toggleEntityFieldMode = () => {
         setEntityFieldMode((prevMode) => !prevMode);
         onChange({
-          field: "",
-          constantValue: "",
+          field: value?.field ?? "",
+          constantValue: value?.constantValue ?? "",
+          constantValueEnabled: !value.constantValueEnabled,
         });
       };
 
@@ -73,7 +74,8 @@ export const YextEntityFieldSelector = <T extends Record<string, any>>(
                 onChange={(selectedEntityField) => {
                   onChange({
                     field: selectedEntityField,
-                    constantValue: "",
+                    constantValue: value?.constantValue ?? "",
+                    constantValueEnabled: false,
                   });
                 }}
                 value={value?.field}
@@ -87,8 +89,9 @@ export const YextEntityFieldSelector = <T extends Record<string, any>>(
               <AutoField
                 onChange={(newConstantValue) =>
                   onChange({
-                    field: "",
+                    field: value?.field ?? "",
                     constantValue: newConstantValue,
+                    constantValueEnabled: true,
                   })
                 }
                 value={value?.constantValue}
@@ -118,7 +121,7 @@ const ToggleMode = ({
   toggleEntityFieldMode: () => void;
 }) => {
   return (
-    <div className="ve-mt-2">
+    <div className="ve-mt-2 ve-w-full">
       <Button onClick={toggleEntityFieldMode} variant="secondary">
         {entityFieldMode ? (
           <>
--- a/src/components/YextEntityFieldSelector.tsx
+++ b/src/components/YextEntityFieldSelector.tsx
@@ -6,7 +6,6 @@ import {
   getFilteredEntityFields,
   RenderEntityFieldFilter,
 } from "../internal/utils/getFilteredEntityFields.ts";
-import { Lock } from "lucide-react";
 
 export type YextEntityField = {
   field: string;
@@ -123,10 +122,7 @@ const ToggleMode = ({
         {constantValueEnabled ? (
           <p>Use Entity Field</p>
         ) : (
-          <>
-            <Lock className="sm-icon" />
-            <p>Use Constant Value</p>
-          </>
+          <p>Use Constant Value</p>
         )}
       </Button>
     </div>
--- a/src/components/YextEntityFieldSelector.tsx
+++ b/src/components/YextEntityFieldSelector.tsx
@@ -127,18 +127,18 @@ const ToggleMode = ({
         <div className="ve-flex ve-items-center ve-space-x-2">
           <RadioGroupItem
             value="false"
-            id="false"
+            id="ve-use-entity-value"
             onClick={() => toggleConstantValueEnabled(false)}
           />
-          <Label htmlFor="false">Use Entity Value</Label>
+          <Label htmlFor="ve-use-entity-value">Use Entity Value</Label>
         </div>
         <div className="ve-flex ve-items-center ve-space-x-2">
           <RadioGroupItem
             value="true"
-            id="true"
+            id="ve-use-constant-value"
             onClick={() => toggleConstantValueEnabled(true)}
           />
-          <Label htmlFor="true">Use Constant Value</Label>
+          <Label htmlFor="ve-use-constant-value">Use Constant Value</Label>
         </div>
       </RadioGroup>
     </div>
--- a/src/components/YextEntityFieldSelector.tsx
+++ b/src/components/YextEntityFieldSelector.tsx
@@ -1,11 +1,13 @@
 import React from "react";
-import { AutoField, FieldLabel, Field, Button } from "@measured/puck";
+import { AutoField, FieldLabel, Field } from "@measured/puck";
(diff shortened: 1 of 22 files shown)


src/components/YextEntityFieldSelector.tsx
```diff
@@ -42,41 +43,51 @@ export const YextEntityFieldSelector = <T extends Record<string, any>>(
     label: props.label,
     render: ({ field, value, onChange }: RenderProps) => {
       const filteredEntityFields = getFilteredEntityFields(props.filter);
+      const [entityFieldMode, setEntityFieldMode] = useState<boolean>(true);
+
+      const toggleEntityFieldMode = () => {
+        setEntityFieldMode((prevMode) => !prevMode);
+        onChange({
```
benlife5, line 50: This makes it so when entityFieldMode is toggled off or on, both fields reset to ""?
I know we discussed this with Aaron a lot but I don't remember where we ended up. My concern with this is that an entity that doesn't have data for this field will return ""

src/components/YextEntityFieldSelector.tsx
```diff
@@ -87,8 +98,37 @@ export const YextEntityFieldSelector = <T extends Record<string, any>>(
...
+
+const ToggleMode = ({
+  entityFieldMode,
+  toggleEntityFieldMode,
+}: {
+  entityFieldMode: boolean;
+  toggleEntityFieldMode: () => void;
+}) => {
+  return (
+    <div className="ve-mt-2">
```
benlife5, line 121: Screenshot 2024-10-02 at 2 33 18 PM
I think this needs some additional layout styles

src/components/YextEntityFieldSelector.tsx
```diff
@@ -42,41 +43,51 @@ export const YextEntityFieldSelector = <T extends Record<string, any>>(
     label: props.label,
     render: ({ field, value, onChange }: RenderProps) => {
       const filteredEntityFields = getFilteredEntityFields(props.filter);
+      const [entityFieldMode, setEntityFieldMode] = useState<boolean>(true);
```
benlife5, line 46: whether it's in entityFieldMode should be based on constantValueEnabled (line 14)

benlife5 (commented):
@mkilpatrick Should we have another item for updating resolveYextEntityField?

src/components/YextEntityFieldSelector.tsx
asanehisa: hmmm constantValueEnabled isn't a prop received by the YextEntityFieldSelector?

src/components/YextEntityFieldSelector.tsx
asanehisa: Oh hmm let me look at slack threads + whats currently returned w/ nothing being manually selected by user

src/components/YextEntityFieldSelector.tsx
benlife5: It should be on `value`... ideally `value` would be typed as `YextEntityField` -- not sure if we missed that or if it's not possible bc it's dynamically reading the Puck field

src/components/YextEntityFieldSelector.tsx
asanehisa: https://github.com/user-attachments/assets/a615910a-614c-4382-b9a8-9122541c8fec

src/components/YextEntityFieldSelector.tsx
asanehisa: oh value is set in the YextEntityFieldSelector though. So user can't initially pass in a value to it? But can change it in the UI after it loads?

mkilpatrick: I'm not sure I fully understand the question, but `constantValueEnabled` is updated based on when the user clicks the button to enable constantValue mode or not. We should use that to determine what to render.

src/components/YextEntityFieldSelector.tsx
```diff
@@ -42,33 +43,45 @@ export const YextEntityFieldSelector = <T extends Record<string, any>>(
...
+                    ...filteredEntityFields.map((entityFieldNameToSchema) => {
+                      return {
+                        label: entityFieldNameToSchema.name,
+                        value: entityFieldNameToSchema.name,
+                      };
+                    }),
+                  ],
+                }}
+                onChange={(selectedEntityField) => {
+                  onChange({
```
mkilpatrick, line 75: Since you set all of the values here, I don't think `toggleEntityFieldMode` is necessary at all. The check on entityFieldMode on line 59 should be `!value?.constantValueEnabled`.

src/components/YextEntityFieldSelector.tsx
```diff
@@ -87,8 +101,37 @@ export const YextEntityFieldSelector = <T extends Record<string, any>>(
...
+}: {
+  entityFieldMode: boolean;
+  toggleEntityFieldMode: () => void;
+}) => {
+  return (
+    <div className="ve-mt-2 ve-w-full">
+      <Button onClick={toggleEntityFieldMode} variant="secondary">
+        {entityFieldMode ? (
+          <>
+            <Lock className="sm-icon" />
```
mkilpatrick, line 128: Idk if we need a lock icon. I'd ask in the channel with a screenshot of what it looks like. It should be full width. It most likely should only change the text.

src/utils/resolveYextEntityField.ts
```diff
@@ -14,7 +14,7 @@ export const resolveYextEntityField = <T>(
   }
 
   // return constant value if field is not set
-  if (entityField.field === "") {
+  if (entityField.field === "" || entityField.constantValueEnabled) {
```
mkilpatrick, line 17: We should only render constant value if it's enabled now. If field is blank then it should just return empty string along with a console warning. Though I'm not 100% sure about that. Let's check with Aaron on the fallback. But IMO if you haven't enabled constant value then it should never use it.

src/components/YextEntityFieldSelector.tsx
asanehisa: Sounds good, that's what I did

src/components/YextEntityFieldSelector.tsx
asanehisa: When you click "Use Constant Value", the constantValueEnabled isn't changed immediately, but only onChange for the fieldValue. That's why I added toggleEntityFieldMode, so it is changed right away on toggle.

src/utils/resolveYextEntityField.ts
asanehisa: Okay! I'll ask

asanehisa: Conclusion, don't fallback. https://yext.slack.com/archives/C06A06BCUUF/p1727961313799639?thread_ts=1727957859.737239&cid=C06A06BCUUF

src/components/YextEntityFieldSelector.tsx
mkilpatrick: Gotcha. But you don't need entityFieldMode

src/components/YextEntityFieldSelector.tsx
mkilpatrick: Did we get an answer on this? Also, is the button 100% width now?

asanehisa: Yup 100% width, and confirmed we'll remove the lock icon

src/utils/resolveYextEntityField.ts
```diff
@@ -38,9 +37,6 @@ export const resolveYextEntityField = <T>(
     console.error("Error in resolveYextEntityField:", e);
   }
 
-  // if field not found, return constant value as a fallback
-  console.warn(
-    `The field ${entityField.field} was not found in the document, defaulting to constant value ${entityField.constantValue}.`
-  );
-  return entityField.constantValue as T;
+  console.warn(`The field ${entityField.field} was not found in the document.`);
```
benlife5, line 40: Maybe have a separate warning for if field === ""? Since that would be because the user didn't select a field rather than the field not existing in the document.

benlife5 (commented):
small comment, otherwise lgtm!

src/utils/resolveYextEntityField.ts
asanehisa: oops good catch! don't wanna spam the logs haha

mkilpatrick: Good idea

src/components/YextEntityFieldSelector.tsx
```diff
@@ -114,17 +119,28 @@ const ToggleMode = ({
...
-        ) : (
-          <p>Use Constant Value</p>
-        )}
-      </Button>
+    <div className="ve-mb-2 ve-w-full">
+      <RadioGroup defaultValue={constantValueEnabled?.toString() ?? "false"}>
+        <div className="ve-flex ve-items-center ve-space-x-2">
+          <RadioGroupItem
+            value="false"
+            id="r1"
```
mkilpatrick, line 130: Update id and labelFor

src/components/YextEntityFieldSelector.tsx
```diff
@@ -114,17 +119,28 @@ const ToggleMode = ({
...
+            onClick={() => toggleConstantValueEnabled(false)}
+          />
+          <Label htmlFor="r1">Use Entity Value</Label>
+        </div>
+        <div className="ve-flex ve-items-center ve-space-x-2">
+          <RadioGroupItem
+            value="true"
+            id="r2"
```
mkilpatrick, line 138: same

src/components/YextEntityFieldSelector.tsx
```diff
@@ -127,18 +127,18 @@ const ToggleMode = ({
         <div className="ve-flex ve-items-center ve-space-x-2">
           <RadioGroupItem
             value="false"
-            id="r1"
+            id="false"
```
mkilpatrick, line 130: It needs an actual id. Ids are unique for DOM elements. I'd call it something like ve-use-entity-value and ve-use-constant-value

asanehisa: Ohh forgot about that. Fixed!

mkilpatrick (approved):

PR #71, opened 2024-10-02
fix: hide sidebar titles in theme mode via DOM

benlife5:
There's no way to get rid of the sidebar section titles via individual component puck overrides.

We could override the entire Puck component, but most of the default Puck components are not exported so we'd have to fork a number of files from Puck into our package and maintain them in parallel.

This solution grabs entire ComponentList from the left side bar and the title from the right side bar and hides them via the DOM.


mkilpatrick (approved):
Interesting solution but much better than having to rebuild it.

PR #72, opened 2024-10-03
feat(util): add buildCssOverridesStyle

briantstephan:
This adds a `buildCssOverridesStyle` util function that takes in an object containing the variable name and value overrides that need to be added to the page in a `style` tag. The `selector` field will be `.components` for our use case but I left it as a parameter in case we want to reuse it later. It also takes in an optional `base` field that gets prepended to the string that gets returned, in case the user needs to add anything else to the page's `<head>` tag besides these styling overrides. 

Verified the function works via the test cases and running it manually and confirming the returned string is what I expected.


src/utils/buildCssOverridesStyle.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const buildCssOverridesStyle = (
+  base: string,
+  selector: string,
+  overrides: CssVariableOverrides
+): string => {
+  return (
+    `${base}<style type="text/css">${selector}{` +
+    Object.entries(overrides)
+      .map(([key, value]) => `${key}:${value}!important`)
```
mkilpatrick, line 13: Add space before !important

src/utils/buildCssOverridesStyle.ts
```diff
@@ -0,0 +1,17 @@
+export type CssVariableOverrides = {
+  [key: string]: string;
+};
+
+export const buildCssOverridesStyle = (
```
mkilpatrick, line 5: Maybe we call it `applyTheme`?

I also think it should take in the document directly and we pull off the theme field, which will be part of the site entity and have a specific name. Presumably that field will be called `theme` but for now (since not a built-in type) it would be `c_theme`.

src/utils/buildCssOverridesStyle.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      '<style type="text/css">.components{--primary:0 68% 51%!important;--primary-foreground:0 0% 100%!important;--secondary:11 100% 26%!important}</style>'
+    );
+  });
+
+  it("should handle an empty overrides object", () => {
+    const selector = "div";
+    const overrides: CssVariableOverrides = {};
+    const result = buildCssOverridesStyle("", selector, overrides);
+
+    expect(result).toBe('<style type="text/css">div{}</style>');
```
mkilpatrick, line 37: If there are no overrides We shouldn't append anything.

src/utils/buildCssOverridesStyle.test.ts
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  it("should handle if a base string is provided", () => {
+    const base = "<style>div{color:blue}</style>";
```
mkilpatrick, line 41: The `other` field is to add any other arbitrary things into <head>. You commit message talks about it being for TW but it's not.

src/utils/buildCssOverridesStyle.ts
```diff
@@ -0,0 +1,17 @@
...
+  base: string,
```
mkilpatrick, line 6: If you make this last it can be optional so a user doesn't need to pass empty string if they don't have other overrides.

mkilpatrick (approved):

PR #75, opened 2024-10-04
fix(components): fixed history desync after publish

jwartofsky-yext:
We were using InitialPuckHistory which does not update to reset the local changes.

We also were not updating the history after clearing local changes.

After these changes, history is updated and clear local changes is disabled after publish.

After publish, user can add and clear local changes as expected.

### DIFF
--- a/src/internal/components/InternalEditor.tsx
+++ b/src/internal/components/InternalEditor.tsx
@@ -172,11 +172,13 @@ export const InternalEditor = ({
               refreshPermissions();
             }, [themeModeActive]);
 
+            const { history } = usePuck();
+
             return customHeader(
               handleClearLocalChanges,
               handleHistoryChange,
               appState.data,
-              puckInitialHistory?.histories[0].state.data, // used for clearing local changes - reset to first puck history
+              history?.histories[0].state.data, // used for clearing local changes - reset to first puck history
               handleSave,
               templateMetadata.isDevMode && !templateMetadata.devOverride,
               themeModeActive,
--- a/src/internal/components/InternalEditor.tsx
+++ b/src/internal/components/InternalEditor.tsx
@@ -172,8 +172,6 @@ export const InternalEditor = ({
               refreshPermissions();
             }, [themeModeActive]);
 
-            const { history } = usePuck();
-
             return customHeader(
               handleClearLocalChanges,
               handleHistoryChange,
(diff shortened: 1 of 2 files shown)


src/internal/components/InternalEditor.tsx
```diff
@@ -172,11 +172,13 @@ export const InternalEditor = ({
               refreshPermissions();
             }, [themeModeActive]);
 
+            const { history } = usePuck();
```
benlife5, line 175: There's already a usePuck a few lines above this

benlife5 (commented):
Nice fix!

src/internal/puck/components/Header.tsx
```diff
@@ -97,6 +97,7 @@ export const customHeader = (
             onClick={async () => {
               await handleSaveData(currentPuckData);
               handleClearLocalChanges();
+              setHistories([{ id: "root", state: { data: currentPuckData } }]);
```
mkilpatrick, line 100: I think what you really want to do is re-setPuckInitialHistory in clearHistory func, or maybe setPuckInitialHistoryFetched(false) so that loadPuckInitialHistory gets rerun

jwartofsky-yext: I was thinking about that, but I feel like re-setting puckInitialHistory goes against what the initial history is meant to be. I don't think we should have two history objects that we update separately as it was already pretty confusing while trying to debug this.

When we clear local changes we update the histories to match the first entry. Before we were using puckInitialHistory, but I think that should just be using puck history directly. I've tested and this setup works consistently and stays synchronized.

We also didn't add updating the histories for saving, which I've added here.

src/internal/components/InternalEditor.tsx
jwartofsky-yext: Fixed

src/internal/puck/components/Header.tsx
mkilpatrick: puckInitialHistory is the one history. What do you mean in your first comment?
> re-setting puckInitialHistory goes against what the initial history is meant to be
> I don't think we should have two history objects that we update separately

puckInitialHistory is used to tell Puck what history state (and histories) to render. What are we going against? And what is the second history object you're referring to?

If we do the work in clearHistory it handles both clearLocalChanges and publish since both of those call clearHistory.

jwartofsky-yext: In most other places outside of initialization, we are using the actual history/setHistories from usePuck. I think we should use that instead to keep things consistent, because this problem was caused by a desync between the two.

I'll move the change over to handleClearLocalChanges. We also have to pass in the data to set the history to.

mkilpatrick (approved):

asanehisa (approved):

PR #76, opened 2024-10-07
feat: receive isThemeMode from storm

asanehisa:


src/internal/components/InternalEditor.tsx
```diff
@@ -48,7 +48,9 @@ export const InternalEditor = ({
   themeConfig,
 }: InternalEditorProps) => {
   const [canEdit, setCanEdit] = useState<boolean>(false);
-  const [themeModeActive, setThemeModeActive] = useState<boolean>(false);
+  const [themeModeActive, setThemeModeActive] = useState<boolean>(
+    !!templateMetadata.isThemeMode
```
mkilpatrick, line 52: Is the coalesce necessary?

src/internal/puck/components/Header.tsx
```diff
@@ -55,10 +56,12 @@ export const customHeader = (
       <div className="header-left">
         <ToggleUIButtons />
         <ToggleEntityFields />
-        <ToggleThemeMode
-          themeModeActive={themeModeActive}
-          toggleThemeMode={setThemeModeActive}
-        />
+        {!isThemeMode && (
+          <ToggleThemeMode
```
mkilpatrick, line 60: This is going to be removed later?

src/internal/components/InternalEditor.tsx
asanehisa: the !! is b/c I didn't wanna make isThemeMode required (for now), so it doesn't break sites w/o changes.

src/internal/puck/components/Header.tsx
asanehisa: Oh we don't want the toggleThemeMode toggle ever? If so, then yes I can remove it later

mkilpatrick: Nope, it will be set from getting to it via the nav instead

mkilpatrick (approved):

PR #77, opened 2024-10-07
refactor: generate theme variables automatically and merge saved state with themeConfig for UI

benlife5:
See https://github.com/YextSolutions/pages-visual-editor-starter/tree/example-auto-css-variables for example use
Particularly https://github.com/YextSolutions/pages-visual-editor-starter/blob/example-auto-css-variables/theme.config.ts

- generates CSS variables from the object structure of theme.config.ts
- creates a default state object from the default values defined in theme.config.ts
- applyChanges merges any saved overrides with the default state
- ThemeSidebar reads the combined state into Puck fields
- ThemeSidebar converts Puck onChange output into CSS variables
- ThemeSidebar re-applies CSS variables to iframe preview
- Color Picker integrated with sidebar (added show/hide, support for hsl/hex/rgb in theme.config.ts)

https://github.com/user-attachments/assets/3e1a8aa6-0707-4876-b52e-81b5fc8ac3b4


src/internal/utils/internalThemeResolver.ts
```diff
@@ -0,0 +1,83 @@
+import {
+  ParentStyle,
+  ThemeConfig,
+  SavedTheme,
+} from "../../utils/themeResolver.ts";
+
+// internalThemeResolver returns a mapping of css variable names to their valuess
```
asanehisa, line 7: values

src/utils/applyTheme.ts
```diff
@@ -1,22 +1,44 @@
...
+export const updateThemeInEditor = (
+  newTheme: SavedTheme,
+  themeConfig: ThemeConfig
+) => {
+  const previewFrame = document.getElementById(
+    "preview-frame"
+  ) as HTMLIFrameElement;
+  if (previewFrame && previewFrame.contentDocument) {
+    const styleOverride = previewFrame?.contentDocument?.getElementById(
+      "visual-editor-theme"
```
mkilpatrick, line 38: Make this id value a const that can be used here and on line 21

src/utils/themeResolver.test.ts
```diff
@@ -9,164 +9,173 @@ import {
...
     "restructures theme config and merges with developer-specified tailwind config, " +
       "prioritizing developer sepecifications",
     () => {
-      const result = themeResolver(developerTailwindConfig, themeConfig);
+      const consoleSpy = vi.spyOn(console, "warn");
+      const result = themeResolver(developerTailwindConfig, testThemeConfig);
       expect(result).toEqual(mergedConfig);
+      expect(consoleSpy).toHaveBeenCalledTimes(3);
+      expect(consoleSpy).toHaveBeenLastCalledWith(
+        "Both theme.config.ts and tailwind.config.ts provided a value for sm. Using the value from tailwind.config.ts (0px)"
```
mkilpatrick, line 34: tailwind.config takes precedence over theme.config?

src/utils/themeResolver.ts
```diff
@@ -94,6 +99,11 @@ export const deepMerge = <T extends PlainObject, U extends PlainObject>(
         result[key] = deepMerge(value1, value2);
       } else {
         // Otherwise, prioritize obj1's value
+        if (value1 && value2) {
```
mkilpatrick, line 102: Similar comment which I know is not your change. I'd think the theme.config takes precedence. We should confirm with Product what they want to have happen, assuming the spec doesn't say it already.

src/utils/applyTheme.test.ts
```diff
@@ -115,13 +115,13 @@ const themeConfig: ThemeConfig = {
         label: "Large Border Radius",
         plugin: "borderRadius",
         type: "number",
-        default: `8px`,
+        default: 8,
```
mkilpatrick, line 118: Hmm, I wonder if this should stay a string or not. If I were writing some default css I wouldn't think to put a naked number and then it gets automatically converted to pixels.

src/utils/applyTheme.ts
```diff
@@ -6,6 +6,8 @@ export type Document = {
   [key: string]: any;
 };
 
+const THEME_STYLE_TAG_ID = "visual-editor-theme" as const;
```
mkilpatrick, line 9: `as const` is required for a const?

src/utils/applyTheme.test.ts
benlife5: Aaron said "For now, I meant for this to default to px. In the future, I could see us wanting to add the option for the dev or marketer to change this."

I think we will eventually want support for picking the unit. I agree this is a little weird so for now maybe we keep the number type represented as a string

src/utils/applyTheme.ts
benlife5: The `as const` sets the type of THEME_STYLE_TAG_ID as `"visual-editor-theme"` vs. `string`
Shouldn't really matter so I can drop it if it just adds clutter

mkilpatrick: I don't think the type matters since it's just used as a string so we can drop it.

jwartofsky-yext (dismissed):
lgtm

src/utils/applyTheme.test.ts
mkilpatrick: Did Aaron want this to be just a number? Maybe it just depends on what the input is actually rendered as. Is it some sort of number slider or is it just a normal input? If that latter we can go with string for now and update later.

benlife5: In the stories, there's not a number field in the UI mockup and in the example theme.config.ts it's just a number.

The sidebar renders the default Puck number field which is just a box with up and down arrows.

I'll switch to string, I think that's more useable/intuitive for now

mkilpatrick: Sounds good.

PR #82, opened 2024-10-10
feat: handle theme data from content and local storage

briantstephan:
This gets the theme data from content/local storage and sends it along to the theme sidebar so that the sidebar is up to date. It also handles saving by sending saved/published changes up to the parent.

Can't test until the data is sent down from the parent.


src/components/Editor.tsx
```diff
@@ -69,7 +84,7 @@ export const Editor = ({
     }
 
     return getLocalStorageKey(
-      !!templateMetadata.isThemeMode,
+      templateMetadata.isThemeMode,
```
mkilpatrick, line 87: Why is theme mode relevant to the non-theme key? We have a separate buildThemeKey func now

src/components/Editor.tsx
```diff
@@ -120,21 +147,38 @@ export const Editor = ({
     }
   }, [templateMetadata?.isDevMode]);
 
+  const clearLocalStorage = () => {
+    if (!templateMetadata?.isThemeMode) {
```
mkilpatrick, line 151: Prefer the positive, so flip this around to `if true clearTheme()`

src/components/Editor.tsx
```diff
@@ -120,21 +147,38 @@ export const Editor = ({
...
+      clearThemeLocalStorage();
+    }
+  };
+
   /**
    * Clears the user's localStorage and resets the current Puck history
    */
-  const clearLocalStorage = () => {
-    devLogger.logFunc("clearLocalStorage");
+  const clearPuckLocalStorage = () => {
```
mkilpatrick, line 161: clearVisualConfigLocalStorage?

src/components/Editor.tsx
```diff
@@ -120,21 +147,38 @@ export const Editor = ({
...
+    devLogger.logFunc("clearPuckLocalStorage");
     window.localStorage.removeItem(buildLocalStorageKey());
```
mkilpatrick, line 163: Should we update the storage key func name too?

src/components/Editor.tsx
```diff
@@ -144,7 +188,13 @@ export const Editor = ({
       return;
     }
     loadPuckInitialHistory(); // do something after state has updated
```
mkilpatrick, line 190: Can you remove this comment? It's not helpful

src/components/Editor.tsx
```diff
@@ -349,16 +526,31 @@ export const Editor = ({
...
+    "deleteThemeSaveState",
+    TARGET_ORIGINS
+  );
+
   const { sendToParent: saveVisualConfigData } = useSendMessageToParent(
     "saveVisualConfigData",
     TARGET_ORIGINS
   );
 
+  const { sendToParent: saveThemeConfigData } = useSendMessageToParent(
```
mkilpatrick, line 549: Group all of the theme stuff together

src/components/Editor.tsx
```diff
@@ -383,7 +575,11 @@ export const Editor = ({
...
     !visualConfigurationDataFetched ||
-    !puckInitialHistoryFetched;
+    !puckInitialHistoryFetched ||
+    (templateMetadata.isThemeMode &&
+      (!themeDataFetched ||
+        !themeSaveStateFetched ||
+        !themeInitialHistoryFetched));
 
   const progress: number =
     (100 * // @ts-expect-error adding bools is fine
```
mkilpatrick, line 585: Progress % needs to correspond with the things guarding isLoading

src/components/Editor.tsx
```diff
@@ -78,13 +78,12 @@ export const Editor = ({
     devLogger.logData("DOCUMENT", document);
   }
 
-  const buildLocalStorageKey = useCallback(() => {
-    if (!templateMetadata) {
+  const buildVisualConfigLocalStorageKey = useCallback(() => {
+    if (!templateMetadata || templateMetadata.isThemeMode) {
```
mkilpatrick, line 82: Don't we still need the visualConfiguration data from localStorage even if you're in theme mode?

jwartofsky-yext:
Data should be sent down from the parent now

src/components/Editor.tsx
briantstephan: Don't think so, I just prevented those fetches from occurring when in theme mode

src/components/Editor.tsx
```diff
@@ -210,7 +210,8 @@ export const Editor = ({
     if (
       !visualConfigurationDataFetched ||
       !saveStateFetched ||
-      !templateMetadata
+      !templateMetadata ||
+      templateMetadata.isThemeMode
```
mkilpatrick, line 214: We need this otherwise nothing will render in the Puck view

src/components/Editor.tsx
```diff
@@ -463,6 +464,9 @@ export const Editor = ({
   }, [templateMetadata]);
 
   useReceiveMessage("getSaveState", TARGET_ORIGINS, (send, payload) => {
+    if (templateMetadata?.isThemeMode) {
```
mkilpatrick, line 467: We need this too

src/components/Editor.tsx
```diff
@@ -483,6 +487,9 @@ export const Editor = ({
     "getVisualConfigurationData",
     TARGET_ORIGINS,
     (send, payload) => {
+      if (templateMetadata?.isThemeMode) {
```
mkilpatrick, line 490: And this

src/components/Editor.tsx
```diff
@@ -574,9 +581,10 @@ export const Editor = ({
     !puckConfig ||
     !templateMetadata ||
     !document ||
-    !saveStateFetched ||
-    !visualConfigurationDataFetched ||
-    !puckInitialHistoryFetched ||
+    (!templateMetadata.isThemeMode &&
```
mkilpatrick, line 584: Revert this

src/internal/puck/components/Header.tsx
```diff
@@ -78,7 +78,7 @@ export const customHeader = (
           </Button>
         )}
         <ClearLocalChangesButton
-          disabled={histories.length === 1}
+          disabled={isThemeMode ? false : histories.length === 1}
```
mkilpatrick, line 81: isThemeMode ? theme.histories.length === 1 : histories.length === 1, same below

src/components/Editor.tsx
mkilpatrick: None of the VConfig stuff should be affected by theme. We need it always.

mkilpatrick (approved):

PR #83, opened 2024-10-10
feat: load themes from content on published pages

benlife5:


src/utils/applyTheme.ts
```diff
@@ -13,7 +12,13 @@ export const applyTheme = (
   themeConfig: ThemeConfig,
   base?: string
 ): string => {
-  const overrides = document.c_theme;
+  const themeTextValue =
+    document?._site?.pagesTheme?.[0]?.themeConfiguration?.data;
```
mkilpatrick, line 16: We need to find the theme for the siteId (which is also on the doc)

mkilpatrick (approved):

PR #85, opened 2024-10-11
chore: clean up theme manager

benlife5:
- uses MutationObserver api to update styles
- moves all console logs to devLogger (should work with Brian's query params changes)
- removes puckRendered since we did not end up using it

### DIFF
--- a/src/components/Editor.tsx
+++ b/src/components/Editor.tsx
@@ -478,7 +478,6 @@ export const Editor = ({
   });
 
   useReceiveMessage("getThemeSaveState", TARGET_ORIGINS, (send, payload) => {
-    devLogger.log("getThemeSaveState PAYLOAD" + JSON.stringify(payload));
     const themeSaveState = {
       history: payload?.history
         ? jsonFromEscapedJsonString(payload?.history)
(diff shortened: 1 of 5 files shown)


src/components/Editor.tsx
```diff
@@ -510,7 +510,7 @@ export const Editor = ({
   );
 
   useReceiveMessage("getThemeData", TARGET_ORIGINS, (send, payload) => {
-    console.log(`payload in getThemeData is ${JSON.stringify(payload)}`);
+    devLogger.log("getThemeData PAYLOAD: " + JSON.stringify(payload));
```
mkilpatrick, line 513: In the other useReceiveMessages we have ie `devLogger.logData("SAVE_STATE", payload);`, can you do the same? And then remove the devlogger below that logs it again after it's json escaped. We don't really need it twice. Same in the others.

src/internal/components/InternalEditor.tsx
```diff
@@ -151,7 +149,6 @@ export const InternalEditor = ({
         overrides={{
           header: () => {
             const { refreshPermissions, config } = usePuck();
-            setPuckRendered(true);
```
mkilpatrick, line 154: I asked if there was a way to do this in Discord. They said we could try getting `iframeRef` from usePuck and then doing iframeRef.onload. Give that a shot to see if it allows us to remove the setTimeout

mkilpatrick: Jk, it's not available yet. Someone just put up a PR for it https://github.com/measuredco/puck/pull/652

chrisvxd: You could try `#preview-frame` to access the iframe

mkilpatrick (approved):

PR #86, opened 2024-10-14
feat: add dev mode theme preview

mkilpatrick:
Adds theme preview for dev mode.

- Also fixes preview for VE in dev mode for the initial load. It was only working after a change was made.
- Now only pulls the published VE data for theme mode
- Properly clears history between theme and VE

### DIFF
--- a/src/components/Editor.tsx
+++ b/src/components/Editor.tsx
@@ -107,6 +107,7 @@ export const Editor = ({
     );
   }, [templateMetadata]);
 
+  // redirect to 404 page when going to /edit page outside of Storm
   useEffect(() => {
     if (typeof window !== "undefined") {
       const ancestors = window.location.ancestorOrigins;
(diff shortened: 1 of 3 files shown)


src/components/Editor.tsx
```diff
@@ -441,6 +474,21 @@ export const Editor = ({
...
+      return;
+    }
+
+    const themeToSend = themeSaveState?.history[themeSaveState.index];
+
+    devLogger.logFunc("sendDevThemeSaveStateData useEffect");
+    sendDevThemeSaveStateData({
+      payload: { devThemeSaveStateData: JSON.stringify(themeToSend) },
+    });
+  }, [themeSaveStateFetched, themeSaveState, templateMetadata]);
```
jwartofsky-yext, line 490: Do you want themeSaveStateFetched here instead of themeInitialHistoryFetched? 

Above you are using puckInitialHistoryFetched.

mkilpatrick: Good catch

benlife5 (approved):

PR #87, opened 2024-10-14
refactor: separate theme manager and visual editor internal editors

benlife5:
Separates out the internal editors and headers for Layout and Theme mode


src/utils/applyTheme.ts
```diff
@@ -1,20 +1,17 @@
...
 
 const THEME_STYLE_TAG_ID = "visual-editor-theme";
-const devLogger = new DevLogger();
 
 export const applyTheme = (
   document: Document,
   themeConfig: ThemeConfig,
   base?: string
 ): string => {
-  devLogger.logFunc("applyTheme");
```
benlife5, line 17: the devlogger object would need to be passed into these functions so I just removed it

src/internal/hooks/useMessage.ts
```diff
@@ -216,3 +226,67 @@ const statusSetter = (setStatus: Dispatch<SetStateAction<MessageStatus>>) => {
     }
   };
 };
+
+export const useMessageSenders = () => {
```
benlife5, line 230: Not sure if there's a better name, but nice to move this boilerplate out of editor.tsx

src/internal/components/InternalLayoutEditor.tsx
```diff
@@ -0,0 +1,134 @@
...
+import React from "react";
+import { useState, useRef, useCallback } from "react";
+import { getVisualConfigLocalStorageKey } from "../utils/localStorageHelper.ts";
+import { TemplateMetadata } from "../types/templateMetadata.ts";
+import { EntityFieldProvider } from "../../components/EntityField.tsx";
+import { LayoutSaveState } from "../types/saveState.ts";
+import { LayoutHeader } from "../puck/components/LayoutHeader.tsx";
+import { DevLogger } from "../../utils/devLogger.ts";
+
+interface InternalLayoutEditorProps {
```
mkilpatrick, line 17: Not you, but can we change these to types instead of interfaces? 99% of the time types should be preferred. Same in other places like InternalThemeEditor.

src/internal/components/InternalThemeEditor.tsx
```diff
@@ -0,0 +1,102 @@
...
+  templateMetadata,
+  publishThemeConfiguration,
+  devLogger,
+  themeConfig,
+  saveThemeSaveState,
+  themeHistory,
+  setThemeHistory,
+  clearThemeHistory,
+}: InternalThemeEditorProps) => {
+  const [canEdit, setCanEdit] = useState<boolean>(false); // helps sync puck preview and save state
```
mkilpatrick, line 40: I wonder if your iframe observer would allows us to not use this now

src/internal/hooks/useMessage.ts
```diff
@@ -216,3 +226,67 @@ const statusSetter = (setStatus: Dispatch<SetStateAction<MessageStatus>>) => {
...
+  const { sendToParent: publishThemeConfiguration } = useSendMessageToParent(
+    "saveThemeData",
+    TARGET_ORIGINS
+  );
+
+  const { sendToParent: openQuickFind } = useSendMessageToParent(
+    "openQuickFind",
+    TARGET_ORIGINS
+  );
+  return {
```
mkilpatrick, line 280: Add empty line above

src/internal/hooks/useMessage.ts
mkilpatrick: I was thinking of doing something like this. Unfortunately it doesn't work for the receivers too since those need a bunch of props.

src/internal/hooks/useMessage.ts
```diff
@@ -216,3 +226,67 @@ const statusSetter = (setStatus: Dispatch<SetStateAction<MessageStatus>>) => {
...
+    "saveThemeSaveState",
+    TARGET_ORIGINS
+  );
+
+  const { sendToParent: deleteThemeSaveState } = useSendMessageToParent(
+    "deleteThemeSaveState",
+    TARGET_ORIGINS
+  );
+
+  const { sendToParent: publishThemeConfiguration } = useSendMessageToParent(
```
mkilpatrick, line 271: Should we group things that are common vs layout vs theme? I think that would be nice. Common would be the things anything would need. The other two are for their own specifics.

src/internal/hooks/useQuickFindShortcut.ts
```diff
@@ -0,0 +1,18 @@
+import { useEffect } from "react";
+import { useMessageSenders } from "./useMessage.ts";
+
+export const useQuickFindShortcut = () => {
```
mkilpatrick, line 4: Nice. Can you add tsdoc?

src/utils/applyTheme.ts
mkilpatrick: You can't create it at the top like it was before?

mkilpatrick (commented):
Looking a lot cleaner!

src/utils/applyTheme.ts
benlife5: Updated to use a static variable so that we can create new DevLogger() anywhere vs. needing to prop drill from the place where it receives the query param

src/internal/components/InternalThemeEditor.tsx
benlife5: When I looked at removing this a few weeks ago, it was a problem anytime you moved a component around. Probably worth looking at again while we're doing this refactoring work

src/components/Editor.tsx
```diff
@@ -93,11 +93,12 @@ export const Editor = ({
     saveThemeSaveState,
     deleteThemeSaveState,
     publishThemeConfiguration,
+    sendDevThemeSaveStateData,
   } = useMessageSenders();
   useQuickFindShortcut();
 
   const buildVisualConfigLocalStorageKey = useCallback(() => {
-    if (!templateMetadata) {
+    if (!templateMetadata || templateMetadata.isThemeMode) {
```
mkilpatrick, line 101: I don't actually think the themeMode check is necessary here or on 115, though it doesn't hurt.

src/internal/components/InternalThemeEditor.tsx
```diff
@@ -90,9 +132,8 @@ export const InternalThemeEditor = ({
           fields: () => (
             <ThemeSidebar
               themeConfig={themeConfig}
-              saveTheme={saveThemeSaveState}
               themeHistory={themeHistory!}
-              setThemeHistory={setThemeHistory}
+              handleThemeChange={handleThemeChange}
```
mkilpatrick, line 136: onThemeChange as the prop - https://medium.com/@master_43681/understanding-and-proper-usage-of-on-and-handle-prefixes-in-react-function-naming-119f0beea400. It's really convention. I'm sure there are other places we should update too.

src/utils/devLogger.ts
```diff
@@ -13,42 +13,45 @@ export type DevLoggerPrefix =
   | "THEME_VALUES_TO_APPLY";
 
 export class DevLogger {
-  constructor(private enabled: boolean = false) {
-    if (!this.enabled) {
+  private static enabled: boolean;
+
+  constructor(enabled: boolean = false) {
```
mkilpatrick, line 18: Any reason for the changes to DevLogger? I prefer how it was.

mkilpatrick: Nvm, saw your comment about making it static.

src/components/Editor.tsx
jwartofsky-yext, line 1: Not you, but I wonder if we could separate loadPuckInitialHistory and loadThemeInitialHistory into separate files.

jwartofsky-yext (dismissed):
lgtm

Big improvement over what it was before

src/utils/devLogger.ts
benlife5: enabled only gets set once, via YSS passing down the query params. So either this needs to be static or the devLogger instance needs to be passed down as a prop from editor anywhere we want to use it.

src/components/Editor.tsx
```diff
@@ -343,6 +357,23 @@ export const Editor = ({
     getVisualConfigLocalStorageKey,
   ]);
 
+  // Handles sending the devSaveStateData on initial load so that nothing isn't rendered for preview.
+  // Subsequent changes are sent via handleHistoryChange in InternalEditor.tsx.
+  useEffect(() => {
```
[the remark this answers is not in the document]
benlife5: I think that would be ideal. I tried it some yesterday and was running into problems with getting everything to load in the correct order and ultimately needing a lot of shared logic. But things are cleaner now so I can give it another go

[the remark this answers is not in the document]
mkilpatrick: Discussed offline. These changes will be another PR.

mkilpatrick (dismissed):

src/components/Editor.tsx
benlife5: Yeah, would be nice. Will try after we merge this PR

mkilpatrick (dismissed):

src/internal/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    return getThemeLocalStorageKey(
+      templateMetadata.isDevMode,
+      templateMetadata.siteId,
+      templateMetadata.themeEntityId
+    );
+  }, [templateMetadata]);
+
+  const clearLocalStorage = () => {
```
jwartofsky-yext, line 48: no need to change for this PR, but I feel like having these two be linked is bad

src/internal/components/InternalThemeEditor.tsx
```diff
@@ -0,0 +1,150 @@
...
+          index: newHistory.index,
+        },
+      });
+    }
+
+    updateThemeInEditor(newThemeValues, themeConfig);
+    setThemeHistory(newHistory);
+  };
+
+  const change = async () => {
```
jwartofsky-yext, line 102: Should puck ever be changing while in theme mode?

src/internal/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+function getThemeLocalStorageKey(
+  isDevMode: boolean,
+  siteId: number,
+  themeEntityId?: number
+): string {
+  if (!themeEntityId) {
+    throw new Error(
+      "Unable to generate local storage key for themes, missing query parameters"
```
jwartofsky-yext, line 112: not you, but could just say "missing themeId" since that is the only query param we check here

jwartofsky-yext (approved):
lgtm

Feel free to ack the comments since they are minor

src/internal/hooks/useLocalStorage.ts
benlife5: Good call, will fix in next PR

src/internal/hooks/useLocalStorage.ts
benlife5: ack, will fix in next PR

src/internal/components/InternalThemeEditor.tsx
benlife5: This will still get called when the puck ui state changes. Not sure if we need to care about that in theme mode (but also not really sure what this actually does in layout mode). Will take a look after the refactoring

PR #88, opened 2024-10-15
refactor: simplify Editor.tsx and create LayoutEditor and ThemeEditor

benlife5:
- Moves the useReceiveMessages into their own file
- Keeps shared logic in Editor.tsx
- Splits the rest into LayoutEditor and ThemeEditor

This _does not_ prevent the API calls for themes in Layout mode and layout save state in Theme mode. The problem is that we trigger fetching all data via sending a iFrameLoaded message to YSS. We could change YSS to have a trigger for shared data, a trigger for layout data, and a trigger for theme data, but that seemed out of scope for this PR. That could also potentially be slower because there would be two stages of concurrent requests instead of just one. 

That being said, this PR will progress past the loading stage in Layout mode even if theme data hasn't returned and vice versa. So, that should speed up loading times on the front end, at the expense of wasted API calls.


src/components/Editor.tsx
```diff
@@ -207,387 +94,44 @@ export const Editor = ({
...
-        saveStateFetched +
-        visualConfigurationDataFetched +
-        puckInitialHistoryFetched +
-        (templateMetadata?.isThemeMode // @ts-expect-error adding bools is fine
-          ? themeDataFetched +
-            themeSaveStateFetched +
-            themeInitialHistoryFetched
-          : 0))) /
-    (templateMetadata?.isThemeMode ? 9 : 6);
+    90 * // @ts-expect-error adding bools is fine
```
benlife5, line 107: Didn't exactly know what to do with the progress bar because there is still the puckInitialHistoryFetched and/or themeInitialHistoryFetched step in the individual editors before we display Puck but I think those should be near-instant because there's no network requests

src/internal/components/LayoutEditor.tsx
```diff
@@ -0,0 +1,234 @@
...
+          clearHistory={clearHistory}
+          templateMetadata={templateMetadata}
+          layoutSaveState={layoutSaveState}
+          saveLayoutSaveState={saveLayoutSaveState}
+          publishVisualConfiguration={publishVisualConfiguration}
+          sendDevSaveStateData={sendDevLayoutSaveStateData}
+          buildVisualConfigLocalStorageKey={buildVisualConfigLocalStorageKey}
+        />
+      ) : (
+        <LoadingScreen progress={90} />
```
asanehisa, line 230: Is LoadingScreen progress={90} purposeful?

benlife5: I set Editor.tsx loads to load from 0 to 90 and then ThemeEditor/LayoutEditor will either be at 90 or be loaded. I'm not sure that's the best way to do it

src/internal/components/LayoutEditor.tsx
```diff
@@ -0,0 +1,234 @@
...
+    layoutSaveState,
+    visualConfigurationData,
+  } = props;
+
+  const {
+    sendDevLayoutSaveStateData,
+    saveLayoutSaveState,
+    publishVisualConfiguration,
+    deleteLayoutSaveState,
+  } = useMessageSenders();
```
mkilpatrick, line 34: What about breaking this out even more?

/hooks/messageSenders -> export commonMessageSenders()
/hooks/layout/messageSenders.ts -> export layoutMessageSenders()
/hooks/theme/messageSenders.ts -> export themeMessageSenders()

I could see naming the files like common/layout/themeMessageSenders.ts but then it's redundant, but helps with opening files when 3 have the same name.

But this would at least prevent us from creating senders (and receivers if you do the same) for things not necessary in layout vs theme.

src/internal/components/LayoutEditor.tsx
mkilpatrick: Maybe annoying, but in Editor.tsx instead of showing loading there at all, pass the number of loading things through which can be added here and show the loading state in each sub editor. Then you can add to that number however many other things need to load in this component to show progress properly.

src/internal/components/ThemeEditor.tsx
```diff
@@ -0,0 +1,261 @@
...
+   * For theme mode, only load from Content
+   */
+  const loadPuckInitialHistory = useCallback(() => {
+    if (!templateMetadata) {
+      return;
+    }
+    devLogger.logFunc("loadPuckInitialHistory");
+
+    // Only load Content data for theme mode
+    if (templateMetadata.isThemeMode) {
```
mkilpatrick, line 66: Presumably we've already guarded the component by this so any isThemeMode checks should be unncessary.

src/internal/hooks/useLocalStorage.ts
```diff
@@ -70,7 +66,8 @@ export const useLocalStorage = (
   };
 
   return {
-    clearLocalStorage,
+    clearThemeLocalStorage,
```
mkilpatrick, line 69: Along the lines of my hook separation comment above, maybe we should do the same for useLocalStorage. The paradigm is always
at root: /x/commonThing
/x/subItem1/item1SpecificThing
/x/subItem2/item2SpecificThing

src/internal/components/LayoutEditor.tsx
benlife5: Tried that a bit but the problem is we still have to wait for templateMetadata before selecting a sub editor, so we still need a loading in the top level editor.

What I did in the latest commit is just have Editor do the 0-100 loading with no loading bar in the subeditors. The subeditors are just waiting for PuckInitialHistory and/or ThemeInitialHistory, both of which do not make any network requests so they load very quickly. https://dev.yext.com/s/1000152098/yextsites/63009/editor#view=EDIT&entityId=8725521 seems just as smooth as it is in production right now.

src/components/Editor.tsx
```diff
@@ -104,14 +104,14 @@ export const Editor = ({
...
       !!templateMetadata +
       !!document +
-      (layoutSaveStateFetched || templateMetadata?.isThemeMode) +
+      ((layoutSaveStateFetched || templateMetadata?.isThemeMode) ?? 0) +
       visualConfigurationDataFetched +
-      (themeDataFetched || !templateMetadata?.isThemeMode) +
-      (themeSaveStateFetched || !templateMetadata?.isThemeMode)) /
+      ((themeDataFetched || !templateMetadata?.isThemeMode) ?? 0) +
+      ((themeSaveStateFetched || !templateMetadata?.isThemeMode) ?? 0)) /
       7);
```
mkilpatrick, line 115: The 7 needs to change based on themeMode too. If it's not theme mode we don't care about waiting for the themeData and blocking the layout component.

benlife5: `(themeSaveStateFetched || !templateMetadata?.isThemeMode) == (false || !false) == true` if it's not theme mode and themeData isn't loaded, so it will continue

mkilpatrick (approved):

PR #91, opened 2024-10-17
feat: add hash to ThemeSaveState

asanehisa:
But on back doesn't go back to previous hash. Can put this up as temp fix if we want?

### DIFF
--- a/src/internal/components/InternalThemeEditor.tsx
+++ b/src/internal/components/InternalThemeEditor.tsx
@@ -10,6 +10,7 @@ import { ThemeSaveState } from "../types/themeSaveState.ts";
 import { ThemeHeader } from "../puck/components/ThemeHeader.tsx";
 import { generateCssVariablesFromPuckFields } from "../utils/internalThemeResolver.ts";
 import { updateThemeInEditor } from "../../utils/applyTheme.ts";
+import { generateId } from "../utils/generateId.ts";
 
 const devLogger = new DevLogger();
 
@@ -69,6 +70,7 @@ export const InternalThemeEditor = ({
     const newHistory = {
       history: [...themeHistory.history, newThemeValues],
       index: themeHistory.index + 1,
+      hash: generateId("history"),
     };
 
     window.localStorage.setItem(
@@ -91,6 +93,7 @@ export const InternalThemeEditor = ({
         payload: {
           history: JSON.stringify(newHistory.history),
           index: newHistory.index,
+          hash: newHistory.hash,
         },
       });
     }
--- a/src/internal/components/ThemeEditor.tsx
+++ b/src/internal/components/ThemeEditor.tsx
@@ -8,6 +8,7 @@ import { ThemeConfig } from "../../utils/themeResolver.ts";
 import { updateThemeInEditor } from "../../utils/applyTheme.ts";
 import { InternalThemeEditor } from "./InternalThemeEditor.tsx";
 import { useThemeMessageSenders } from "../hooks/theme/useMessageSenders.ts";
+import { generateId } from "../utils/generateId.ts";
 
 const devLogger = new DevLogger();
 
@@ -46,7 +47,7 @@ export const ThemeEditor = (props: ThemeEditorProps) => {
   const [puckInitialHistoryFetched, setPuckInitialHistoryFetched] =
     useState<boolean>(false);
   const [themeInitialHistory, setThemeInitialHistory] =
-    useState<ThemeSaveState>({ history: [{}], index: 0 });
+    useState<ThemeSaveState>({ history: [{}], index: 0, hash: "" });
   const [themeInitialHistoryFetched, setThemeInitialHistoryFetched] =
     useState<boolean>(false);
 
@@ -98,6 +99,7 @@ export const ThemeEditor = (props: ThemeEditorProps) => {
         setThemeInitialHistory({
           history: localHistories,
           index: localHistoryIndex,
+          hash: generateId("localThemeHistory"),
         });
         setThemeInitialHistoryFetched(true);
         return;
@@ -111,6 +113,7 @@ export const ThemeEditor = (props: ThemeEditorProps) => {
         setThemeInitialHistory({
           history: [themeData],
           index: 0,
+          hash: generateId("themeHistory"),
         });
       }
       setThemeInitialHistoryFetched(true);
@@ -128,6 +131,7 @@ export const ThemeEditor = (props: ThemeEditorProps) => {
         setThemeInitialHistory({
           history: [themeData],
           index: 0,
+          hash: generateId("themeHistory"),
         });
       }
       setThemeInitialHistoryFetched(true);
@@ -144,6 +148,7 @@ export const ThemeEditor = (props: ThemeEditorProps) => {
       const themeInitialHistory = {
         history: history,
         index: history.length - 1,
+        hash: generateId("themeHistory"),
       };
       setThemeInitialHistory(themeInitialHistory);
       setThemeInitialHistoryFetched(true);
@@ -156,6 +161,7 @@ export const ThemeEditor = (props: ThemeEditorProps) => {
     setThemeInitialHistory({
       history: [{}],
       index: 0,
+      hash: "",
     });
     setThemeInitialHistoryFetched(true);
   }, [
--- a/src/internal/puck/components/ThemeHeader.tsx
+++ b/src/internal/puck/components/ThemeHeader.tsx
@@ -8,6 +8,7 @@ import { updateThemeInEditor } from "../../../utils/applyTheme.ts";
 import { EntityFieldsToggle } from "../ui/EntityFieldsToggle.tsx";
 import { UIButtonsToggle } from "../ui/UIButtonsToggle.tsx";
 import { ClearLocalChangesButton } from "../ui/ClearLocalChangesButton.tsx";
+import { generateId } from "../../utils/generateId.ts";
 
 type ThemeHeaderProps = {
   onPublishTheme: () => Promise<void>;
@@ -62,6 +63,7 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
             setThemeHistory({
               history: [themeHistory?.history[0]],
               index: 0,
+              hash: themeHistory?.hash ?? generateId("themeHistory"),
             });
           }}
         />
@@ -76,6 +78,7 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
                   themeHistory?.history[themeHistory?.history.length - 1],
                 ],
                 index: 0,
+                hash: "",
               });
             }}
           >
--- a/src/internal/utils/generateId.ts
+++ b/src/internal/utils/generateId.ts
@@ -1,0 +1,4 @@
+import { v4 as uuidv4 } from "uuid";
+
+export const generateId = (type?: string | number) =>
+  type ? `${type}-${uuidv4()}` : uuidv4();
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -13,7 +13,7 @@ export default defineConfig(() => ({
       formats: ["es", "cjs"] as LibraryFormats[], // typescript is unhappy without this forced type definition
     },
     rollupOptions: {
-      external: ["react", "react-dom", "@measured/puck"],
+      external: ["react", "react-dom", "@measured/puck", "uuid"],
       output: {
         globals: {
           react: "React",
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
     "@types/react-color": "^3.0.12",
     "@types/react-dom": "^18.3.0",
     "@types/semver": "^7.5.8",
+    "@types/uuid": "^10.0.0",
     "@typescript-eslint/eslint-plugin": "^8.0.1",
     "@typescript-eslint/parser": "^8.0.1",
     "@vitejs/plugin-react": "^4.3.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -130,6 +130,9 @@ devDependencies:
   '@types/semver':
     specifier: ^7.5.8
     version: 7.5.8
+  '@types/uuid':
+    specifier: ^10.0.0
+    version: 10.0.0
   '@typescript-eslint/eslint-plugin':
     specifier: ^8.0.1
     version: 8.0.1(@typescript-eslint/parser@8.0.1)(eslint@9.8.0)(typescript@5.5.4)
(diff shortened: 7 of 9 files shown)


src/internal/utils/generateId.ts
```diff
@@ -0,0 +1,4 @@
+import { v4 as uuidv4 } from "uuid";
+
+export const generateId = (type?: string | number) =>
+  type ? `${type}-${uuidv4()}` : uuidv4();
```
mkilpatrick, line 4: Why does type matter? We can just use uuidv4 directly can't we?

vite.config.ts
```diff
@@ -13,7 +13,7 @@ export default defineConfig(() => ({
       formats: ["es", "cjs"] as LibraryFormats[], // typescript is unhappy without this forced type definition
     },
     rollupOptions: {
-      external: ["react", "react-dom", "@measured/puck"],
+      external: ["react", "react-dom", "@measured/puck", "uuid"],
```
mkilpatrick, line 16: We needed to externalize uuid? If we do that I don't think it'd work in a starter without the starter having uuid as a dep, though I could be wrong.

src/internal/utils/generateId.ts
asanehisa: Oh yeah we can, I was just copying puck's method of generateId so it'd match the saveState hash's. I can remove this

vite.config.ts
asanehisa: Oh hmmm I did that b/c

[vite]: Rollup failed to resolve import "uuid" from "/Users/asanehisa/visual-editor/src/internal/utils/generateId.ts".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`

asanehisa: yay tested + works fine in the pages visual editor starter (even without adding uuid as dep to that starter)

src/internal/utils/generateId.ts
mkilpatrick: Yea we don't need type so let's remove

src/internal/components/ThemeEditor.tsx
```diff
@@ -95,6 +96,7 @@ export const ThemeEditor = (props: ThemeEditorProps) => {
         setThemeInitialHistory({
           history: localHistories,
           index: localHistoryIndex,
+          hash: uuidv4(),
```
mkilpatrick, line 99: We need to update what the theme history contains. It needs to also include the hash. Hashes should only be generated when a new theme history is added. Since it was created and put into localStorage, we can't set new hashes when pulling it back out. Otherwise we'll never be able to find the correct history based on hash.

src/internal/components/ThemeEditor.tsx
```diff
@@ -108,6 +110,7 @@ export const ThemeEditor = (props: ThemeEditorProps) => {
         setThemeInitialHistory({
           history: [themeData],
           index: 0,
+          hash: uuidv4(),
```
mkilpatrick, line 113: It's ok here, but also needs to be part of the history themeData object. See how we do it for initialPuckHistory (which this whole func basically emulates). 

```
setPuckInitialHistory({
          histories: [{ id: "root", state: { data: visualConfigurationData } }],
          appendData: false,
        });
```
id === hash here. We could just call it "root" as well.

src/internal/components/ThemeEditor.tsx
```diff
@@ -125,6 +128,7 @@ export const ThemeEditor = (props: ThemeEditorProps) => {
...
```
mkilpatrick, line 131: "root" here too

src/internal/components/ThemeEditor.tsx
```diff
@@ -141,6 +145,7 @@ export const ThemeEditor = (props: ThemeEditorProps) => {
       const themeInitialHistory = {
         history: history,
         index: history.length - 1,
+        hash: themeSaveState.hash,
```
mkilpatrick, line 148: We had to not use localStorage and set things up differently when we didn't have hash because we couldn't just look up by index as it wasn't unique. For example, in loadPuckInitialHistory we get the localStorage and look for the hash that matches saveState, but for themes we only had index so there's nothing to look up. Now that we have hash we should go back to using the same type of logic. Can you do that? Again, it's going to require changing the history object around I think.

src/internal/hooks/theme/useMessageReceivers.ts
```diff
@@ -32,6 +33,7 @@ export const useThemeMessageReceivers = () => {
         ? jsonFromEscapedJsonString(payload?.history)
         : [],
       index: payload?.index ?? 0,
+      hash: payload?.hash ?? uuidv4(),
     };
     devLogger.logData("THEME_SAVE_STATE", payload);
```
mkilpatrick, line 38: Not you but move this up to be the first line of the function.

src/internal/hooks/theme/useMessageReceivers.ts
```diff
@@ -32,6 +33,7 @@ export const useThemeMessageReceivers = () => {
         ? jsonFromEscapedJsonString(payload?.history)
         : [],
       index: payload?.index ?? 0,
+      hash: payload?.hash ?? uuidv4(),
```
mkilpatrick, line 36: uuid should already be set (going forward). We can clear the localStorage and themeSaveState tables so we don't need this. You can also setThemeSaveState directly from the payload instead of recreating the object here. The one thing I'm noticing though is that we json escape it here but we don't in the getSaveState receiver. We should make them consistent. Idk which is better to use. json escape right here or do it later? Probably here, which means we probably want to update getSaveState. We can do that in another PR though.

src/internal/puck/components/ThemeHeader.tsx
```diff
@@ -62,6 +63,7 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
             setThemeHistory({
               history: [themeHistory?.history[0]],
               index: 0,
+              hash: themeHistory?.hash ?? uuidv4(),
```
mkilpatrick, line 66: When would themeHistory not have hash set on it now?

src/internal/puck/components/ThemeHeader.tsx
```diff
@@ -76,6 +78,7 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
                   themeHistory?.history[themeHistory?.history.length - 1],
                 ],
                 index: 0,
+                hash: uuidv4(),
```
mkilpatrick, line 81: You don't want to generate a new one here. Use the one on themeHistory

src/internal/types/themeData.ts
```diff
@@ -0,0 +1,15 @@
...
+};
+
+export type ThemeHistories = {
+  histories: History[];
+  index?: number;
+};
+
+export type History<D = any> = {
+  state: D;
+  id?: string;
```
mkilpatrick, line 14: This shouldn't be optional for us. I know it is in the Puck History but that's because they changed it to be optional for the user. We want it always present.

src/internal/types/themeData.ts
```diff
@@ -0,0 +1,15 @@
+export type ThemeSaveState = {
+  history: any; // json object
+  index: number;
+  hash: string;
+};
+
+export type ThemeHistories = {
+  histories: History[];
+  index?: number;
```
mkilpatrick, line 9: I don't think we want this optional either. Is there a case where this wouldn't have a value?

src/internal/types/themeData.ts
```diff
@@ -0,0 +1,15 @@
...
+};
+
+export type History<D = any> = {
```
mkilpatrick, line 12: This shouldn't need to be generic. Themes have a specific format, which is just an array of strings.

mkilpatrick, line 12: I'd call this ThemeHistory

src/internal/types/themeData.ts
```diff
@@ -0,0 +1,15 @@
...
+  state: D;
```
mkilpatrick, line 13: What do other people think about this property name? Obviously it aligns with Puck but we don't need to. Maybe `value`?

src/internal/types/themeData.ts
```diff
@@ -0,0 +1,15 @@
+export type ThemeSaveState = {
+  history: any; // json object
```
mkilpatrick, line 2: This should be of type ThemeHistory (after name change).

src/internal/types/themeData.ts
```diff
@@ -0,0 +1,15 @@
...
+  index: number;
```
mkilpatrick, line 3: We won't need to save this soon because hash is superseding it. Probably just remove.

src/internal/components/InternalThemeEditor.tsx
```diff
@@ -62,35 +63,40 @@ export const InternalThemeEditor = ({
...
     };
 
     const newHistory = {
-      history: [...themeHistory.history, newThemeValues],
-      index: themeHistory.index + 1,
+      histories: [
+        ...themeHistory.histories,
+        { id: uuidv4(), state: { data: newThemeValues } },
+      ] as History[],
+      index: themeHistory.histories.length,
```
mkilpatrick, line 74: Should this be length - 1?

src/internal/components/ThemeEditor.tsx
```diff
@@ -95,9 +96,11 @@ export const ThemeEditor = (props: ThemeEditorProps) => {
       if (localHistoryArray) {
         devLogger.log("Theme Dev Mode - Using theme localStorage");
         const localHistories = JSON.parse(localHistoryArray);
-        const localHistoryIndex = localHistories.length - 1;
+        const localHistoryIndex = JSON.parse(localHistoryArray).findIndex(
```
mkilpatrick, line 99: devMode doesn't use saveState. This should just set the index to the last item just like loadPuckInitialHistory

src/internal/components/ThemeEditor.tsx
```diff
@@ -143,7 +146,7 @@ export const ThemeEditor = (props: ThemeEditorProps) => {
         ? [themeData, ...themeSaveState.history]
         : [...themeSaveState.history];
       const themeInitialHistory = {
-        history: history,
+        histories: history,
```
mkilpatrick, line 149: All of line 143-164 needs to be updated to match what we do in loadPuckInitialHistory

src/internal/components/ThemeEditor.tsx
```diff
@@ -42,8 +42,9 @@ export const ThemeEditor = (props: ThemeEditorProps) => {
   >();
   const [puckInitialHistoryFetched, setPuckInitialHistoryFetched] =
     useState<boolean>(false);
-  const [themeInitialHistory, setThemeInitialHistory] =
-    useState<ThemeSaveState>({ history: [{}], index: 0 });
+  const [themeInitialHistory, setThemeInitialHistory] = useState<
```
mkilpatrick, line 45: Not you, but I don't think the term "initial" history makes sense for themes. For Puck it does because you need to initialize Puck with the initial state of the history and then Puck deals with the history stuff internally, so we need multiple variables for it. For theme, we're maintaining that state ourselves so we basically just need to load the themeHistory and then save it.

src/internal/puck/components/ThemeHeader.tsx
```diff
@@ -64,29 +64,43 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
...
         <ClearLocalChangesButton
-          disabled={themeHistory?.history.length === 1}
+          disabled={themeHistory?.histories?.length === 1}
           onClearLocalChanges={() => {
             clearThemeHistory();
             if (themeConfig) {
-              updateThemeInEditor(themeHistory?.history[0], themeConfig);
+              updateThemeInEditor(
+                themeHistory?.histories?.[0]?.state
+                  ?.data as unknown as SavedTheme,
```
mkilpatrick, line 73: updateThemeInEditor should use the same type as our theme state. We shouldn't need to cast like this if we do things right.

src/internal/puck/components/ThemeHeader.tsx
```diff
@@ -64,29 +64,43 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
...
             });
           }}
         />
         {!isDevMode && (
           <Button
             variant="secondary"
-            disabled={themeHistory?.history.length === 1}
+            disabled={themeHistory?.histories?.length === 1}
             onClick={async () => {
               await onPublishTheme();
```
mkilpatrick, line 92: I wonder if onPublishTheme should handle updating the themeHistory state too, rather than the button action doing it.

src/internal/puck/components/ThemeHeader.tsx
```diff
@@ -64,29 +64,43 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
...
               setThemeHistory({
-                history: [
-                  themeHistory?.history[themeHistory?.history.length - 1],
+                histories: [
+                  {
+                    id: "root",
```
mkilpatrick, line 96: I don't think we want to rest the id. This only made sense for Puck. We already have a generated id right? All we're doing here is publishing and then setting the themeHistories back to the last item, which should have a hash already from when it was generated.

src/internal/components/InternalThemeEditor.tsx
asanehisa: ah not this one cuz it should be themeHistory.histories.length -1 + 1 (for the new history)

src/internal/puck/components/ThemeHeader.tsx
asanehisa: maybe? in the LayoutHeader the state is updated in button action like this. Rn onPublishTheme is just 

  const handlePublishTheme = async () => {
    devLogger.logFunc("saveThemeData");
    publishThemeConfiguration({
      payload: {
        saveThemeData: JSON.stringify(
          themeHistory?.histories?.[themeHistory.index ?? 0]?.state?.data
        ),
      },
    });
  };

src/internal/components/InternalThemeEditor.tsx
```diff
@@ -21,8 +22,8 @@ type InternalThemeEditorProps = {
   publishThemeConfiguration: (data: any) => void;
   themeConfig?: ThemeConfig;
   saveThemeSaveState: (data: any) => void;
-  themeHistory?: ThemeSaveState;
-  setThemeHistory: (themeHistory: ThemeSaveState) => void;
+  themeHistory?: ThemeHistories;
```
mkilpatrick, line 25: Call this themeHistories to be consistent with the type. There is another type called ThemeHistory so then the variable is confusing since it's referring to something different.

src/internal/components/InternalThemeEditor.tsx
```diff
@@ -62,35 +63,39 @@ export const InternalThemeEditor = ({
...
+      JSON.stringify(newHistory.histories)
     );
 
     if (templateMetadata.isDevMode && !templateMetadata.devOverride) {
       devLogger.logFunc("sendDevThemeSaveStateData");
       sendDevThemeSaveStateData({
         payload: {
           devThemeSaveStateData: JSON.stringify(
-            newHistory.history[newHistory.index]
+            newHistory.histories[newHistory.index ?? 0]?.state?.data
```
mkilpatrick: index isn't optional so we shouldn't need the coalesce, same in other places

src/internal/types/themeData.ts
```diff
@@ -1,15 +1,18 @@
...
+  histories: ThemeHistory[];
+  index: number;
 };
 
-export type History<D = any> = {
-  state: D;
-  id?: string;
+export type ThemeHistory = {
+  state: {
+    data?: {
```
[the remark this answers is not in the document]
asanehisa: no doesn't need second level, was just matching Puck History

src/internal/components/InternalThemeEditor.tsx
```diff
@@ -51,27 +51,40 @@ export const InternalThemeEditor = ({
...
       },
     });
+
+    clearThemeHistory();
+
+    setThemeHistories({
+      histories: [
+        {
+          id: themeHistories?.histories[themeHistories.index]?.id ?? "",
+          data: themeHistories?.histories?.[themeHistories.index]?.data ?? {},
```
mkilpatrick, line 65: Turn this into an object that can be reused here and on line 54.
```
const currentThemeHistory = themeHistories?.histories[themeHistories.index];
```

I'm not sure what to do if currentThemeHistory is undefined, but can that ever actually happen?

asanehisa: Hmmm technically it seems like no? Since we require themeHistoryFetched to be true before getting to this page.

mkilpatrick (approved):

PR #93, opened 2024-10-17
fix: update puck history with new layout's config in theme mode

briantstephan:
While in theme mode, this sets the puck history to the updated `puckInitialHistory`, which is derived from the visualConfigurationData. This ensures that the editor updates accordingly when a different layout is selected.

Tested manually by running a local build of visual-editor with the starter. Changing the layout in the header made the editor update with the new layout after about half a second.


src/internal/components/InternalThemeEditor.tsx
```diff
@@ -45,6 +45,14 @@ export const InternalThemeEditor = ({
 }: InternalThemeEditorProps) => {
   const [canEdit, setCanEdit] = useState<boolean>(false); // helps sync puck preview and save state
 
+  const {
+    history: { setHistories },
+  } = usePuck();
```
benlife5, line 50: Does this work? I thought usePuck had to be inside a <Puck> component

briantstephan: You're right, just moved it to ThemeHeader and it works correctly now. I had forgotten to re-pack my changes before testing it in InternalThemeHeader.

benlife5 (approved):

mkilpatrick (approved):

PR #94, opened 2024-10-17
fix: sidebar loses focus after user input

benlife5:
The sidebar Puck override was re-mounting on every state change, effectively creating new fields each time and losing user state. This moves the themeHistory state lower in the hierarchy and maintains it separately from themeInitialHistory.

- number inputs now stay focused
- color inputs now stay open until you click outside them

### DIFF
--- a/src/internal/puck/components/ColorSelector.tsx
+++ b/src/internal/puck/components/ColorSelector.tsx
@@ -1,57 +1,39 @@
-import React, { useState, useCallback } from "react";
-import { Field, FieldLabel } from "@measured/puck";
+import React, { useState } from "react";
+import { FieldLabel } from "@measured/puck";
 import { RenderProps } from "../../utils/renderEntityFields.ts";
 import { Color, ColorResult, SketchPicker } from "react-color";
 
-export type ColorSelectorProps = {
-  label: string;
-};
+export const ColorSelector = ({ field, value, onChange }: RenderProps) => {
+  const [isOpen, setIsOpen] = useState(false);
 
-export const ColorSelector = (props: ColorSelectorProps): Field => {
-  const colorSelectorRenderer = useCallback(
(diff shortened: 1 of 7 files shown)


src/internal/puck/components/ColorSelector.tsx
```diff
@@ -8,10 +8,8 @@ export type ColorSelectorProps = {
 };
 
 export const ColorSelector = (props: ColorSelectorProps): Field => {
-  return {
-    type: "custom",
-    label: props.label,
-    render: ({ field, value, onChange }: RenderProps) => {
+  const colorSelectorRenderer = useCallback(
```
mkilpatrick, line 11: Got some test errors

benlife5:
Very likely going to conflict with Alexis's #91, will wait to that to merge first

src/internal/puck/components/ColorSelector.tsx
```diff
@@ -1,57 +1,39 @@
...
 
-export type ColorSelectorProps = {
-  label: string;
-};
-
-export const ColorSelector = (props: ColorSelectorProps): Field => {
-  const colorSelectorRenderer = useCallback(
-    ({ field, value, onChange }: RenderProps) => {
-      const [isOpen, setIsOpen] = useState(false);
+export const ColorSelector = ({ field, value, onChange }: RenderProps) => {
```
mkilpatrick, line 6: Does this change how someone needs to use the component now since it's returning something different? Ideally a user can just use it straight up in their config without needing to defined the 'custom' stuff themselves, though it's not the end of the world if that's the case.

benlife5: The user shouldn't ever need to use this component directly. When the user specifies `type: "color"` in the `theme.config.ts`, we render this component in the sidebar in `src/internal/utils/constructThemePuckFields.ts`

PR #99, opened 2024-10-22
fix: sidebar loses focus after input

benlife5:
- creates distinction between themeHistories and themeInitialHistories
  - themeInitialHistories is what we load from Content/DB/localstorage and is updated on Publish/Clear
  - themeHistories is what is current in the editor, is written to DB/localstorage, and sent to dev previews
- Freezes sidebar in useCallback 
  - maintains a separate themeHistoriesInternal state in parallel to themeHistories
  - reloads when themeInitialHistories updates


src/internal/puck/components/ThemeSidebar.tsx
```diff
@@ -6,16 +6,15 @@ import {
   constructThemePuckFields,
   constructThemePuckValues,
 } from "../../utils/constructThemePuckFields.ts";
-import { ThemeHistory } from "../../types/themeData.ts";
 
 type ThemeSidebarProps = {
   themeConfig?: ThemeConfig;
-  themeHistory: ThemeHistory[];
+  themeValues: Record<string, any>;
```
mkilpatrick, line 12: There's a type for this now.

src/internal/components/InternalThemeEditor.tsx
```diff
@@ -38,13 +37,31 @@ export const InternalThemeEditor = ({
...
-  themeHistories,
-  setThemeHistories,
+  themeInitialHistories,
+  setThemeInitialHistories,
   clearThemeHistory,
   sendDevThemeSaveStateData,
   buildThemeLocalStorageKey,
 }: InternalThemeEditorProps) => {
   const [canEdit, setCanEdit] = useState<boolean>(false); // helps sync puck preview and save state
+  const [themeHistories, setThemeHistories] = useState<
```
mkilpatrick, line 47: This seems more confusing to me. I commented something similar in Alexis' PR. For the non-theme flow we have initialHistories because that's what's passed to Puck to get the history (undo/redo) working. Puck also keeps the history state itself which we need to work with. That's ultimately why there are two things going on. For themes though I don't think we need both. We get the theme data from the DB/localStorage/Content. Can't we just setThemeHistories there and update themeHistories as a user makes changes? Why do we need to store the state of both things?

benlife5: The problem is the useCallback on line 135 in InternalThemeEditor. That is what avoids remounting the sidebar on every change, which is what was causing the focus issue. However, if it has an empty dependency array, then clicking Clear Local Changes doesn't update the sidebar, because it has no way of knowing that button was clicked. If the dependency is themeHistories (without tracking themeInitialHistories separately), then it remounts every change and is still broken. So, this triggers a remount on publish/clear via an update to the themeInitialHistories, but avoids it on regular updates to themeHistories.

I don't love this solution -- it definitely adds back in complexity that would be nice to avoid. I'm very open to other solutions, I just haven't found anything else that worked. I can take another look though

benlife5: [Posted on the Discord about the core problem](https://discord.com/channels/1153376562259951687/1297913620897857556/1298337516591517716)

PR #101, opened 2024-10-23
refactor: move useDocument, DocumentProvider from pages

asanehisa:


src/hooks/index.ts
```diff
@@ -4,3 +4,8 @@ export {
   EntityFieldsProvider,
   usePlatformBridgeEntityFields,
 } from "./useEntityFields.tsx";
+export {
+  DocumentProvider,
+  useDocument,
+  DocumentContext,
+} from "./useDocument.js";
```
jwartofsky-yext, line 11: should be .tsx I think

PR #102, opened 2024-10-23
feat: add Image component

jwartofsky-yext:
Adds an image component which can be used in the pages starter repo.

Note that to use this component, all instances of useDocument must be replaces with the VE version.


jwartofsky-yext:
> looks good, not sure if u included the useDoc code on purpose or not
> 
> Edit: Ah it prob happened accidentally, it'll be fine if we merge the useDoc branch first. Also in [mine](https://github.com/yext/visual-editor/pull/103/files#diff-b96dae4653c651669a3c7a95ddf01ed8823a967561b1a74951e2640b2bcbc174) I made a new folder, unsure if we wanna use the same components folder that like Editor and EntityField are in

I think we may want to have separate paths like you have, I started a discussion here:
https://yext.slack.com/archives/C02985ZF96Y/p1729706151848339

src/components/Image.tsx
```diff
@@ -0,0 +1,134 @@
...
+  aspectRatio,
+  rounded,
+}) => {
+  const document = useDocument();
+  const resolvedImage = resolveYextEntityField<ImageProps["image"]>(
+    document,
+    imageField
+  );
+
+  console.log(resolvedImage);
```
mkilpatrick, line 100: Remove

src/components/Image.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  return (
+    <EntityField displayName="Image" fieldId={imageField.field}>
+      <div
+        className={cn(
+          imageWrapperVariants({ size, aspectRatio }),
+          rounded && "overflow-hidden rounded-lg"
+        )}
+      >
+        <Image image={resolvedImage} className="w-full h-full object-cover" />
```
mkilpatrick, line 114: I think we might want to allow someone to set up a constant value which would allow them to give an http url. Probably other fields required by Image as well.

jwartofsky-yext: To do this I think we will need to support type.image as a constant value type

https://github.com/yext/visual-editor/blob/main/src/components/editor/YextEntityFieldSelector.tsx#L24

https://github.com/yext/visual-editor/pull/102/files#diff-2650600b809d4593f7d2a1efa547ef3e03a4583c3287603546655298cb4d8828R52

jwartofsky-yext: Going to add constant field support for images as a separate PR so we can close this out

benlife5 (approved):

PR #103, opened 2024-10-23
feat: add header text component

asanehisa:
Will put up another CR later w/ size as a number


src/atomic-components/headingText.tsx
```diff
@@ -0,0 +1,61 @@
...
+          {resolveYextEntityField(document, text)}
+        </Heading>
+      </EntityField>
+    );
+  }
+);
+
+HeadingText.displayName = "HeadingText";
+
+const headingTextFields: Fields<HeadingTextProps> = {
```
benlife5, line 32: I think we want the 4 properties listed in the doc under Theme Styling -> Heading Text to be props here too. Allows the user to set the globally or at the component level

src/atomic-components/headingText.tsx
benlife5, line 1: I think we usually do PascalCase filenames for these

src/atomic-components/index.ts
```diff
@@ -0,0 +1,2 @@
+export { HeadingTextComponent } from "./headingText.js";
+export type { HeadingTextProps } from "./headingText.js";
```
benlife5, line 2: If you add `"./components: "./dist/atomic-components"` to package.json, does `import {HeadingTextComponent} from "@yext/visual-editor/atomic-components"` work in the starter?

benlife5 (commented):
Nice!

src/atomic-components/headingText.tsx
jwartofsky-yext: Aaron had camelCase in the example. 

https://github.com/YextSolutions/pages-visual-editor-starter/tree/aaron/theme-demo/src/components/atoms

Maybe it's something we do for the atomic components?

src/atomic-components/index.ts
mkilpatrick: It should but that wouldn't be the path we want. I think all atomic and macro should just be at /components. The editor ones can still live at root.

src/atomic-components/headingText.tsx
asanehisa: Hmm I'll do PascalCase

src/atomic-components/headingText.tsx
asanehisa: Oh hmmm I was looking at Components -> Heading Text for the two props settable by users. 

I was thinking Theme Styling -> Heading Text was the theme props?

Might be misunderstanding

benlife5: Yeah sorry that's what I had thought at first but then Aaron & I talked about this some last Friday and I guess it didn't make it into the doc. [Here's](https://yext.slack.com/archives/C06A06BCUUF/p1729524337505309?thread_ts=1729285727.447869&cid=C06A06BCUUF) him talking about it for another component. 

Basically, Theme Manager will set a default styles for each but then the component can override. You already have that for color and size. (Also text transform doesn't work in TM because of tailwind stuff :/ so it has to be on the component level)

asanehisa: Ah gotcha

src/components/puck/atoms/heading.tsx
```diff
@@ -13,6 +13,10 @@ const headingVariants = cva("font-bold", {
       5: "font-heading5 text-heading5-fontSize text-heading5-color",
       6: "font-heading6 text-heading6-fontSize text-heading6-color",
     },
+    weight: {
+      default: "font-normal",
```
benlife5, line 17: I think maybe do
default: ""
normal: "font-normal"
bold: "font-bold"

Otherwise, LGTM

jwartofsky-yext (approved):

benlife5 (approved):

benlife5 (approved):

jwartofsky-yext (approved):

PR #105, opened 2024-10-23
refactor: move editor components to components/editor

jwartofsky-yext:
Create folders for components/puck and components/puck/atoms

Tested on dev account

Layout and theme manager work as expected


benlife5:
Can you test importing something in the starter from the src/components/puck folder? I think we need to add a new path to the export section of package.json

jwartofsky-yext:
> Can you test importing something in the starter from the src/components/puck folder? I think we need to add a new path to the export section of package.json

With the current setup, as long as the index.ts in src connects all the way down to /puck and /puck/atoms they should all be importable through import { * } from @yext/visual-editor

I can change it so that we have a unique export @yext/visual-editor/components which exports only what is under src/components/puck

jwartofsky-yext:
replaced with #106

PR #112, opened 2024-10-24
fix: header component

asanehisa:
fixed some tailwind not working due to lack of "ve-" prefix
changed size from select to number


jwartofsky-yext (approved):
lgtm

jwartofsky-yext:
Can you add explanation for what these changes do?

PR #115, opened 2024-10-25
wip: fix: published pages use the wrong layout

benlife5:
We weren't checking the site id on the layout so it would pick the first one regardless of site id.

This fixes production sites. However, it breaks dev mode because the dev mode document has siteId: 0. Not really sure how to handle this case because defaulting to the first layout doesn't really make sense.


benlife5:
Fixed by https://github.com/yext/visual-editor/pull/137

PR #117, opened 2024-10-28
fix: formatting of entity fields when used in devMode 

asanehisa:
Fixes the problem where the schemaFields in devMode were not formatted correctly and thus not always showing up as options in the EntityFieldSelector


benlife5 (approved):
Thanks for figuring this out!

PR #118, opened 2024-10-28
feat: update getFilteredEntityFields to return lists

briantstephan:
This updates `getFilteredEntityFields` to accept a new `includeListsOnly` field in the filter. When this field is not provided or set to `false`, the behavior is the same as before, and list fields are filtered out. If it is set to `true`, then list fields are returned and non-list fields are filtered out.


benlife5 (approved):
Nice!

PR #119, opened 2024-10-28
feat: support address constant values

benlife5:
Screenshot 2024-10-28 at 2 11 34 PM


asanehisa (approved):
has Image code too, assuming invalid address doesn't cause any problems! (since we are only showing it on sites as a string (not like a link or map)

jwartofsky-yext (approved):
lgtm

PR #124, opened 2024-10-30
feat: use theme and visualConfig from entityDocument

jwartofsky-yext:
We were already getting the entityDocument and therefore don't need separate calls to getVisualConfigurationData or getThemeData.

This also fixes the desync where we are waiting for the page document to update.

Themes update in layout editor almost immediately now.

Theme manager still works as expected.

### DIFF
--- a/src/internal/utils/parseConfigsFromDocument.ts
+++ b/src/internal/utils/parseConfigsFromDocument.ts
@@ -1,24 +1,43 @@
+import { ThemeData } from "../types/themeData.ts";
+import { Data } from "@measured/puck";
+import { resolveVisualEditorData } from "../../utils/resolveVisualEditorData.ts";
 import { jsonFromEscapedJsonString } from "./jsonFromEscapedJsonString.ts";
 
-type documentConfigs = {
-  visualConfigurationData?: any;
-  themeData?: any;
+type VisualEditorData = {
+  visualConfigurationData?: Data;
+  themeData?: ThemeData;
 };
 
-export const parseConfigsFromDocument = (document: any): documentConfigs => {
+export const parseConfigsFromDocument = (
+  document: any,
+  pageSet: string,
+  siteId: number
+): VisualEditorData => {
   if (document) {
-    const themeString =
-      document?._site?.pagesTheme?.[0]?.themeConfiguration?.data;
-    const visualConfigString =
-      document?.pageLayouts?.[0]?.visualConfiguration?.data;
+    const resolvedData = resolveVisualEditorData(
+      { document: document },
+      pageSet,
+      siteId
+    );
+    const visualConfigurationData = resolvedData.document.visualTemplate;
+    const themes: Record<string, any>[] =
+      resolvedData.document._site.pagesTheme;
+
+    let themeDataString: string = "{}";
+    if (themes?.length > 0) {
+      themeDataString =
+        siteId !== 0
+          ? (themes.find(
+              (theme) => Number(theme.themeConfiguration.siteId) === siteId
+            )?.themeConfiguration.data ?? "{}")
+          : themes[0].themeConfiguration.data; // siteId is not set on local data documents, so default to the first one
+    }
+
+    const themeData = jsonFromEscapedJsonString(themeDataString);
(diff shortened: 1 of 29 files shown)


src/internal/utils/parseConfigsFromDocument.ts
```diff
@@ -0,0 +1,25 @@
...
+
+type documentConfigs = {
+  visualConfigurationData?: any;
+  themeData?: any;
+};
+
+export const parseConfigsFromDocument = (document: any): documentConfigs => {
+  if (document) {
+    const themeString =
+      document?._site?.pagesTheme?.[0]?.themeConfiguration?.data;
```
benlife5, line 11: Does this need to check that the site id on the theme and layout match the site id? Also need a way to handle dev mode but that's part of [SUMO-6628](https://yexttest.atlassian.net/jira/software/c/projects/SUMO/boards/97?selectedIssue=SUMO-6628)

benlife5 (commented):
Does switching layouts/entities still work? I think it should but just want to check since those message receivers were involved

src/internal/utils/parseConfigsFromDocument.ts
jwartofsky-yext: Fixed the issue, prod mode works as expected now.

src/internal/utils/parseConfigsFromDocument.ts
```diff
@@ -0,0 +1,25 @@
+import { jsonFromEscapedJsonString } from "./jsonFromEscapedJsonString.ts";
+
+type documentConfigs = {
+  visualConfigurationData?: any;
```
mkilpatrick, line 4: Type these. You can see the types from the PR I just merged.

src/internal/utils/parseConfigsFromDocument.ts
```diff
@@ -0,0 +1,25 @@
...
+  themeData?: any;
+};
+
+export const parseConfigsFromDocument = (document: any): documentConfigs => {
+  if (document) {
+    const themeString =
+      document?._site?.pagesTheme?.[0]?.themeConfiguration?.data;
+    const visualConfigString =
+      document?.pageLayouts?.[0]?.visualConfiguration?.data;
```
mkilpatrick, line 13: This needs to use the fallback logic of site > layout > entity.

src/internal/utils/parseConfigsFromDocument.ts
```diff
@@ -0,0 +1,25 @@
+import { jsonFromEscapedJsonString } from "./jsonFromEscapedJsonString.ts";
+
+type documentConfigs = {
```
mkilpatrick, line 3: Capitalize and rename to `VisualEditorData`

src/internal/utils/parseConfigsFromDocument.ts
benlife5: This works even if the site entity has multiple themes/layouts for different site ids?

jwartofsky-yext: Updated to use resolveVisualEditorData and it does now

src/internal/utils/parseConfigsFromDocument.ts
```diff
@@ -0,0 +1,44 @@
...
+    const resolvedData = resolveVisualEditorData(
+      { document: document },
+      pageSet,
+      siteId
+    );
+    const visualConfigurationData = resolvedData.document.visualTemplate;
+    const themes: Record<string, any>[] =
+      resolvedData.document._site.pagesTheme;
+
+    let themeDataString: string = "{}";
```
asanehisa, line 26: This is interesting? idk if it'd be cleaner to do return 
      visualConfigurationData: visualConfigurationData,
      themeData:{},
 if the themes.length = 0.
 
but also can ignore this, overall lgtm

jwartofsky-yext:
Themes are not updating properly after merging the latest

Investigating..

jwartofsky-yext:
Replaced with: https://github.com/yext/visual-editor/pull/130

PR #127, opened 2024-10-31
feat: correctly support constant values for text lists

briantstephan:
This change allows for a list of constant text values to be assigned to a TextList component. Deleting values, creating new values, and editing existing ones all work correctly.

Screenshot 2024-10-31 at 3 34 28 PM

### DIFF
--- a/src/internal/puck/constant-value-fields/TextList.tsx
+++ b/src/internal/puck/constant-value-fields/TextList.tsx
@@ -3,6 +3,8 @@ import { AutoField, IconButton, Button, CustomField } from "@measured/puck";
 import { Trash2 as TrashIcon } from "lucide-react";
 import { Plus as PlusIcon } from "lucide-react";
 
+const TEXT_LIST_BUTTON_COLOR: string = "#969696";
+
 export const TEXT_LIST_CONSTANT_CONFIG: CustomField<string[]> = {
   type: "custom",
   render: ({ onChange, value }) => {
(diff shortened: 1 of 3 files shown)


src/internal/puck/constant-value-fields/TextList.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { Trash2 as TrashIcon } from "lucide-react";
+import { Plus as PlusIcon } from "lucide-react";
+
+export const TEXT_LIST_CONSTANT_CONFIG: CustomField<string[]> = {
+  type: "custom",
+  render: ({ onChange, value }) => {
+    const [localItems, setLocalItems] = useState<string[]>(value);
+
+    const updateItem = (index: number, value: string) => {
+      const updatedItems = [...localItems];
```
mkilpatrick, line 12: This does a shallow clone so if it's a list of objects then it wouldn't work (would shallow clone level one but use a reference to the object deeper, in which case structuredClone is the answer), but these are strings so we should be ok.

src/internal/puck/constant-value-fields/TextList.tsx
```diff
@@ -0,0 +1,70 @@
...
+              marginBottom: "8px",
+              gap: "4px",
+            }}
+          >
+            <AutoField
+              field={{ type: "text" }}
+              value={item}
+              onChange={(itemValue) => updateItem(index, itemValue)}
+            />
+            <span style={{ color: "#969696" }}>
```
mkilpatrick, line 45: Make the color value a const

src/internal/puck/constant-value-fields/TextList.tsx
briantstephan: Yeah it should be fine for now. If we use this for other list types in the future we can update it then.

mkilpatrick (approved):

PR #128, opened 2024-10-31
feat: add visual editor components

benlife5:
- copied from components branch of the starter
- updated to use new theme.config format

Some components were using `twMerge` to merge their `classNames` prop with the `cva` class names. This causes some problems for theme manger styles since classes we have overloaded utilities (ex `text-body-fontWeight` and `text-body-color`). Switched to `clsx` in those places which just does a join so that they both are applied. Will look at `twMerge` configuration to see if we can still do a smarter merge if I have time later

https://github.com/YextSolutions/pages-visual-editor-starter/pull/160


jwartofsky-yext (dismissed):

src/components/puck/index.ts
```diff
@@ -1,4 +1,16 @@
 export * from "./atoms/index.ts";
+export { AddressComponent, type AddressProps } from "./Address.tsx";
```
mkilpatrick, line 2: I think we should have an index.ts in the puck folder and then export everything from that file here like above

jwartofsky-yext (approved):

mkilpatrick (approved):

PR #133, opened 2024-11-04
feat: add NumberOrDefaultField

benlife5:
Used in the components to set font size, should work for other number values like padding or border radius if wanted

 

https://github.com/user-attachments/assets/a0c4e31e-b03b-429b-a997-077455e0f5c3

### DIFF
--- a/src/components/editor/NumberOrDefaultField.tsx
+++ b/src/components/editor/NumberOrDefaultField.tsx
@@ -1,0 +1,49 @@
+import React from "react";
+import { CustomField, FieldLabel, AutoField } from "@measured/puck";
+
+export type NumberOrDefault = number | "default";
+
+export type NumberFieldWithDefaultOptionProps = {
+  label: string;
+  defaultCustomValue: number;
+};
+
+type NumberFieldWithDefaultOptionRenderProps = {
+  value: NumberOrDefault;
+  onChange: (e: NumberOrDefault) => void;
+};
+
+export const NumberFieldWithDefaultOption = ({
+  label,
+  defaultCustomValue,
+}: NumberFieldWithDefaultOptionProps): CustomField<NumberOrDefault> => {
+  return {
+    type: "custom",
+    render: ({ onChange, value }: NumberFieldWithDefaultOptionRenderProps) => {
+      return (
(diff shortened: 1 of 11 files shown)


src/components/puck/HeadingText.tsx
```diff
@@ -8,6 +8,7 @@ import {
   YextEntityField,
   YextEntityFieldSelector,
 } from "../editor/YextEntityFieldSelector.js";
+import { NumberFieldWithDefaultOption } from "../editor/NumberOrDefaultField.js";
```
mkilpatrick, line 11: This has .js import but others you changed to .tsx. Which do we prefer? Does .tsx work?

benlife5: Ah good catch. The VSCode auto import uses .js but we do .tsx everywhere else

mkilpatrick (approved):

src/components/editor/NumberOrDefaultField.tsx
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import { CustomField, FieldLabel, AutoField } from "@measured/puck";
+
+export type NumberOrDefault = number | "default";
+
+export type NumberFieldWithDefaultOptionProps = {
+  label: string;
+  defaultCustomValue: number;
```
jwartofsky-yext, line 8: why defaultCustomValue and not default?

benlife5: This field is supposed to be the value that is in the number input by default when you toggle to custom mode.
default/defaultValue seemed confusing since "default" represents the other mode where it just uses the value from the theme.
But yeah I found it tricky to name a lot of these variables names -- open to suggestions if you have them

mkilpatrick: If we want two separate defaults I think `default` and `defaultCustomValue` makes sense. Feel free to get Aaron's opinion on it.

benlife5: Checked with Aaron, he said this is good for now and hopefully we'll get some user feedback soon-ish and can adjust as needed

PR #134, opened 2024-11-04
feat: use localStorage when in dev mode

jwartofsky-yext:
When in devMode, LayoutEditor uses the theme from local storage.

When in devMode, ThemeEditor uses the layout from local storage.

Tested alongside changes to YextSitesStorm.
https://gerrit.yext.com/c/alpha/+/275033

Verified that when layout or theme is updated in dev mode, the change can be seen in the other mode.


src/internal/components/LayoutEditor.tsx
```diff
@@ -63,7 +65,26 @@ export const LayoutEditor = (props: LayoutEditorProps) => {
    */
   useEffect(() => {
     devLogger.logData("THEME_DATA", themeData);
```
mkilpatrick, line 67: Log the data actually being used, aka have a logger for the localStorage data and move this to the if themeData conditional.

src/utils/devLogger.ts
```diff
@@ -3,7 +3,9 @@ export type DevLoggerPrefix =
   | "PUCK_CONFIG"
   | "SAVE_STATE"
   | "THEME_SAVE_STATE"
+  | "LOCAL_VISUAL_CONFIGURATION_DATA"
```
mkilpatrick, line 6: We don't use this key when we use the local data in layout mode. We instead use VISUAL_CONFIGURATION_DATA with another log saying it's using localStorage. We should be consistent.

src/internal/components/ThemeEditor.tsx
```diff
@@ -74,10 +74,7 @@ export const ThemeEditor = (props: ThemeEditorProps) => {
       const localHistories = JSON.parse(localHistoryArray) as History[];
       const localHistoryIndex = localHistories.length - 1;
       if (localHistoryIndex >= 0) {
-        devLogger.logData(
-          "LOCAL_VISUAL_CONFIGURATION_DATA",
-          localHistories[localHistoryIndex].state
-        );
+        devLogger.log("Theme Dev Mode - Using layout data from local storage");
```
mkilpatrick, line 77: We still want to log the data though. Are we doing that?

jwartofsky-yext: Whenever we receive themeData or visualConfigurationData it gets logged by the receiver.

We don't log in this specific case where we use the localStorage though, but there is a message to let the user know that it is using localStorage and they can access the message in console log if they wish.

mkilpatrick (approved):

mkilpatrick (approved):

PR #136, opened 2024-11-04
fix: layout editor no longer requires themeId

jwartofsky-yext:
If themeId is not present, it will simply not try to apply the theme from local storage.

### DIFF
--- a/src/components/editor/Editor.tsx
+++ b/src/components/editor/Editor.tsx
@@ -95,7 +95,7 @@ export const Editor = ({
     !templateMetadata ||
     !document ||
     !visualConfigurationDataFetched ||
-    !themeDataFetched;
+    (templateMetadata?.themeEntityId && !themeDataFetched);
 
   const progress: number =
     60 * // @ts-expect-error adding bools is fine
@@ -103,7 +103,7 @@ export const Editor = ({
       !!templateMetadata +
       !!document +
       visualConfigurationDataFetched +
-      themeDataFetched) /
+      (themeDataFetched || !templateMetadata?.themeEntityId)) /
       5);
 
   return (
--- a/src/components/editor/Editor.tsx
+++ b/src/components/editor/Editor.tsx
@@ -95,7 +95,7 @@ export const Editor = ({
     !templateMetadata ||
     !document ||
     !visualConfigurationDataFetched ||
-    (templateMetadata?.themeEntityId && !themeDataFetched);
+    !themeDataFetched;
 
   const progress: number =
     60 * // @ts-expect-error adding bools is fine
@@ -103,7 +103,7 @@ export const Editor = ({
       !!templateMetadata +
       !!document +
       visualConfigurationDataFetched +
-      (themeDataFetched || !templateMetadata?.themeEntityId)) /
+      themeDataFetched) /
       5);
 
   return (
--- a/src/internal/components/LayoutEditor.tsx
+++ b/src/internal/components/LayoutEditor.tsx
@@ -64,7 +64,7 @@ export const LayoutEditor = (props: LayoutEditorProps) => {
    * Apply the themes from Content
    */
   useEffect(() => {
-    if (!themeConfig || !templateMetadata?.themeEntityId) {
+    if (!themeConfig) {
       return;
     }
     const localHistoryArray = window.localStorage.getItem(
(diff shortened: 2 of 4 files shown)


src/internal/components/LayoutEditor.tsx
```diff
@@ -64,7 +64,7 @@ export const LayoutEditor = (props: LayoutEditorProps) => {
    * Apply the themes from Content
    */
   useEffect(() => {
-    if (!themeConfig) {
+    if (!themeConfig || !templateMetadata?.themeEntityId) {
```
mkilpatrick, line 67: The dep array should have templateMetadata on it

src/components/editor/Editor.tsx
```diff
@@ -95,15 +95,15 @@ export const Editor = ({
     !templateMetadata ||
     !document ||
     !visualConfigurationDataFetched ||
-    !themeDataFetched;
+    (templateMetadata?.themeEntityId && !themeDataFetched);
```
mkilpatrick, line 98: I still feel like we should set themeDataFetched to true even if there's no theme data to fetch so we don't have to check multiple things in multiple places. This makes it more confusing IMO because it requires us to understand more about the intricacies of how the theme stuff works. As a counter example, we don't check multiple things for vcData, we just check that it was fetched.

src/internal/hooks/useMessageReceivers.ts
```diff
@@ -31,7 +31,13 @@ export const useCommonMessageReceivers = (
 
   // Theme from Content
   const [themeData, setThemeData] = useState<ThemeData>();
-  const [themeDataFetched, setThemeDataFetched] = useState<boolean>(false); // needed because themeData can be empty
+  const [themeDataFetched, setThemeDataFetched] = useState<boolean>(
+    !templateMetadata?.themeEntityId
+  ); // needed because themeData can be empty
+
+  useEffect(() => {
+    setThemeDataFetched(!templateMetadata?.themeEntityId);
```
mkilpatrick, line 39: Isn't this already handled on line 74?

src/internal/components/LayoutEditor.tsx
mkilpatrick, line 67: I don't think this is the appropriate spot for this check. If anything the check should happen in useThemeLocalStorage which would return "" and is already handled by line 74.

src/internal/hooks/useMessageReceivers.ts
jwartofsky-yext: not exactly

if themeEntityId doesn't exist, then we want to setThemeDataFetched to true

the call to getThemeData will not occur if themeEntityId does not exist

There's also the situation where the page is loading
ThemeEntityId is 0
We setThemeDataFetched to true (because themeEntityId is 0)
ThemeEntityId updates to an actual id
We now want to have themeDataFetched set to false

This useEffect handles this case because whenever themeEntityId changes we update the themeDataFetched boolean

mkilpatrick: I looked into why we don't need to do this for the veData and it's because we always make the query. I think we should do the same for theme to be consistent. So in ThemeEditorMessenger.tsx (alpha) we remove the themeEntityId check on `enabled: iFrameLoaded > 0 && !!themeEntityId,` and if themeEntityId is nullish then we just return nothing. That way on this side themeDataFetched will just get set without all of this extra logic.

jwartofsky-yext: https://gerrit.yext.com/c/alpha/+/275129

src/internal/hooks/useMessageReceivers.ts
```diff
@@ -62,7 +62,10 @@ export const useCommonMessageReceivers = (
   );
 
   useReceiveMessage("getThemeData", TARGET_ORIGINS, (send, payload) => {
-    const themeData = jsonFromEscapedJsonString(payload as unknown as string);
+    const payloadString = payload as unknown as string;
+    const themeData = payloadString
```
mkilpatrick, line 66: Does no content return nothing, opposed to an empty object?

jwartofsky-yext: Yes it returns nothing

This was needed in order to handle NoContent

asanehisa:
oh failing tests?

jwartofsky-yext:
> oh failing tests?

those aren't related to this PR

it's missing fontSize in GetDirections

mkilpatrick (approved):

asanehisa (approved):

PR #137, opened 2024-11-04
feat: handle siteId in resolveVisualEditorData

jwartofsky-yext:
If there is no siteId on the document then siteId is ignored.


benlife5:
I think this still doesn't support dev mode with multiple sites using the same site entity. If siteId == 0, it'll pass through that conditional statement, but then it'll pick whichever layout matches the template name first

benlife5:
> Yes, it's not the full fix yet, but this handles prod mode without changing dev mode.

Ah ok sorry, that makes sense

mkilpatrick (approved):

PR #138, opened 2024-11-05
feat: add shadcn cli registry

benlife5:
https://github.com/YextSolutions/pages-visual-editor-starter/pull/164

Usage:
`REGISTRY_URL=https://reliably-numerous-kit.pgsdemo.com/components npx shadcn add`
`HeaderText` and `BodyText` are currently available


src/components/puck/registry/components.ts
```diff
@@ -0,0 +1,17 @@
...
+  {
+    name: "heading",
+    type: "registry:component",
+    files: ["atoms/heading.tsx"],
+  },
+  {
+    name: "HeadingText",
+    type: "registry:component",
+    registryDependencies: [
+      "https://reliably-numerous-kit.pgsdemo.com/components/heading.json",
```
mkilpatrick, line 13: Make the url a const so we can update it in one spot when we change it later.

src/components/puck/registry/build-registry.ts
```diff
@@ -0,0 +1,101 @@
...
+      null,
+      2
+    );
+    const jsonPath = `${PUBLIC_FOLDER_BASE_PATH}/${component.name}.json`;
+    await writeFileRecursive(jsonPath, json);
+
+    // import new json file in index.ts
+    await fs.appendFile(
+      PUBLIC_FOLDER_INDEX_PATH,
+      `import { default as ${component.name} } from "./${component.name}.json"\n`,
```
mkilpatrick, line 82: Is there a way to to make it work like:
`npx shadcn add https://reliably-numerous-kit.pgsdemo.com/components/fullList.json HeadingText`

I see in their example it's what you have, but it would be nice to point to a single entrypoint and name the component, rather than a different url per component. I assume the answer is no. It's worth seeing if there are any issues for something like that and if not making one.

benlife5: Switched around the serving repo so now
`REGISTRY_URL=https://reliably-numerous-kit.pgsdemo.com/components npx shadcn add HeadingText` works

src/components/puck/registry/components.ts
benlife5: Ack, with the new changes it can just be `heading`

benlife5:
> I think this seems like something users will have to be able to add to their repos if they don't use our starter, in that case, would some basic README / documentation be helpful to explain how to implement ?

Yeah, for sure. I think we may want to wait for this to solidify before writing docs thought because it's still experimental

src/components/puck/registry/build-registry.ts
mkilpatrick: Awesome 🔥

mkilpatrick: What defines the url path or filename now?

benlife5: It's `REGISTRY_URL/styles/STYLE/NAME.json` with NAME being what you put after add and STYLE being read from  the `components.json` file in the destination repo (this is the structure of pulling normal shadcn components). I think I need to TAL a little more at how much pulling our components would block you from pulling shadcn components and vice versa

mkilpatrick: Well pulling shadcn components you just wouldn't defined a registry-url, right? So you're saying NAME.json is coming from the fact that the file is named components.json? Was just looking to know how components gets set as the filename for https://reliably-numerous-kit.pgsdemo.com/components.

And the domain doesn't matter in this package then now, right? Because the user just puts in whatever the registry url is.

mkilpatrick: Realizing now that you export the registry from this package and import it in another [repo](https://github.com/yext-pages/visual_editor_library_4174974/blob/master/index.js) which builds it. Let's have this repo do that directly instead. We won't need to export the registry anymore, but instead have a registry build cmd in package.json that generates the json files into `dist` and then have a config.yaml and artifacts.json. The build step would need to copy artifacts.json to dist as well.

Note that there's two separate things. The first is a normal build for the lib which goes to dist. But to host the registry files we'd have a separate build command which also goes to dist. Both build commands shouldn't ever need to be run at the same time since there are two separate purposes.

benlife5:
This PR now has this repo directly powering the registry host at https://www.yext.com/s/4174974/yextsites/151161/branch/116486/deploys/recent

`REGISTRY_URL=https://reliably-numerous-kit.pgsdemo.com/components npx shadcn add` also lists available components

Going to take a look at `init` now

.nvmrc
```diff
@@ -0,0 +1 @@
+v18.20.4
```
mkilpatrick, line 1: Is this file necessary?

config.yaml
```diff
@@ -0,0 +1,10 @@
+buildConfiguration:
+  buildCommand: cd src/components/puck/registry && npm run generate
```
mkilpatrick, line 2: Let's continue to use pnpm instead of npm.

src/components/puck/atoms/heading.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from "react";
 import { cva, type VariantProps } from "class-variance-authority";
-import { type NumberOrDefault } from "../../editor/NumberOrDefaultField.tsx";
+import { type NumberOrDefault } from "../../editor/NumberOrDefaultField.js";
```
mkilpatrick, line 3: Regression back to .js? But I'm assuming it's related to the requirements for components.

src/components/puck/registry/build-registry.ts
```diff
@@ -1,15 +1,41 @@
-// Based on https://github.com/shadcn-ui/ui/blob/main/apps/www/scripts/build-registry.mts
+// Based on https://github.com/bwestwood11/ui-cart/blob/main/scripts/build-registry.ts
+// which is a simplified version of https://github.com/shadcn-ui/ui/blob/main/apps/www/scripts/build-registry.mts
+import fs from "fs";
```
mkilpatrick, line 3: Use `from "node:fs/promises"` and then you don't need to do fs.promises below

src/components/puck/registry/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "visual-editor-component-registry",
+  "version": "0.0.0",
+  "description": "Created by Yext",
```
mkilpatrick, line 4: A component library registry for visual-editor components to work with the ShadCN CLI

tailwind.config.ts
```diff
@@ -1,7 +1,12 @@
 import type { Config } from "tailwindcss";
 
 export default {
-  content: ["./src/**/*.{html,js,jsx,ts,tsx}"],
+  content: [
+    "./src/internal/**/*.{html,js,jsx,ts,tsx}",
```
mkilpatrick, line 5: Is this to exclude registry? Can you glob all in src but exclude registry?

src/components/puck/atoms/heading.tsx
benlife5: The starter's tsconfig doesn't allow ts imports -- can update that and keep these as .tsx if we want

config.yaml
benlife5: Is there a way to have the build step in the platform use pnpm?

.nvmrc
benlife5: Will replace with an env var in the platform

src/components/puck/registry/README.md
```diff
@@ -20,3 +20,13 @@ Add a new object to `ui` in `components.ts`.
 - Do not use .ts or .tsx imports
 - Imports from the @yext/visual-editor package must have a local path that is captured by the
   `IMPORT_PATTERN` in `build-registry.ts`
+
+## Running shadcn locally
+
+Useful for debugging the our registry with the shadcn CLI.
```
mkilpatrick, line 26: Phrasing off

src/components/puck/atoms/heading.tsx
mkilpatrick: Nah this is ok

config.yaml
mkilpatrick: https://yext.slack.com/archives/CC7B2751A/p1731077760800839

benlife5: switched

mkilpatrick (dismissed):

mkilpatrick (approved):

PR #142, opened 2024-11-07
fix: adjust components' exported names

asanehisa:


benlife5:
I think we should probably swap these in each component file so that it's less confusing (bc rn there is Address.tsx has Address and AddressComponent, but then we import AddressComponent and rename it to Address)

asanehisa:
Actually I think we can ignore this item til after Demos, can potentially throw stuff off

asanehisa:
Sorry about the weird pushing, but tested + the renamed exports all work in the starter!

mkilpatrick (approved):

benlife5 (approved):

PR #144, opened 2024-11-07
fix: adjust email padding

asanehisa:
Screenshot 2024-11-07 at 5 19 28 PM


benlife5 (approved):
LGTM, I think we should probably eventually add padding as a prop to all of these components because it's highly situational when you want it vs. not, but this is good for now

PR #147, opened 2024-11-11
fix: cannot access before initialization error

benlife5:
The components were exported before their dependencies, which caused vite to throw an error in the starter.

Also removed `typeof config` from Address.tsx, which was a bad autocompleted import.


asanehisa (approved):

mkilpatrick (approved):

jwartofsky-yext (approved):
lgtm

PR #150, opened 2024-11-11
feat: add constant value support for CTA

jwartofsky-yext:
Constant value includes name and url fields

Tested on dev account

Verified that constant values can be set properly

### DIFF
--- a/src/components/editor/YextEntityFieldSelector.tsx
+++ b/src/components/editor/YextEntityFieldSelector.tsx
@@ -13,6 +13,7 @@ import { IMAGE_CONSTANT_CONFIG } from "../../internal/puck/constant-value-fields
 import { TEXT_CONSTANT_CONFIG } from "../../internal/puck/constant-value-fields/Text.tsx";
 import { ADDRESS_CONSTANT_CONFIG } from "../../internal/puck/constant-value-fields/Address.tsx";
 import { TEXT_LIST_CONSTANT_CONFIG } from "../../internal/puck/constant-value-fields/TextList.tsx";
+import { CTA_CONSTANT_CONFIG } from "../../internal/puck/constant-value-fields/CallToAction.tsx";
 
 const devLogger = new DevLogger();
 
@@ -33,6 +34,7 @@ const TYPE_TO_CONSTANT_CONFIG: Record<string, Field<any>> = {
   "type.phone": TEXT_CONSTANT_CONFIG,
   "type.image": IMAGE_CONSTANT_CONFIG,
   "type.address": ADDRESS_CONSTANT_CONFIG,
+  c_cta: CTA_CONSTANT_CONFIG,
 };
 
 const getConstantConfigFromType = (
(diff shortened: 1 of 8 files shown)


src/internal/puck/constant-value-fields/CallToAction.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import { CustomField, FieldLabel, AutoField } from "@measured/puck";
+import { CTAProps } from "../../../components/puck/atoms/cta.tsx";
+
+export const CTA_CONSTANT_CONFIG: CustomField<CTAProps> = {
+  type: "custom",
+  render: ({ onChange, value }) => {
+    return (
+      <>
+        <FieldLabel label={"Label"} className="ve-inline-block ve-pt-4">
```
mkilpatrick, line 10: All of the const-value-fields are pretty repetitive. Maybe we should simplify them to just iterate over an object that defines the label and property name, then all of these components would basically be one-liners.

src/components/puck/CtaWrapper.tsx
```diff
@@ -78,9 +78,9 @@ export const CTAWrapperComponent: ComponentConfig<CTAWrapperProps> = {
     entityField: {
       field: "",
       constantValue: {
-        label: "Call to Action",
+        name: "Call to Action",
```
benlife5, line 81: Not you but the this shouldn't be variant and fontSize, it should be label and link

src/components/editor/YextEntityFieldSelector.tsx
```diff
@@ -33,6 +34,7 @@ const TYPE_TO_CONSTANT_CONFIG: Record<string, Field<any>> = {
   "type.phone": TEXT_CONSTANT_CONFIG,
   "type.image": IMAGE_CONSTANT_CONFIG,
   "type.address": ADDRESS_CONSTANT_CONFIG,
+  c_cta: CTA_CONSTANT_CONFIG,
```
benlife5, line 37: Also not you, but the starter has a custom CTA type which is link and label, but I think we're supposed to be supporting the built-in Call To Action type which is link and label, but link has a type of URL/Phone/Email/Other

src/internal/puck/constant-value-fields/CallToAction.tsx
jwartofsky-yext: some of these are more complicated

such as TextList which has buttons

or Text which just uses the TextField type

We could create a shared helper method for the others though

src/components/editor/YextEntityFieldSelector.tsx
jwartofsky-yext: I'll update to use the built-in type in another PR

benlife5: Sounds good. I'm not sure we want to hardcode anything for c_cta though because you could create your own custom type called c_cta and it be completely different

jwartofsky-yext: Updated to use type.cta

benlife5 (approved):

asanehisa (approved):

mkilpatrick (approved):

PR #151, opened 2024-11-11
feat: add additional button tw from theme 

asanehisa:
With this [PR](https://github.com/YextSolutions/pages-visual-editor-starter/pull/170)

### DIFF
--- a/src/components/puck/atoms/button.tsx
+++ b/src/components/puck/atoms/button.tsx
@@ -5,10 +5,12 @@ import { clsx } from "clsx";
 import { NumberOrDefault } from "../../editor/NumberOrDefaultField.tsx";
 
 const buttonVariants = cva(
-  "py-4 components inline-flex items-center justify-center whitespace-nowrap rounded-button-borderRadius text-button-fontSize font-button-fontWeight bg-button-backgroundColor text-button-textColor ring-offset-background transition-colors focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:pointer-events-none disabled:opacity-50 ",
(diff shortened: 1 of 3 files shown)


mkilpatrick (approved):

src/components/puck/atoms/button.tsx
```diff
@@ -5,7 +5,7 @@ import { clsx } from "clsx";
 import { NumberOrDefault } from "../../editor/NumberOrDefaultField.tsx";
 
 const buttonVariants = cva(
-  "py-4 components inline-flex items-center justify-center whitespace-nowrap rounded-button-borderRadius text-button-fontSize font-button-fontWeight ring-offset-background transition-colors focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:pointer-events-none disabled:opacity-50 ",
+  "py-4 components inline-flex items-center justify-center whitespace-nowrap rounded-button-borderRadius text-button-fontSize font-button-fontWeight bg-button-backgroundColor text-button-textColor ring-offset-background transition-colors focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:pointer-events-none disabled:opacity-50 ",
```
benlife5, line 8: This will get combined with the variants below so every button will two bg classes, two text color classes, etc.

I would say to move these into the primary variant and update the border colors and focus/hover states too. Not sure how we want to handle the secondary/outline/link variants.

asanehisa: Oh yeah with cva, the base values get overridden by the variants (if selected).

I think I can make a default variant with those values though + make the default default?

benlife5: I might be misremembering but I think cva just concatenates the base with the selected variants and then ideally we'd use `cn` to disambiguate overlapping  but the custom TM classes don't work well with that by default. 

Maybe check with aaron about how these colors should be working? I think default could make sense. I remember being confused about how the buttons variants were supposed to react to the theme when we first added the theme

asanehisa: added default where the button-backgroundColor and button-textColor are used in place of primary / secondary

benlife5 (dismissed):

benlife5 (approved):

jwartofsky-yext (approved):

PR #155, opened 2024-11-13
fix: theme manager dev mode stuck loading

benlife5:
puckInitialHistoryFetched was never set to true in dev theme mode

### DIFF
--- a/src/internal/hooks/useMessageReceivers.ts
+++ b/src/internal/hooks/useMessageReceivers.ts
@@ -63,7 +63,9 @@ export const useCommonMessageReceivers = (
 
   useReceiveMessage("getThemeData", TARGET_ORIGINS, (send, payload) => {
     const payloadString = payload as unknown as string;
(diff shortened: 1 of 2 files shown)


src/internal/hooks/useMessageReceivers.ts
```diff
@@ -63,9 +63,7 @@ export const useCommonMessageReceivers = (
 
   useReceiveMessage("getThemeData", TARGET_ORIGINS, (send, payload) => {
     const payloadString = payload as unknown as string;
-    const themeData = payloadString
-      ? jsonFromEscapedJsonString(payloadString)
-      : {};
+    const themeData = payloadString ? JSON.parse(payloadString) : {};
```
mkilpatrick, line 66: Are you sure we don't need the escaped function? This comes from the DB right?

benlife5: Was able to use single quotes for the font names, which also fixed the issue, so we can just leave this as it is

PR #157, opened 2024-11-13
feat: add Promo component

jwartofsky-yext:
Adds a new component "Promo" which includes a heading, description, image, and cta

All props work as expected

image


benlife5:
LGTM, can you add it to the registry in `src/components/puck/registry/components.ts` too?

src/components/puck/Promo.tsx
```diff
@@ -0,0 +1,340 @@
...
+  resolveYextEntityField,
+  EntityField,
+  YextEntityField,
+  YextEntityFieldSelector,
+  NumberOrDefault,
+  HeadingProps,
+  NumberFieldWithDefaultOption,
+  CTAProps,
+  ImageWrapperProps,
+  Section,
```
benlife5, line 16: Move the component imports to the more specific imports like "./atoms/Body.js so that when it is pulled via the CLI it uses the local version. (Keep things like YextEntityField imported from index.js so that it gets imported from @yext/visual-editor when pulled down)

src/components/puck/registry/components.ts
```diff
@@ -108,4 +108,9 @@ export const ui: Registry = [
     type: "registry:ui",
     files: ["TextList.tsx"],
   },
+  {
+    name: "Promo",
+    type: "registry:ui",
+    files: ["Promo.tsx"],
```
benlife5, line 114: add registryDependencies with the other components used inside Promo

src/components/puck/Promo.tsx
```diff
@@ -0,0 +1,340 @@
...
+  Heading,
+  Body,
+  CTA,
+} from "../../index.js";
+import { imageWrapperVariants } from "./Image.tsx";
```
benlife5, line 21: Do .js imports in the components because the starter has a more basic tsconfig that doesn't allow ts imports

benlife5 (approved):

mkilpatrick (approved):

PR #160, opened 2024-11-15
fix: dev mode ui state stuck in theme manager

benlife5:
If the localstorage layout data in dev mode includes hidden sidebars, theme manager dev mode kept resetting to hidden sidebars after any styles were updated

### DIFF
--- a/src/internal/components/ThemeEditor.tsx
+++ b/src/internal/components/ThemeEditor.tsx
@@ -1,5 +1,11 @@
 import React, { useCallback, useEffect, useState } from "react";
-import { InitialHistory, Config, Data } from "@measured/puck";
+import {
+  InitialHistory,
+  Config,
+  Data,
+  History,
+  AppState,
+} from "@measured/puck";
 import { TemplateMetadata } from "../types/templateMetadata.ts";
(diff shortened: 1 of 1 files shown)


src/internal/components/ThemeEditor.tsx
```diff
@@ -71,13 +77,18 @@ export const ThemeEditor = (props: ThemeEditorProps) => {
...
+        localHistoryArray
+      ) as History<AppState>[];
       const localHistoryIndex = localHistories.length - 1;
       if (localHistoryIndex >= 0) {
         devLogger.log("Theme Dev Mode - Using layout data from local storage");
         setPuckInitialHistory({
           // @ts-expect-error https://github.com/measuredco/puck/issues/673
-          histories: localHistories,
+          histories: localHistories.map((h) => {
+            // strip ui state
```
mkilpatrick, line 89: Would we want to do this everywhere and not just devMode specifically?

benlife5: For Theme Manager, the object from content in prod mode doesn't have ui data on it
For Layout Editor, we could do this but I'm not sure we would want to... I think it makes sense that if you hide a sidebar and refresh the page it would still be hidden

mkilpatrick (approved):

PR #164, opened 2024-11-20
feat: use select for fontSize

briantstephan:
This makes fontSize fields use a select rather than a number input. The px font size shown in the selector correctly updates based on any changes the user makes in their tailwind config.


src/components/puck/atoms/heading.tsx
```diff
@@ -1,10 +1,25 @@
 import * as React from "react";
 import { cva, type VariantProps } from "class-variance-authority";
-import { type NumberOrDefault } from "../../editor/NumberOrDefaultField.js";
 
 // Define the variants for the heading component
 const headingVariants = cva("components", {
   variants: {
+    fontSize: {
+      default: "",
```
mkilpatrick, line 8: Are these the same everywhere? Should we share this?

src/utils/VisualEditorProvider.tsx
```diff
@@ -0,0 +1,42 @@
...
+  children: React.ReactNode;
+};
+
+type EditorProps = {
+  entityFields: YextSchemaField[];
+  tailwindConfig: TailwindConfig;
+};
+
+type VisualEditorProviderProps<T> = UniversalProps<T> &
+  AllOrNothing<EditorProps>;
```
mkilpatrick, line 23: So both entityFields and tailwindConfig are required or neither?

src/components/editor/FontSizeSelector.tsx
```diff
@@ -0,0 +1,100 @@
...
+  if (typeof customFontSize == "string") {
+    return convertToPixels(customFontSize);
+  }
+  if (Array.isArray(customFontSize) && customFontSize.length > 0) {
+    return convertToPixels(customFontSize[0]);
+  }
+  return 0;
+};
+
+export const FontSizeSelector = (): Field => {
```
benlife5, line 68: Can you add an optional prop for label? Theoretically could have a component where you'd be setting the font size for more than one thing

src/hooks/index.ts
```diff
@@ -1,11 +1,10 @@
...
 } from "./useEntityFields.tsx";
+export { useDocument, DocumentContext } from "./useDocument.tsx";
 export {
-  DocumentProvider,
-  useDocument,
-  DocumentContext,
-} from "./useDocument.tsx";
+  useTailwindConfig,
+  TailwindConfigContext,
+} from "./useTailwindConfig.tsx";
```
benlife5, line 10: Update the hooks README for the new provider setup

src/components/puck/atoms/button.tsx
```diff
@@ -19,6 +18,22 @@ const buttonVariants = cva(
         large: "h-11 px-8",
         icon: "h-10 w-10",
       },
+      fontSize: {
```
benlife5, line 21: I don't think these need to have the ! (but double check that). We added the ! to heading because the fontSize variants can conflict with the headingLevel text size variants

src/components/puck/atoms/heading.tsx
benlife5: It's slightly different across the atoms. We could potentially try to standardize, but I think you could import Heading or Body from VE if you want to use it out-of-the-box and if you're looking for further customization you probably want direct access. Not sure though

src/utils/VisualEditorProvider.tsx
briantstephan: Yes, this allows the VisualEditorProvider to be used for both the edit and location templates, since only the DocumentProvider was being used in the location template previously.

src/components/puck/atoms/button.tsx
briantstephan: They do need it, the font size on the button doesn't seem to change without it.

benlife5: ah there's  `text-button-fontSize` in the base string at the top. Try removing that (and update the default below to use text-button-fontSize

src/components/puck/atoms/heading.tsx
briantstephan: I agree, I think it's best for the time being to keep them separate.

src/components/puck/atoms/button.tsx
briantstephan: Ah yeah that worked.

benlife5 (approved):

mkilpatrick (approved):

PR #167, opened 2024-11-21
fix: font selector does not work with recent puck canaries

benlife5:
Upcoming puck fixes make the font selector re-mount more, so we need to use useCallback (like we do for the entire ThemeSidebar)

This works with our current puck version, but isn't needed until we upgrade again


mkilpatrick (approved):

colton-demetriou (approved):

benlife5:
Planning to wait on merging until we actually go to upgrade puck versions so that we can test with the latest puck version

benlife5:
Confirmed working with Puck 0.17.1

PR #170, opened 2024-11-22
refactor: custom twMerge and tailwind cleanup

benlife5:
1. Adds themeMangerCn, which can merge tailwind classes while also respecting the custom classes we create for theme manger
    - The big benefit of this is consistent disambiguation of class names, especially allowing you to pass a className string to our components that will override the built-in variants 
     - Doesn't really work with modified theme configs (would need to follow our arbitrary naming patterns) -- could potentially build a twMerge config that reads theme.config, but that would be a lot more complicated (and I would think that if you're making a lot of changes to the theme, you're probably building custom components with custom styling logic)
3. Adds a separate cn that respects the `ve-` prefix for the components we that use internally in the editor (more similar to the way it was before we added the component library)
    - also removed a duplicated internal button component


jwartofsky-yext (approved):
lgtm

mkilpatrick (approved):

asanehisa (approved):

PR #173, opened 2024-12-16
refactor: rename messengers

benlife5:
While we are doing breaking changes, we might as well update the message names that were created pre-themes, which can be confusing.

https://gerrit.yext.com/c/alpha/+/278839

### DIFF
--- a/src/internal/components/LayoutEditor.tsx
+++ b/src/internal/components/LayoutEditor.tsx
@@ -18,23 +18,18 @@ const devLogger = new DevLogger();
 type LayoutEditorProps = {
   puckConfig: Config;
   templateMetadata: TemplateMetadata;
-  visualConfigurationData: Data;
+  layoutData: Data;
   themeData: ThemeData;
   themeConfig: ThemeConfig | undefined;
 };
 
 export const LayoutEditor = (props: LayoutEditorProps) => {
-  const {
-    puckConfig,
-    templateMetadata,
-    visualConfigurationData,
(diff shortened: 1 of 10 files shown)


src/internal/components/LayoutEditor.tsx
```diff
@@ -18,23 +18,18 @@ const devLogger = new DevLogger();
...
 
 export const LayoutEditor = (props: LayoutEditorProps) => {
-  const {
-    puckConfig,
-    templateMetadata,
-    visualConfigurationData,
-    themeData,
-    themeConfig,
-  } = props;
+  const { puckConfig, templateMetadata, layoutData, themeData, themeConfig } =
```
briantstephan, line 27: Nit, this formatting is a little weird. Should change it to be like how it was before (but with updated names)

benlife5:
Rebased on main

src/internal/components/LayoutEditor.tsx
benlife5: Yeah I agree I like the old one better but this is what prettier did 🤷‍♂️

jwartofsky-yext (approved):
lgtm

src/internal/components/LayoutEditor.tsx
briantstephan: Oh nvm then

briantstephan (approved):

asanehisa (approved):

PR #183, opened 2025-01-07
refactor: move pages-components to external peer to enable analytics

benlife5:
By making pages-components external, the VE components will use the same analytics context as the starter.

Confirmed that with this changes (and adding the provider to the starter), page views and link clicks appear in Snowflake.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,6 @@
     "@radix-ui/react-slot": "^1.1.0",
     "@radix-ui/react-switch": "^1.1.0",
     "@radix-ui/react-tooltip": "^1.1.2",
-    "@yext/pages-components": "^1.0.2",
     "class-variance-authority": "^0.7.0",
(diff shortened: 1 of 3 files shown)


asanehisa (approved):

package.json
```diff
@@ -107,6 +106,7 @@
   },
   "peerDependencies": {
     "react": "^17.0.2 || ^18.2.0",
-    "react-dom": "^17.0.2 || ^18.2.0"
+    "react-dom": "^17.0.2 || ^18.2.0",
+    "@yext/pages-components": "^1.0.0"
```
asanehisa, line 110: do we want this to be 1.0.0 or 1.0.4?

mkilpatrick: 1.0.0 is fine since it has the caret

mkilpatrick (approved):

PR #186, opened 2025-01-13
fix: entityField Wrapper

asanehisa:
On empty entityField, keep tooltip

On constant, don't show any tooltip. 

Within Card component, when EntityField is toggled, no changes to sizing

### DIFF
--- a/src/components/puck/Address.tsx
+++ b/src/components/puck/Address.tsx
@@ -83,9 +83,8 @@ const Address = ({
         <div>
           <EntityField
             displayName="Address"
-            fieldId={
-              addressField.constantValueEnabled ? "constant value" : "address"
-            }
+            fieldId="address"
+            constantValueEnabled={addressField.constantValueEnabled}
           >
             <RenderAddress
               address={address as AddressType}
--- a/src/components/puck/Address.tsx
+++ b/src/components/puck/Address.tsx
@@ -83,7 +83,7 @@ const Address = ({
         <div>
           <EntityField
             displayName="Address"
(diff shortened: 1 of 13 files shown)


src/components/puck/Emails.tsx
```diff
@@ -137,6 +137,10 @@ const Emails: React.FC<EmailsProps> = ({
     resolvedEmailList = [resolvedEmailList];
   }
 
+  if (resolvedEmailList?.length === 0) {
+    return <></>;
+  }
+
   return (
     <EntityField displayName="Email List" fieldId={emailListField.field}>
```
benlife5, line 145: I think the other issue is that even if constant value is enabled the tooltip displays the entity field name/id. (I think this is an issue in other components too)

Also -- I know on some components I made it say `Component Name (constant value)` if constant value was enabled and I think I remember Aaron saying to remove that haha so maybe verify how the entity field tooltips should be formatted with him

asanehisa: Oh! ok!

benlife5 (approved):
🔥

src/components/puck/Address.tsx
```diff
@@ -83,9 +83,8 @@ const Address = ({
         <div>
           <EntityField
             displayName="Address"
-            fieldId={
-              addressField.constantValueEnabled ? "constant value" : "address"
-            }
+            fieldId="address"
```
mkilpatrick, line 86: The other components use {fieldname.field}

mkilpatrick (approved):

PR #189, opened 2025-01-14
fix: card component fixes

jwartofsky-yext:
Fixes several issues with the Card component

Adds a level selector and background color selector to the card component

Verified that changing the level changes the size of text according to what is in the theme for Heading 1, Heading 2, etc

Color selector works as expected

Fixes an issue where changing "constant value" mode of one field would change the setting on another

### DIFF
--- a/src/components/editor/YextEntityFieldSelector.tsx
+++ b/src/components/editor/YextEntityFieldSelector.tsx
@@ -102,7 +102,6 @@ export const YextEntityFieldSelector = <T extends Record<string, any>, U>(
       return (
         <>
           <ConstantValueModeToggler
-            label={props.label}
             constantValueEnabled={value?.constantValueEnabled}
             toggleConstantValueEnabled={toggleConstantValueEnabled}
           />
@@ -127,16 +126,15 @@ export const YextEntityFieldSelector = <T extends Record<string, any>, U>(
 };
 
 const ConstantValueModeToggler = ({
-  label,
   constantValueEnabled,
   toggleConstantValueEnabled,
 }: {
-  label: string;
   constantValueEnabled: boolean;
   toggleConstantValueEnabled: (constantValueEnabled: boolean) => void;
 }) => {
-  const entityButtonId = `ve-use-entity-value-${label.toLowerCase()}`;
-  const constantButtonId = `ve-use-constant-value-${label.toLowerCase()}`;
+  const random = Math.floor(Math.random() * 999999);
+  const entityButtonId = `ve-use-entity-value-${random}`;
+  const constantButtonId = `ve-use-constant-value-${random}`;
   return (
     <div className="ve-mb-2 ve-w-full">
       <RadioGroup defaultValue={constantValueEnabled?.toString() ?? "false"}>
--- a/src/components/editor/YextEntityFieldSelector.tsx
+++ b/src/components/editor/YextEntityFieldSelector.tsx
@@ -102,6 +102,7 @@ export const YextEntityFieldSelector = <T extends Record<string, any>, U>(
       return (
         <>
           <ConstantValueModeToggler
+            label={props.label}
             constantValueEnabled={value?.constantValueEnabled}
             toggleConstantValueEnabled={toggleConstantValueEnabled}
           />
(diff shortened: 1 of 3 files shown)


mkilpatrick (approved):

benlife5 (approved):

asanehisa (dismissed):

jwartofsky-yext:
Now includes fixes for AC1-5: https://yexttest.atlassian.net/browse/SUMO-6838

AC6 seems to be something with Puck. Will be handled separately

src/components/puck/Card.tsx
```diff
@@ -280,31 +312,39 @@ const CardWrapper = ({
...
         <div className="flex flex-col justify-center gap-y-4 md:gap-y-8 p-4 md:px-16 md:py-0 w-full break-all">
           {heading?.text && (
-            <Heading
-              fontSize={heading.fontSize}
-              color={heading.color}
-              transform={heading.transform}
-            >
-              {resolveYextEntityField(document, heading.text)}
-            </Heading>
+            <EntityField displayName="Heading" fieldId={heading.text.field}>
```
benlife5, line 315: Whoever merges second between you and @asanehisa's #186 should updates these as well

src/components/puck/Card.tsx
```diff
@@ -380,6 +422,7 @@ export const CardComponent: ComponentConfig<CardProps> = {
       fontSize: "default",
       variant: "primary",
     },
+    backgroundColor: "bg-palette-background",
```
benlife5, line 425: Ambiguous wording in the ticket... can you verify if this is supposed to default to the Palette color "Background" or use have an option "Default" that is `bg-grid-backgroundColor` under the hood

Ie 
Screenshot 2025-01-14 at 3 44 47 PM
vs. 

Screenshot 2025-01-14 at 3 47 34 PM

in theme editor

jwartofsky-yext:
> Now includes fixes for AC1-5: https://yexttest.atlassian.net/browse/SUMO-6838
> 
> AC6 seems to be something with Puck. Will be handled separately

AC6 has been fixed

src/components/puck/Card.tsx
jwartofsky-yext: It appears like this:
image

Which is the same as how the background selector for Banner appears

The background selector for footer is a dropdown like in your image however.

I'll get clarification

jwartofsky-yext: https://yext.slack.com/archives/C06A06BCUUF/p1736888711051729

jwartofsky-yext: Resolved by swapping to the design used in Footer

Banner should be swapped over to the selector aswell

src/components/puck/Card.tsx
jwartofsky-yext: Ack

src/components/editor/YextEntityFieldSelector.tsx
```diff
@@ -132,24 +132,27 @@ const ConstantValueModeToggler = ({
   constantValueEnabled: boolean;
   toggleConstantValueEnabled: (constantValueEnabled: boolean) => void;
 }) => {
+  const random = Math.floor(Math.random() * 999999);
```
mkilpatrick, line 135: Why do we need this?

jwartofsky-yext: It resolves AC6 from here: https://yexttest.atlassian.net/browse/SUMO-6838

The issue was that when we click one of the buttons, it would update the button for another field in the same component.

Using unique ids resolves the problem.

I tried at first with passing props.label through, but found that it is not always unique (often comes through as just "Entity" for multiple fields).

ex: https://github.com/yext/visual-editor/blob/main/src/components/puck/Card.tsx#L69

Since we won't likely have more than 3 or so entity fields per component, using a random number with 5 digits should prevent the overlap.

mkilpatrick: Can you use something off of the field from render? It's 6 digits btw.

mkilpatrick (approved):

PR #192, opened 2025-01-17
feat: adjust Grid and Flex

asanehisa:
Flex + Grid both take up entire page. The content inside of Grid has a max-width that the user can select.

Both Flex and Grid have vertical and horizontal padding props as well as a gap prop. 

Dragging between Dropzones is now enabled

https://github.com/user-attachments/assets/8ead5be1-964d-4612-a3cb-be0a4cfc673d

Screenshot 2025-01-23 at 12 05 49 PM

Gap, verticalPadding, and horizontalPadding all use the same dropdown. Essentially the tailwind options 0 to 24.


package.json
```diff
@@ -57,6 +57,7 @@
     "react-color": "^2.19.3",
     "sonner": "^1.5.0",
     "tailwind-merge": "^2.4.0",
+    "ts-node": "^10.9.2",
```
mkilpatrick, line 60: You shouldn't need to add ts-node

package.json
```diff
@@ -37,7 +37,7 @@
     "generate-registry": "tsx src/components/puck/registry/build-registry.ts"
   },
   "dependencies": {
-    "@measured/puck": "0.17.4",
+    "@measured/puck": "0.18.0",
```
mkilpatrick, line 40: We're not ready for 18, unless you want to hold this off until we upgrade.

asanehisa: Yup will have to wait til we ready for 18 as Aaron wanted the new Drag and Drop features to work for Flex + Grid.

package.json
asanehisa: Screenshot 2025-01-22 at 1 11 11 PM
Hmmm can't seem to pnpm i and pack without it.

mkilpatrick: Hmm that's odd. I don't get that error even with trying to upgrade Puck. Maybe it works because I have it installed globally? But wouldn't others get this error too? @benlife5 have you seen this?

benlife5: yeah that's weird since the error is coming from dependency's dependency
When I check out this branch, I can pnpm i, pnpm rm ts-node, and pnpm pack fine. Maybe the lock file was messed up because you're in a git merge state?

mkilpatrick (approved):

mkilpatrick (approved):

PR #198, opened 2025-01-21
fix: limit font-size for buttons

jwartofsky-yext:
For buttons, which are used in cta and
Get Directions, selecting a font-size larger than 5xl would make the text bleed out of the button.

This change removes the font size options larger than 4xl per Aaron's request.

### DIFF
--- a/src/components/puck/Promo.tsx
+++ b/src/components/puck/Promo.tsx
@@ -11,7 +11,6 @@ import {
   FontSizeSelector,
 } from "../../index.js";
 import { Body } from "./atoms/body.js";
-import { ButtonProps } from "./atoms/button.js";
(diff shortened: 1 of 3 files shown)


src/components/puck/Promo.tsx
```diff
@@ -11,6 +11,7 @@ import {
   FontSizeSelector,
 } from "../../index.js";
 import { Body } from "./atoms/body.js";
+import { ButtonProps } from "./atoms/button.js";
```
benlife5, line 14: Can this be passed through CTAProps (like CTAProps["fontSize"])?

jwartofsky-yext: Done

benlife5 (approved):

mkilpatrick (approved):

PR #208, opened 2025-01-22
feat: add link type to CTA

jwartofsky-yext:
Adds Link Type as a field for CTAs
    
    The Link Type is passed from Content if an entity field is
    used
    
    The Link Type may be set when using a constant value
    
    When Link Type is type Phone, the link url will be changed to
    'tel:{linkValue}'


packages/visual-editor/src/components/puck/atoms/cta.tsx
```diff
@@ -30,7 +32,9 @@ const CTA = ({
       borderRadius={borderRadius}
       fontSize={fontSize}
     >
-      <Link href={link ?? ""}>{label}</Link>
+      <Link href={linkType == "PHONE" ? `tel:${link}` : (link ?? "")}>
```
benlife5, line 35: I think you can pass `cta={link , linkType, label}` to Link and it will handle this with some validation

jwartofsky-yext: Done

packages/visual-editor/src/components/puck/atoms/cta.tsx
```diff
@@ -30,7 +32,14 @@ const CTA = ({
       borderRadius={borderRadius}
       fontSize={fontSize}
     >
-      <Link href={link ?? ""}>{label}</Link>
+      <Link
+        cta={{
```
mkilpatrick, line 36: Noting that this actually changes the underlying analytics action from LINK to CALL_TO_ACTION and eventName from "link" to "cta", but I think that's appropriate given the name of the atom.

mkilpatrick (dismissed):

packages/visual-editor/src/components/puck/atoms/cta.tsx
jwartofsky-yext: Good to know

mkilpatrick (approved):

benlife5 (approved):

PR #210, opened 2025-01-23
feat: add theme mode to localDev mode

jwartofsky-yext:
Adds support for Theme Manager to localDev mode

Simply click the text that says "Layout Mode" at the center of the header to swap modes.

Note that sometimes you have to click on something in puck after swapping modes for the theme to update.

image


starter/src/templates/dev.tsx
```diff
@@ -95,23 +97,36 @@ export const getPath: GetPath<TemplateProps> = ({ document }) => {
...
-      <Editor
-        document={document}
-        componentRegistry={componentRegistry}
-        themeConfig={themeConfig}
-        localDev={true}
-      />
-    </VisualEditorProvider>
+      <div>
+        <div className={"flex-container"}>
+          <ChakraButton className={"toggle-button"} onClick={() => {
```
mkilpatrick, line 108: Can we just use the same button component that clear local changes uses? We don't need to import a whole new library just for a button. It doesn't even look like a button in your screenshot, just text.

mkilpatrick: With real local dev idk if it will be more complicated to get dev mode working where we moved the theme mgr button, but I'd rather not move this into visual-editor for now. Just use a normal html button and slap some styling on it so it doesn't look horrendous.

jwartofsky-yext:
git hook started failing because it couldn't find eslint config

this required me to move it to the root

starter/src/index.css
```diff
@@ -7,8 +7,13 @@
   justify-content: center;
   align-items: center;
   height: 80px;
+  margin-bottom: -80px;
```
mkilpatrick, line 10: Almost never use negative margins. It means something else is pushing the element and negative means you're trying to overwrite that.

starter/src/templates/dev.tsx
```diff
@@ -16,11 +16,10 @@ import {
   YextSchemaField,
 } from "@yext/visual-editor";
 import { themeConfig } from "../../theme.config";
-import { buildSchema } from "../utils/buildSchema";
+import { buildSchema } from "../utils/buildSchema.ts";
```
mkilpatrick, line 19: This came back?

eslint.config.mjs
```diff
@@ -21,7 +21,7 @@ export default [
   ...compat.extends(
     "eslint:recommended",
     "plugin:react/recommended",
-    "plugin:@typescript-eslint/recommended"
+    "plugin:@typescript-eslint/recommended",
```
mkilpatrick, line 24: Wouldn't this have failed before?

jwartofsky-yext: I think because it's the last element the comma is optional

eslint added it though, so I'll keep it

starter/src/templates/dev.tsx
jwartofsky-yext: yup

I remembered the original reason I had it

Not having it makes eslint fail

starter/src/index.css
jwartofsky-yext: The negative margin here allows the button to appear inside of the header. We could also have the button appear above the header, but that looks ugly.

A cleaner solution would be to add it to the visual-editor properly, but we wanted to avoid that for now.

eslint.config.mjs
jwartofsky-yext: I think the .github husky hook wasn't triggering until I used chmod to change the permissions on the file.

mkilpatrick (approved):

PR #213, opened 2025-01-27
feat: adjust labels and values for props 

asanehisa:
Font Size is adjusted to look like 
image

Font Weight and Border Radius are likewise adjusted to look similar with the options being the tailwind defaults. 

Maximum Width prop is updated to show px in label.


packages/visual-editor/src/components/puck/atoms/button.tsx
```diff
@@ -34,10 +34,15 @@ const buttonVariants = cva(
       },
```
benlife5, line 34: Presuming we don't use size: default anywhere else for <Button>, I think you can remove it from the variants

packages/visual-editor/src/utils/themeConfigOptions.ts
```diff
@@ -0,0 +1,31 @@
+export const getFontSizeOptions = () => {
```
benlife5, line 1: It's pretty self-explanatory but maybe add a small entry about these in packages/visual-editor/src/utils/README.md

packages/visual-editor/src/utils/visualEditorFonts.ts
```diff
@@ -74,15 +74,15 @@ const constructGoogleFontLinkTags = (fonts: FontRegistry) => {
...
-  { label: "Thin", value: "100" },
-  { label: "Extralight", value: "200" },
-  { label: "Light", value: "300" },
-  { label: "Normal", value: "400" },
-  { label: "Medium", value: "500" },
-  { label: "Semibold", value: "600" },
-  { label: "Bold", value: "700" },
-  { label: "Extrabold", value: "800" },
-  { label: "Black", value: "900" },
+  { label: "Thin (100)", value: "100" },
```
benlife5, line 77: Do you think we should move this to themeConfigOptions.ts?

packages/visual-editor/src/components/puck/atoms/button.tsx
asanehisa: Agree, will do

packages/visual-editor/src/utils/themeConfigOptions.ts
asanehisa: Ok!

packages/visual-editor/src/utils/visualEditorFonts.ts
asanehisa: Sure I think that could be helpful for having them all in one spot

asanehisa: Actually maybe I won't since defaultWeightOptions isn't used in the same way as the others. Instead it is used locally in visualEditorFont.tsx and compared with the specific font selected to return the actual fontWeightOptions.

benlife5 (dismissed):

packages/visual-editor/src/utils/themeResolver.test.ts
```diff
@@ -50,15 +50,15 @@ const getColorOptions = () => {
...
-    { label: "Thin", value: "100" },
-    { label: "Extralight", value: "200" },
-    { label: "Light", value: "300" },
-    { label: "Normal", value: "400" },
-    { label: "Medium", value: "500" },
-    { label: "Semibold", value: "600" },
-    { label: "Bold", value: "700" },
-    { label: "Extrabold", value: "800" },
-    { label: "Black", value: "900" },
+    { label: "Thin (100)", value: "100" },
```
mkilpatrick, line 53: Didn't we get the values from TW for other variables, opposed to hardcoding?

mkilpatrick (approved):

packages/visual-editor/src/utils/themeResolver.test.ts
benlife5: src/components/editor/FontSizeSelector.tsx has some logic for that

I think getWeightOptions is fine hardcoded because I don't think you'd ever remap the names to different numbers for font weight

The new getFontSizeOptions just a set of defaults for use in the theme.config (right Alexis?) so that's also fine because a developer could define their own options if they were changing the tailwind values.

getBorderRadiusOptions is used in the components, so maybe that should be more like FontSizeSelector? If we want to support updating the px numbers in the dropdown if a developer changed the tailwind config.

@asanehisa should getBorderRadiusOptions be used in CTAWrapper too?

asanehisa: >  getFontSizeOptions is just a set of defaults for use in the theme.config (right Alexis?)
Yup! 

Hmmm I can update BorderRadiusOptions to be closer to FontSizeSelector so the px values are updated if the developer changes their tailwind config.

and add getBorderRadiusOptions to CTAWrapper !

benlife5 (approved):

jwartofsky-yext (approved):

mkilpatrick (approved):

PR #217, opened 2025-01-29
fix: update shadcn component registry

benlife5:
shadcn updated the shape of their cli registry so ours was broken


mkilpatrick (approved):

asanehisa (approved):

benlife5:
https://github.com/YextSolutions/pages-visual-editor-starter/pull/205
Pinning the shadcn version in the starter so that this doesn't happen again unless we know we want to update

PR #219, opened 2025-01-30
feat: add analytics ui to theme editor

benlife5:
Adds a toggle to the left sidebar in Theme Editor to switch between theme mode and analytics mode

When in analytics mode, the right sidebar in Theme Editor has new puck fields, right now it is just Google Tag Manager ID

This fields is saved to an object at `document.__.theme.siteAttributes`

https://github.com/user-attachments/assets/76ade0ad-40ce-4743-87c9-3111dc309180


packages/visual-editor/src/internal/puck/components/theme-editor-sidebars/AnalyticsFieldsSidebar.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+type AnalyticsFieldsSidebarProps = {
+  themeData: ThemeData;
+  onThemeChange: OnThemeChangeFunc;
+};
+
+export const AnalyticsFieldsSidebar = (props: AnalyticsFieldsSidebarProps) => {
+  const { themeData, onThemeChange } = props;
+
+  const handleThemeChange = (key: string, value: any) => {
```
mkilpatrick, line 20: Weird name for this in the context of analytics even though I know it's setting the themeData.

packages/visual-editor/src/internal/puck/components/theme-editor-sidebars/ThemeEditorLeftSidebar.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const themeEditorModes = {
+  theme: {
+    title: "Theme",
+    description:
+      "Edit your site's theme, including your color palette, fonts, and text size.",
+  },
+  analytics: {
+    title: "Analytics",
+    description: "Edit your site's Google Analytics settings.",
```
mkilpatrick, line 11: Verify with Aaron the title/desc for these.

packages/visual-editor/src/internal/puck/components/theme-editor-sidebars/ThemeEditorRightSidebar.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  const themeData =
+    themeHistoriesRef.current?.histories[themeHistoriesRef.current?.index].data;
+
+  return (
+    <div>
+      <Alert>
+        <AlertDescription>
+          Changes will affect all page sets and layouts for the entire site.
```
mkilpatrick, line 45: I wonder if we still want this at all or only for theme

packages/visual-editor/src/internal/puck/components/theme-editor-sidebars/AnalyticsFieldsSidebar.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const newValues = {
+      ...themeData,
+      siteAttributes: {
```
mkilpatrick, line 23: Is it going to be a problem that this key is at the same level as the rest of the theme stuff? Does it need to be excluded in the applyTheme function now?

packages/visual-editor/src/internal/puck/components/theme-editor-sidebars/AnalyticsFieldsSidebar.tsx
benlife5: Yeah "Theme" as both the overall mode and as a sub-mode can get confusing. I'm hoping we figure out a better naming scheme and then can come back through an update these variable names

packages/visual-editor/src/internal/puck/components/theme-editor-sidebars/AnalyticsFieldsSidebar.tsx
benlife5: Fortunately applyTheme already only copies over values that match a key generated by the themeConfig. So any other keys will be available on document.__.theme but they won't get picked up by the theme/styling system

jwartofsky-yext (approved):
lgtm

packages/visual-editor/src/internal/puck/components/theme-editor-sidebars/ThemeEditorLeftSidebar.tsx
benlife5: Aaron said to go ahead with this copy for now and we can change it later

mkilpatrick (approved):

PR #220, opened 2025-01-31
feat: add applyAnalytics

benlife5:
Returns a Google Tag Manager script that uses the Google Tag Manager ID set in the Theme Editor

### DIFF
--- a/packages/visual-editor/src/utils/README.md
+++ b/packages/visual-editor/src/utils/README.md
@@ -448,3 +448,23 @@ Returns a list of border radius options to be optionally used in the theme.confi
 | rounded-2xl    | 2XL (16px)    | 16px   |
 | rounded-3xl    | 3XL (24px)    | 24px   |
 | rounded-full   | Full (9999px) | 9999px |
+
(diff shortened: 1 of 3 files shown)


benlife5:
@mkilpatrick The doc says "Tags go as high up in the `<head>` tag in the Site template as possible, ideally immediately after the <head> tag." I think this would require a pagesjs change? When this is used in the `other` section of the HeadConfig, it injects the script about in the middle of the `<head>` tag
Screenshot 2025-01-31 at 10 28 52 AM

benlife5:
> @mkilpatrick The doc says "Tags go as high up in the `<head>` tag in the Site template as possible, ideally immediately after the tag." I think this would require a pagesjs change? When this is used in the `other` section of the HeadConfig, it injects the script about in the middle of the `<head>` tag 

Discussed offline and we're going keep this unless it actually is a problem later. Seems unlikely that the google script would load meaningfully slower being placed after the few non-script tags we have above it

mkilpatrick (approved):

PR #221, opened 2025-02-05
feat(deps): add lz-compression for local storage

jwartofsky-yext:
Verified that dev mode local storage loads as expected.

Refreshing page and toggling between theme and layout mode works normally.

### DIFF
--- a/packages/visual-editor/src/internal/components/LayoutEditor.tsx
+++ b/packages/visual-editor/src/internal/components/LayoutEditor.tsx
@@ -12,6 +12,7 @@ import { ThemeConfig } from "../../utils/themeResolver.ts";
 import { updateThemeInEditor } from "../../utils/applyTheme.ts";
 import { useThemeLocalStorage } from "../hooks/theme/useLocalStorage.ts";
 import { useCommonMessageSenders } from "../hooks/useMessageSenders.ts";
+import * as lzstring from "lz-string";
 
 const devLogger = new DevLogger();
 
@@ -72,8 +73,8 @@ export const LayoutEditor = (props: LayoutEditorProps) => {
     if (!themeConfig) {
       return;
     }
-    const localHistoryArray = window.localStorage.getItem(
-      buildThemeLocalStorageKey()
+    const localHistoryArray = lzstring.decompress(
+      window.localStorage.getItem(buildThemeLocalStorageKey()) || ""
     );
 
     // use theme from localStorage when in dev mode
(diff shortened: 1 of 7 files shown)


packages/visual-editor/src/internal/components/LayoutEditor.tsx
```diff
@@ -72,8 +73,8 @@ export const LayoutEditor = (props: LayoutEditorProps) => {
     if (!themeConfig) {
       return;
     }
-    const localHistoryArray = window.localStorage.getItem(
-      buildThemeLocalStorageKey()
+    const localHistoryArray = lzstring.decompress(
```
benlife5, line 76: What happens if the localstorage item isn't a compressed string?

jwartofsky-yext: If it's an empty string it works fine

Not sure about a non-compressed string. However with this change we only save compressed strings to the local storage.

In the recent patchset I also made it so we only save to localStorage in dev mode

Going to test in us3dev to verify

packages/visual-editor/src/internal/components/LayoutEditor.tsx
```diff
@@ -168,8 +169,8 @@ export const LayoutEditor = (props: LayoutEditorProps) => {
     }
 
     // Check localStorage for existing Puck history
-    const localHistoryArray = window.localStorage.getItem(
-      buildVisualConfigLocalStorageKey()
+    const localHistoryArray = lzstring.decompress(
```
benlife5, line 172: If this PR is also going to remove localstorage in prod mode, then go ahead and remove this read too

packages/visual-editor/src/internal/components/LayoutEditor.tsx
benlife5: Sounds good -- shouldn't have to worry too much about it if it's only dev mode. Maybe try to test that you can still click the Clear Local Changes button if this read fails because ideally you wouldn't ever have to manually delete the localstorage entry

benlife5: SGTM. Did you want to go ahead and remove the prod mode localstorage reads too?

packages/visual-editor/src/internal/components/LayoutEditor.tsx
jwartofsky-yext: Done

benlife5 (dismissed):

benlife5 (approved):

PR #222, opened 2025-02-05
fix: update button font size options

asanehisa:
Screenshot 2025-02-05 at 4 00 20 PM

fixed it so that 5xl, 6xl, 7xl... doesn't show up anymore. There's a separate CR to update the starter's theme.config

### DIFF
--- a/packages/visual-editor/src/components/editor/FontSizeSelector.tsx
+++ b/packages/visual-editor/src/components/editor/FontSizeSelector.tsx
@@ -4,22 +4,29 @@ import { useTailwindConfig } from "../../hooks/useTailwindConfig.tsx";
 import { TailwindConfig } from "../../utils/themeResolver.ts";
 import { ChevronDown } from "lucide-react";
 
-const fontSizeOptions = [
-  { label: "Default", value: "default", px: "" },
-  { label: "XS", value: "xs", px: "12" },
-  { label: "SM", value: "sm", px: "14" },
-  { label: "Base", value: "base", px: "16" },
-  { label: "LG", value: "lg", px: "18" },
-  { label: "XL", value: "xl", px: "20" },
-  { label: "2XL", value: "2xl", px: "24" },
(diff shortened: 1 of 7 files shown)


packages/visual-editor/src/components/editor/FontSizeSelector.tsx
```diff
@@ -4,22 +4,29 @@ import { useTailwindConfig } from "../../hooks/useTailwindConfig.tsx";
...
-  { label: "2XL", value: "2xl", px: "24" },
-  { label: "3XL", value: "3xl", px: "30" },
-  { label: "4XL", value: "4xl", px: "36" },
-  { label: "5XL", value: "5xl", px: "48" },
-  { label: "6XL", value: "6xl", px: "60" },
-  { label: "7XL", value: "7xl", px: "72" },
-  { label: "8XL", value: "8xl", px: "96" },
-  { label: "9XL", value: "9xl", px: "128" },
-];
+const fontSizeOptions = (isButton = false) => {
```
benlife5, line 7: Nit but maybe make this param includeLargeSizes or something like that? So that it's more extensible to other components that don't work with the super large sizes or to a different button type that does accept them

starter/theme.config.ts
```diff
@@ -366,7 +366,16 @@ export const themeConfig: ThemeConfig = {
         label: "Font Size",
         type: "select",
         plugin: "fontSize",
-        options: getFontSizeOptions(),
+        options: [
```
benlife5, line 369: Maybe update getFontSizeOptions() to have the same param? Also this is probably harder than it seems but it would be nice if getFontSizeOptions and fontSizeOptions depended on the same hardcoded values to reduce the chance of slight inconsistencies

asanehisa: oooh agree

benlife5 (approved):
Nice!

PR #223, opened 2025-02-10
feat: update Emails to use Link component

jwartofsky-yext:
All other components are already using <Link>

Verified that emails component works as expected.

### DIFF
--- a/packages/visual-editor/src/components/puck/Emails.tsx
+++ b/packages/visual-editor/src/components/puck/Emails.tsx
@@ -11,7 +11,9 @@ import {
   YextEntityFieldSelector,
   getFontWeightOverrideOptions,
   FontSizeSelector,
+  CTA,
 } from "../../index.js";
+import { LinkTypes } from "@yext/pages-components";
 
 const emailsVariants = cva("components list-inside font-body-fontFamily", {
   variants: {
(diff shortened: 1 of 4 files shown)


packages/visual-editor/src/components/puck/Emails.tsx
```diff
@@ -155,7 +156,13 @@ const Emails: React.FC<EmailsProps> = ({
           .map((text: any, index: any) => (
             <li key={index} className={`mb-2 flex items-center`}>
               <img className={"mr-2"} src={mailIcon} />
-              <span>{includeHyperlink ? <a href={text}>{text}</a> : text}</span>
+              <span>
+                {includeHyperlink ? (
+                  <Link href={text ?? "#"}>{text}</Link>
```
mkilpatrick, line 161: All of our Link components are going to need to be updated to use my [PR](https://github.com/yext/js/pull/80/files#diff-b4cf3f017db1502174fa9c7644c143c261a44bd24d98e075590c9d316373422cR10). I added some linkTypes. We should also update anything that has a specific use case to use the cta prop which allows you to set the linkType explicitly, rather than it being inferred. We should only use a generic href like this for a generic url case.

jwartofsky-yext: Done

packages/visual-editor/src/components/puck/Emails.tsx
```diff
@@ -144,23 +145,34 @@ const Emails: React.FC<EmailsProps> = ({
...
             <li key={index} className={`mb-2 flex items-center`}>
               <img className={"mr-2"} src={mailIcon} />
-              <span>{includeHyperlink ? <a href={text}>{text}</a> : text}</span>
+              {includeHyperlink ? (
+                <CTA
+                  label={text}
+                  link={text}
+                  linkType={"Email"}
+                  variant={"link"}
+                  className={classNameCn}
```
benlife5, line 170: Does passing className here override the styling of CTA with variant link?

benlife5 (approved):

mkilpatrick (approved):

PR #225, opened 2025-02-10
feat: add page level fields for Title and Description

jwartofsky-yext:
Adds page level fields for title and description in the puck editor.

They can use entity values or constant values.

They are set to use default values pointing to the name and description fields, however these do not seem to work.

I've posted a question on the Puck discord here to determine if this behavior is expected. : https://discord.com/channels/1153376562259951687/1338624057968689243/1338624057968689243

I was not able to get the defaultProps for the root to work with basic text fields either. The label field also does not seem to work.

### DIFF
--- a/packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
+++ b/packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
@@ -13,6 +13,7 @@ import { EntityFieldProvider } from "../../components/editor/EntityField.tsx";
 import { LayoutSaveState } from "../types/saveState.ts";
 import { LayoutHeader } from "../puck/components/LayoutHeader.tsx";
 import { DevLogger } from "../../utils/devLogger.ts";
+import { YextEntityFieldSelector } from "../../components/editor/YextEntityFieldSelector.tsx";
 import * as lzstring from "lz-string";
 
 const devLogger = new DevLogger();
@@ -122,7 +123,38 @@ export const InternalLayoutEditor = ({
   return (
     <EntityFieldProvider>
       <Puck
-        config={puckConfig}
+        config={{
+          ...puckConfig,
+          root: {
+            ...puckConfig.root,
+            fields: {
+              title: YextEntityFieldSelector<any, string>({
+                label: "Title",
+                filter: {
+                  types: ["type.string"],
+                },
+              }),
+              description: YextEntityFieldSelector<any, string>({
+                label: "Description",
+                filter: {
+                  types: ["type.string"],
+                },
+              }),
+            },
+            defaultProps: {
+              title: {
+                field: "name",
+                constantValue: "",
+                constantValueEnabled: false,
+              },
+              description: {
+                field: "description",
+                constantValue: "",
+                constantValueEnabled: false,
+              },
+            },
+          },
+        }}
         data={{}} // we use puckInitialHistory instead
         initialHistory={puckInitialHistory}
         onChange={change}


packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
```diff
@@ -122,7 +123,38 @@ export const InternalLayoutEditor = ({
...
+                filter: {
+                  types: ["type.string"],
+                },
+              }),
+            },
+            defaultProps: {
+              title: {
+                field: "name",
+                constantValue: "",
+                constantValueEnabled: false,
```
asanehisa, line 148: does defaultProps work if we have constantValueEnabled to false? also idk if this helps https://puckeditor.com/docs/integrating-puck/root-configuration#setting-default-props

jwartofsky-yext: DefaultProps should work this way. It does for how we use entity fields in other places.

I've also tried with basic defaultProps as described in that link, but they still wouldn't update.

Another issue I'm noticing is that when the root props change, the history does not seem to get updated.

asanehisa: random but r u testing in a starter repo u have? if so make sure the puck version is up to date

jwartofsky-yext: It's on 0.18, but I'll try 0.18.2

asanehisa: oh i'd try 18.1 since thats what ve-lib is on (i think)

jwartofsky-yext: Got it to work

packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
```diff
@@ -101,28 +104,69 @@ export const InternalLayoutEditor = ({
...
-  const change = async () => {
+  const change = async (data: any) => {
     if (isLoading) {
       return;
     }
     if (!canEdit) {
       setCanEdit(true);
       return;
     }
+    const rootString = JSON.stringify(data.root);
```
benlife5, line 127: Is this needed? If the problem is that the puck config object is regenerating every render, try injecting the title/description fields into puckConfig in a useMemo that has a dependency on the puckConfig prop

packages/visual-editor/src/utils/getPageMetadata.ts
```diff
@@ -0,0 +1,31 @@
+import { resolveYextEntityField } from "./resolveYextEntityField.ts";
+
+export type RootConfig = {
+  title: string;
+  description: string;
+  [key: string]: any;
+};
+
+export function getPageMetadata(document: any): RootConfig {
```
benlife5, line 9: Add to readme

packages/visual-editor/src/utils/getPageMetadata.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (!layoutString) {
+    return emptyMetaData;
+  }
+  const layout = JSON.parse(layoutString);
+  const root = layout?.root;
+  if (!root) {
+    return emptyMetaData;
+  }
+  const metaData: RootConfig = {
+    title: resolveYextEntityField(document, root.title) ?? "",
```
benlife5, line 24: I think resolveYextEntityField handles undefined values so you don't need to do the emptyMetaData stuff above

packages/visual-editor/src/utils/getPageMetadata.ts
```diff
@@ -0,0 +1,31 @@
...
+    description: resolveYextEntityField(document, root.description) ?? "",
+  };
+  Object.keys(root).forEach((key: string) => {
```
benlife5, line 27: Shouldn't this handle title/description without needing to do them separately?

packages/visual-editor/src/utils/getPageMetadata.ts
```diff
@@ -0,0 +1,31 @@
+import { resolveYextEntityField } from "./resolveYextEntityField.ts";
+
+export type RootConfig = {
+  title: string;
+  description: string;
+  [key: string]: any;
+};
+
+export function getPageMetadata(document: any): RootConfig {
+  const emptyMetaData = {
```
mkilpatrick, line 10: Just return an empty object. Defining keys here forces us to update this code as new keys are added. The caller can null check. I mean technically other parts of this code would change too so maybe naming them isn't a bad idea after all? @benlife5 thoughts? The only problem is remembering to update this as we add more options which I don't love.

benlife5: I don't think needing to update the code here is a big deal but I would tend towards returning undefined if there's no value so that the caller has to decide how to handle that situation and can use `??`

packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
jwartofsky-yext: This onChange function triggers any time there is a change to the puck config. We want to minimize state updates because every time the state updates, inputs get deselected. If we update the state whenever the root changes, that will make the inputs unusable in the root.

The variables here make it so that we only update the state once to mark that we have root changes, which enables Publish and Clear Local Changes

If we already have the flag for root changes, then we don't trigger any more state updates.

The puck config already has the root with the title and config inside of it.

I'm not sure what you mean by inject the title and description into the puckConfig. Those get added to the config by puck itself.

packages/visual-editor/src/utils/getPageMetadata.ts
jwartofsky-yext: Done

packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
benlife5: I don't think we should need to track root changes separately if usePuck was outputting root changes into the histories array properly, but it seems like it's not and I know you have that discord thread open.

I was wondering if the field was getting deselected each time because the object reference being passed as the config to <Puck> is changing each render. If so, something like this might help, I'm not sure
```
const config = React.useMemo(() => {
return {
    ...puckConfig,
    root: {
      ...puckConfig.root,
      fields: {
        title: YextEntityFieldSelector<any, string>({
          label: "Title",
          filter: {
            types: ["type.string"],
          },
        }),
...
}, [puckConfig])

<Puck
  config={config}
...
```

jwartofsky-yext: I think we ideally shouldn't have to do any of this state stuff for the root at all. I'm pretty sure there's a bug with Puck.

I removed all of the root state stuff for now.

The root fields are saved, published, and cleared along with the other changes properly. However the Publish and Clear Local Changes buttons won't be clickable unless there are other changes.

jwartofsky-yext:
A workaround we can do for now to allow changes to root to be published/saved/cleared would be to detect changes to root in <Puck> onChanged, then inject them into the history

I will attempt this now

mkilpatrick (approved):

PR #226, opened 2025-02-11
style: rename manger -> manager

asanehisa:


benlife5 (approved):
My bad, thanks for fixing :)

PR #229, opened 2025-02-12
fix: prevent re-renders on entity field tooltip toggle

briantstephan:
This memoizes the children that are passed into the entity field tooltip so that they do not re-render when the entity field tooltip toggle is enabled or disabled.

I tested the change locally and confirmed that nothing re-renders, and the tooltips are cleanly added and removed as expected.

### DIFF
--- a/packages/visual-editor/src/components/editor/EntityField.tsx
+++ b/packages/visual-editor/src/components/editor/EntityField.tsx
@@ -8,12 +8,13 @@ import {
 } from "../../internal/puck/ui/Tooltip.tsx";
 import React from "react";
 
-const MemoizedChildren = React.memo(
-  ({ children }: { children: React.ReactNode }) => {
-    return <>{children}</>;
-  }
-);
-MemoizedChildren.displayName = "MemoizedChildren";
+const MemoizedChildren = React.memo(function MemoizedChildren({
+  children,
+}: {
+  children: React.ReactNode;
+}) {
+  return <>{children}</>;
+});
 
 type EntityFieldProps = {
   displayName?: string;
--- a/packages/visual-editor/src/components/editor/EntityField.tsx
+++ b/packages/visual-editor/src/components/editor/EntityField.tsx
@@ -8,6 +8,13 @@ import {
 } from "../../internal/puck/ui/Tooltip.tsx";
 import React from "react";
 
+const MemoizedChildren = React.memo(
+  ({ children }: { children: React.ReactNode }) => {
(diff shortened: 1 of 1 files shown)


packages/visual-editor/src/components/editor/EntityField.tsx
```diff
@@ -8,6 +8,13 @@ import {
 } from "../../internal/puck/ui/Tooltip.tsx";
 import React from "react";
 
+const MemoizedChildren = React.memo(
+  ({ children }: { children: React.ReactNode }) => {
+    return <>{children}</>;
+  }
+);
+MemoizedChildren.displayName = "MemoizedChildren";
```
mkilpatrick, line 16: What's this for?

briantstephan: Just to eliminate an eslint error that was complaining about the component not having a display name

mkilpatrick: Hmm, it shouldn't. Which rule is it hitting? I don't get errors in the default setup in the playground
https://typescript-eslint.io/play/#ts=5.7.2&fileType=.tsx&code=JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcARFDvmQNwBQduEAdqvALJYgTABeWAEwDCAC2AAbAZWZwAvJiowAdCC4QAFHThx1Abzi4xk6XAC%2BALjj7DEqVmaXseZU-wA5CAKxmAlHIB8VlracJQwAK5QMgA8-ro2xvam0QD0-vTapnQ%2B9CkpcJzcfIKittJKAsCoYOLIAJ5uyKpy5IU8-MJGdsy0DEysEOJYSuIQAObqbcWdZfY5QA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

mkilpatrick (dismissed):

packages/visual-editor/src/components/editor/EntityField.tsx
briantstephan: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/display-name.md

Updated to fix it the way they recommend.

mkilpatrick (approved):

PR #232, opened 2025-02-14
feat: add undo/redo buttons to theme manager

briantstephan:
This adds undo and redo buttons to theme manager. Confirmed locally that the undo and redo buttons work as expected; they are disabled when there is nothing to go back to or go forward to, and they correctly update the index of the theme history. Making an edit while there are changes to "redo" erases those changes and moves the index to the end of the history with the new edit.

### DIFF
--- a/packages/visual-editor/src/internal/puck/components/ThemeHeader.tsx
+++ b/packages/visual-editor/src/internal/puck/components/ThemeHeader.tsx
@@ -64,10 +64,7 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
     if (!themeHistories) {
       return false;
     }
-    if (themeHistories.index === 0) {
-      return false;
-    }
-    return true;
+    return themeHistories.index > 0;
   };
 
   const undo = () => {
@@ -75,8 +72,8 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
       return;
     }
     setThemeHistories({
-      histories: themeHistories?.histories,
-      index: themeHistories?.index - 1,
+      histories: themeHistories.histories,
+      index: themeHistories.index - 1,
     });
   };
 
@@ -84,10 +81,7 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
     if (!themeHistories) {
       return false;
     }
-    if (themeHistories.index === themeHistories.histories.length - 1) {
-      return false;
-    }
-    return true;
+    return themeHistories.index < themeHistories.histories.length - 1;
   };
 
   const redo = () => {
@@ -95,8 +89,8 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
       return;
     }
     setThemeHistories({
-      histories: themeHistories?.histories,
-      index: themeHistories?.index + 1,
+      histories: themeHistories.histories,
+      index: themeHistories.index + 1,
     });
   };
 
--- a/packages/visual-editor/src/internal/puck/components/ThemeHeader.tsx
+++ b/packages/visual-editor/src/internal/puck/components/ThemeHeader.tsx
@@ -9,6 +9,7 @@ import { UIButtonsToggle } from "../ui/UIButtonsToggle.tsx";
 import { ClearLocalChangesButton } from "../ui/ClearLocalChangesButton.tsx";
 import { InitialHistory, usePuck } from "@measured/puck";
 import { ThemeData, ThemeHistories } from "../../types/themeData.ts";
+import { RotateCcw, RotateCw } from "lucide-react";
 
 type ThemeHeaderProps = {
   onPublishTheme: () => Promise<void>;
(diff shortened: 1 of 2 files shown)


packages/visual-editor/src/internal/puck/components/ThemeHeader.tsx
```diff
@@ -59,6 +60,46 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
...
+      histories: themeHistories?.histories,
+      index: themeHistories?.index - 1,
+    });
+  };
+
+  const canRedo = (): boolean => {
+    if (!themeHistories) {
+      return false;
+    }
+    if (themeHistories.index === themeHistories.histories.length - 1) {
```
asanehisa, line 87: maybe themeHistories.histories?.length

packages/visual-editor/src/internal/puck/components/ThemeHeader.tsx
```diff
@@ -59,6 +60,46 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
...
+    }
+    if (themeHistories.index === 0) {
+      return false;
+    }
+    return true;
+  };
+
+  const undo = () => {
+    if (!themeHistories) {
+      return false;
```
asanehisa, line 75: unsure if this is meant to return false or just return

packages/visual-editor/src/internal/puck/components/ThemeHeader.tsx
```diff
@@ -67,6 +108,23 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
       </div>
       <div className="header-center"></div>
       <div className="actions">
+        <Button
```
asanehisa, line 111: Can we add an aria-label for the undo and redo buttons?

packages/visual-editor/src/internal/puck/components/ThemeHeader.tsx
briantstephan: shouldn't be necessary since `histories` should always exist if `themeHistories` exists

packages/visual-editor/src/internal/puck/components/ThemeHeader.tsx
briantstephan: Done

packages/visual-editor/src/internal/puck/components/ThemeHeader.tsx
briantstephan: should just be `return`, fixed

packages/visual-editor/src/internal/puck/components/ThemeHeader.tsx
```diff
@@ -59,6 +60,46 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
...
+      return false;
+    }
+    if (themeHistories.index === 0) {
+      return false;
+    }
+    return true;
+  };
+
+  const undo = () => {
+    if (!themeHistories) {
```
mkilpatrick, line 74: I don't think we need this check if canUndo already has it, plus themeHistories is using optional chaining

packages/visual-editor/src/internal/puck/components/ThemeHeader.tsx
```diff
@@ -59,6 +60,46 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
...
+      return false;
+    }
+    if (themeHistories.index === themeHistories.histories.length - 1) {
+      return false;
+    }
+    return true;
+  };
+
+  const redo = () => {
+    if (!themeHistories) {
```
mkilpatrick, line 94: Same

packages/visual-editor/src/internal/puck/components/ThemeHeader.tsx
```diff
@@ -59,6 +60,46 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
...
+      histories: themeHistories?.histories,
+      index: themeHistories?.index - 1,
+    });
+  };
+
+  const canRedo = (): boolean => {
+    if (!themeHistories) {
+      return false;
+    }
+    if (themeHistories.index === themeHistories.histories.length - 1) {
```
mkilpatrick, line 87: Can this just be `return themeHistories.index < themeHistories.histories.length - 1`?

jwartofsky-yext: If for some reason themeHistories is false, you can add ? and it will always work:

```return themeHistories?.index < themeHistories?.histories.length - 1```

packages/visual-editor/src/internal/puck/components/ThemeHeader.tsx
briantstephan: It's not strictly necessary but TS complains if we don't have it, since it thinks `themeHistories` might be undefined in `setThemeHistories`. Going to leave it since this seems like a cleaner solution than using `??` within `setThemeHistories`.

packages/visual-editor/src/internal/puck/components/ThemeHeader.tsx
briantstephan: Yep, updated

mkilpatrick (approved):

PR #233, opened 2025-02-18
fix: fixed root props not updating or saving properly

jwartofsky-yext:
If root.props is missing from the history state, then root prop changes do not trigger updates to histories or index from usePuck.

By ensuring that there are props on the history state, we ensure that the root props update properly.

Verified that root prop changes update and save properly in local dev mode and in production mode.


jwartofsky-yext:
not needed

Handled by this change: https://github.com/YextSolutions/pages-visual-editor-starter/commit/352be0ba5363921920de3c93f88693dc7aaf106a

PR #234, opened 2025-02-18
feat: add link styling section to theme config

benlife5:
Adds Font Color and Font Size as options for the header and footer section in the theme config, and to a new section called Link Styling.

### DIFF
--- a/packages/visual-editor/src/components/puck/Emails.tsx
+++ b/packages/visual-editor/src/components/puck/Emails.tsx
@@ -13,7 +13,7 @@ import {
   FontSizeSelector,
 } from "../../index.js";
 
-const emailsVariants = cva("list-inside font-body-fontFamily", {
+const emailsVariants = cva("components list-inside font-body-fontFamily", {
   variants: {
     fontSize: {
       default: "",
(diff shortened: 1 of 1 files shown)


starter/theme.config.ts
```diff
@@ -241,7 +241,7 @@ export const themeConfig: ThemeConfig = {
         type: "select",
         plugin: "fontSize",
         options: getFontSizeOptions(),
-        default: "16px",
+        default: "24px",
```
benlife5, line 244: Updated to match what is current in the starter

packages/visual-editor/src/components/puck/Emails.tsx
```diff
@@ -10,25 +10,21 @@ import {
...
   YextEntityFieldSelector,
   getFontWeightOverrideOptions,
+  FontSizeSelector,
 } from "../../index.js";
 
 const emailsVariants = cva("list-inside font-body-fontFamily", {
   variants: {
     fontSize: {
-      default: "text-body-fontSize",
+      default: "",
```
asanehisa, line 19: hmmm i assume we will want a default size

benlife5: I moved it to line 130 so that it could switch between the link default and the body default but you're right that it's confusing -- updated to use the cva way to do that

packages/visual-editor/src/components/puck/Emails.tsx
```diff
@@ -10,25 +10,21 @@ import {
...
-const emailsVariants = cva("list-inside font-body-fontFamily", {
+const emailsVariants = cva("components list-inside font-body-fontFamily", {
   variants: {
     fontSize: {
-      default: "text-body-fontSize",
+      default: "",
       xs: "text-xs",
       sm: "text-sm",
-      medium: "text-base",
+      base: "text-base",
```
mkilpatrick, line 22: Why rename from medium to base? It seems odd that there is no medium now but there is sm, lg, etc. Also I'd think it would be md as shorthand like the others.

benlife5: It's a tailwind thing 🤷
The other components were already switched over when we added the standardized FontSizeSelector

mkilpatrick: Gotcha

mkilpatrick (approved):

asanehisa (approved):

PR #235, opened 2025-02-19
feat: add breadcrumbs component

asanehisa:
- Last item in breadcrumbs is not a link and is same fontSize as everything else but text-body-color
- Breadcrumb links should be underlined with no underline on hover and text-link-color
- no DM = don’t crash, just return nothing
- Separator should be hardcoded to `/` (for now)
- Component **should not** have entity field props but rather be hardcoded to use field that starts with `dm_directoryParents_`
- Add Breadcrumbs to the starter’s main config ([draft PR ](https://github.com/YextSolutions/pages-visual-editor-starter/pull/211))

This will not work before edits to pageset stream are made by Spruce. Won't break anything either.


starter/src/dev.config.ts
```diff
@@ -7901,6 +7905,42 @@ export const devTemplateStream = {
             ],
           },
         },
+        {
```
asanehisa, line 7908: hardcoded for now

packages/visual-editor/src/components/puck/Breadcrumbs.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { useDocument } from "../../index.js";
+import { ComponentConfig } from "@measured/puck";
+import { Key } from "react";
+
+export type BreadcrumbsProps = {
+  separator?: string;
+};
+
+const getDirectoryParents = (
+  document: any
```
benlife5, line 11: you can type this as Record<string, any> for a little more specificity

packages/visual-editor/src/components/puck/Breadcrumbs.tsx
```diff
@@ -0,0 +1,78 @@
...
+): Array<{ slug: string; name: string }> => {
+  for (const key in document) {
+    if (key.startsWith("dm_directoryParents_")) {
```
benlife5, line 14: @mkilpatrick do we need to handle the situation that an entity is in two DMs?

packages/visual-editor/src/components/puck/Breadcrumbs.tsx
```diff
@@ -0,0 +1,78 @@
...
+const Breadcrumb = ({
+  name,
+  href,
+  linkClassName,
+}: {
+  name: string;
+  href: string;
+  linkClassName?: string;
+}) => {
+  if (href) {
```
benlife5, line 30: Consulting had a MaybeLink component that I think could be used here. I put it in my header PR but your PR will probably merge first so you could copy it over if you think it will be useful

packages/visual-editor/src/components/puck/Breadcrumbs.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const document = useDocument<any>();
+  const breadcrumbs = getDirectoryParents(document);
+
+  return (
+    <div>
+      {breadcrumbs?.length && (
+        <nav className="my-4" aria-label="Breadcrumb">
+          <ol className="components flex flex-wrap text-link-fontSize text-body-color">
+            {breadcrumbs.map(
+              ({ name, slug }: { name: string; slug: string }, idx: Key) => {
```
benlife5, line 52: Is this typing necessary? I think breadcrumbs.map may do it automatically

packages/visual-editor/src/components/puck/Breadcrumbs.tsx
```diff
@@ -0,0 +1,78 @@
...
+            )}
+          </ol>
+        </nav>
+      )}
+    </div>
+  );
+};
+
+export const Breadcrumbs: ComponentConfig<BreadcrumbsProps> = {
+  render: (props) => <BreadcrumbsComponent {...props} />,
```
benlife5, line 77: Does label get implied because exported variable is Breadcrumbs?

starter/src/ve.config.tsx
```diff
@@ -64,6 +67,7 @@ export const mainConfig: Config<MainProps> = {
     Address,
     Banner,
     BodyText,
+    Breadcrumbs,
```
benlife5, line 70: Update `visual-editor/packages/visual-editor/src/components/puck/registry/components.ts` too

packages/visual-editor/src/components/puck/Breadcrumbs.tsx
asanehisa: Hmmmm I don't know how I'd handle that besides picking one randomly (still this has no props for user to set)

packages/visual-editor/src/components/puck/Breadcrumbs.tsx
asanehisa: Yup rn label is just Breadcrumbs

packages/visual-editor/src/components/puck/Breadcrumbs.tsx
asanehisa: oh ur right, it wasn't at some point haha

packages/visual-editor/src/components/puck/Breadcrumbs.tsx
asanehisa: Ok! Using exact same as urs

packages/visual-editor/src/components/puck/Breadcrumbs.tsx
mkilpatrick: I'm hoping that Spruce only puts the appropriate DM fields in the stream. If that's the case then this is fine as there can only be one DM per pageset.

https://yext.slack.com/archives/CC7B2751A/p1739996805528379?thread_ts=1739983888.459289&cid=CC7B2751A

mkilpatrick (approved):

PR #236, opened 2025-02-19
refactor: header changes

benlife5:
Updates the header to use the fields from the site stream and mostly use consulting's code, including a mobile layout.

Need to test more after the site stream stuff is completed, because the mobile menu is inaccessible without in-platform dev and I don't think the logo height/width is quite right for pulling from content


packages/visual-editor/src/components/puck/Header.tsx
```diff
@@ -85,97 +51,138 @@ const headerFields: Fields<HeaderProps> = {
...
-              </li>
-            ))}
-          </ul>
-        </div>
+
+        <EntityField displayName="Header Links" fieldId={"site.header.links"}>
+          <HeaderLinks links={links} />
+        </EntityField>
+
+        <button
```
asanehisa, line 108: this prob needs an aria-label?

packages/visual-editor/src/components/puck/Header.tsx
```diff
@@ -85,97 +51,138 @@ const headerFields: Fields<HeaderProps> = {
...
+          className="flex md:hidden absolute p-4 right-0 top-1/2 -translate-y-1/2"
```
asanehisa, line 109: -translate-y-1/2 ?

packages/visual-editor/src/components/puck/Header.tsx
```diff
@@ -85,97 +51,138 @@ const headerFields: Fields<HeaderProps> = {
...
+        `${isOpen ? "visible" : "hidden"} ${headerVariants({ backgroundColor })}` +
+        "absolute top-full left-0 right-0 h-screen z-50"
+      }
+    >
+      <div className="container">
+        <ul className="flex flex-col">
+          {links.map((item: CTA, idx) => (
+            <li key={item.link}>
+              <Link
+                className="py-3 block text-header-linkColor text-header-linkFontSize"
```
asanehisa, line 178: i think you'll need "components"

packages/visual-editor/src/components/puck/assets/fa_bars_classic_solid_v5.svg
asanehisa, line 1: oh did we wanna add these assets to our repo rather than just using lucide react's stuff?

asanehisa (commented):
is testing via the visual-editor/starter flow with pnpm run dev possible?

packages/visual-editor/src/components/puck/Header.tsx
benlife5: yeah idk why the toggle would move a little when the menu options... maybe because the icons are different proportions? I'll test once I can get in platform

benlife5:
> Oh can u add theme changes to the visual-editor/starter package?

https://github.com/YextSolutions/pages-visual-editor-starter/pull/210

packages/visual-editor/src/components/puck/assets/fa_bars_classic_solid_v5.svg
benlife5: Switched to the same library you used

benlife5:
@mkilpatrick you were right -- can test mobile via Puck's interactive mode
Screenshot 2025-02-20 at 11 38 16 AM

asanehisa (dismissed):

mkilpatrick (approved):

packages/visual-editor/src/components/puck/Header.tsx
```diff
@@ -85,97 +50,139 @@ const headerFields: Fields<HeaderProps> = {
...
+          width={props.logo.image.width}
+        />
+      </div>
+    </MaybeLink>
+  );
+};
+
+const HeaderLinks = (props: { links: CTA[] }) => {
+  return (
+    <div className="hidden md:flex items-center">
```
mkilpatrick, line 142: I had a similar comment on Alexis' PR. If there are no items we shouldn't render the div at all. You'll need to check that everything else is styled correctly in this case. And same for any other fields, like no logo. I also imagine if there are no links at all then we wouldn't want to render the mobile menu thing.

benlife5:
Force pushed because the diff was being weird. Updated with image size handling and more null checking for links. Tested in a site with a hard coded site stream, seems good to go. I can try to set up a site for Aaron to qa header and footer once this is merged

mkilpatrick (approved):

asanehisa (approved):

PR #238, opened 2025-02-20
fix: editor flickers when saving state

benlife5:
I'm not really sure why this fixes the flickering but it does. The problem seemed tied to layoutSaveState finishing and the config shouldn't be related, but Puck's rendering is often a little weird

Non-flickering example: https://www.yext.com/s/4259018/yextsites/153063/editor#pagesetId=newloc5&layoutId=newloc5-default-layout&themeId=ve-5&entityId=1095527880

### DIFF
--- a/packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
+++ b/packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
@@ -120,41 +120,45 @@ export const InternalLayoutEditor = ({
     }
   };
 
+  const puckConfigWithRootFields = React.useMemo(() => {
+    return {
+      ...puckConfig,
+      root: {
+        ...puckConfig.root,
+        fields: {
+          title: YextEntityFieldSelector<any, string>({
+            label: "Title",
+            filter: {
+              types: ["type.string"],
+            },
+          }),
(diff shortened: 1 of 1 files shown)


jwartofsky-yext:
Nice

packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
```diff
@@ -120,41 +120,45 @@ export const InternalLayoutEditor = ({
...
-                  types: ["type.string"],
-                },
-              }),
-              description: YextEntityFieldSelector<any, string>({
-                label: "Description",
-                filter: {
-                  types: ["type.string"],
-                },
-              }),
+  const puckConfigWithRootFields = React.useMemo(() => {
```
mkilpatrick, line 123: Do we need to do this with theme too?

mkilpatrick (approved):

packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
benlife5: It wasn't happening in the theme editor, I think because the config isn't modified

PR #240, opened 2025-02-25
fix: footer social media links

asanehisa:
Prepend the correct base URLs for instagram and twitter as they are handles (rather than URLs)
- Instagram uses instagram.com/{handle}
- Twitter uses twitter.com/{username}

Rest of site entity has URL in the fieldName and takes in a URL.

### DIFF
--- a/packages/visual-editor/src/components/puck/Footer.tsx
+++ b/packages/visual-editor/src/components/puck/Footer.tsx
@@ -61,37 +61,51 @@ const FooterComponent: React.FC<FooterProps> = (props) => {
   const socialLinks = [
     {
       name: "facebook",
-      link: document?._site?.facebookPageUrl,
+      link: document?._site?.facebookPageUrl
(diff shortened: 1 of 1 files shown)


packages/visual-editor/src/components/puck/Footer.tsx
```diff
@@ -81,7 +83,9 @@ const FooterComponent: React.FC<FooterProps> = (props) => {
     },
     {
       name: "twitter",
-      link: document?._site?.twitterHandle,
+      link: document?._site?.twitterHandle
+        ? `https://twitter.com/${document._site.twitterHandle}`
```
jwartofsky-yext, line 87: Shouldn't we also do this for pinterest, tiktok, etc?

asanehisa: those seem to be URLs in the site entity rather than handles

PR #244, opened 2025-02-26
refactor: update components to use Link

jwartofsky-yext:
Updates GetDirections to use CTA instead of a Button + Link combo.

Updates various other components to add a Link Type prop or set one by default.

Fixes a VE crash when entering a blank constant value for Promo or Card CTA link.

Verified that the appearance and functionality of the components are not any different than before.

### DIFF
--- a/packages/visual-editor/src/components/puck/Card.tsx
+++ b/packages/visual-editor/src/components/puck/Card.tsx
@@ -382,7 +382,7 @@ const CardWrapper = ({
             <CTA
               variant={cta.variant}
               label={resolvedCTA.label ?? ""}
-              link={resolvedCTA?.link || "#"}
+              link={resolvedCTA.link || "#"}
               fontSize={cta.fontSize}
               linkType={cta.linkType}
             />
--- a/packages/visual-editor/src/components/puck/Card.tsx
+++ b/packages/visual-editor/src/components/puck/Card.tsx
@@ -13,7 +13,7 @@ import {
   getFontWeightOverrideOptions,
 } from "../../index.js";
 import { Body } from "./atoms/body.js";
-import { CTA, CTAProps } from "./atoms/cta.js";
+import { CTA, CTAProps, linkTypeFields } from "./atoms/cta.js";
 import {
   Heading,
   HeadingProps,
@@ -58,6 +58,7 @@ interface CardProps {
   };
   cta: {
     entityField: YextEntityField<CTAProps>;
+    linkType: CTAProps["linkType"];
     variant: CTAProps["variant"];
     fontSize: CTAProps["fontSize"];
   };
@@ -271,6 +272,7 @@ const cardFields: Fields<CardProps> = {
         ],
       },
       fontSize: FontSizeSelector(),
+      linkType: linkTypeFields,
     },
   },
   backgroundColor: {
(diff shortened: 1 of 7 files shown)


benlife5 (approved):

packages/visual-editor/src/components/puck/GetDirections.tsx
```diff
@@ -92,17 +92,15 @@ const GetDirections = ({
       fieldId={coordinateField.field}
       constantValueEnabled={coordinateField.constantValueEnabled}
     >
-      <Button
-        asChild
-        variant={variant}
+      <CTA
+        label={"Get Directions"}
+        link={searchQuery ?? "#"}
+        linkType={"URL"}
```
mkilpatrick, line 98: There is an underlying DRIVING_DIRECTIONS linkType for the Link component. You'll need to update CTA accordingly (there are other linkTypes as well).

packages/visual-editor/src/components/puck/GetDirections.tsx
```diff
@@ -94,8 +94,8 @@ const GetDirections = ({
     >
       <CTA
         label={"Get Directions"}
-        link={searchQuery ?? "#"}
-        linkType={"URL"}
+        link={searchQuery || "#"}
+        linkType={"DRIVING_DIRECTIONS"}
```
benlife5, line 98: Update this for the Address component's get directions too

Otherwise LGTM

benlife5 (approved):

packages/visual-editor/src/components/puck/Card.tsx
```diff
@@ -379,9 +381,10 @@ const CardWrapper = ({
           {resolvedCTA && (
             <CTA
               variant={cta.variant}
-              label={resolvedCTA.name}
-              link={resolvedCTA.link ?? "#"}
+              label={resolvedCTA.label ?? ""}
+              link={resolvedCTA?.link || "#"}
```
mkilpatrick, line 385: You already check resolvedCTA exists on 381 so you shouldn't need the first question mark. But don't you need to use || above too? Why would it be necessary here and not above, or vice versa?

packages/visual-editor/src/components/puck/Promo.tsx
```diff
@@ -261,8 +263,9 @@ const PromoWrapper: React.FC<PromoProps> = ({
           {resolvedCTA && (
             <CTA
               variant={cta.variant}
-              label={resolvedCTA.name}
-              link={resolvedCTA.link ?? "#"}
+              label={resolvedCTA.label ?? ""}
+              link={resolvedCTA?.link || "#"}
```
mkilpatrick, line 267: Same first question mark comment

packages/visual-editor/src/components/puck/atoms/cta.tsx
```diff
@@ -13,6 +14,19 @@ export interface CTAProps
   fontSize?: ButtonProps["fontSize"];
 }
 
+const linkTypeFields: Field = {
+  type: "select",
+  label: "Link Type",
+  options: [
+    { label: "URL", value: "URL" },
+    { label: "Email", value: "EMAIL" },
+    { label: "Phone", value: "PHONE" },
```
mkilpatrick, line 23: Is there not a component that shows a phone number somewhere?

packages/visual-editor/src/components/puck/Card.tsx
jwartofsky-yext: Label can be null or empty without it crashing

The editor will crash if link is empty though, so that's why we need || since it will choose "#" if resolvedCTA.link is an empty string

packages/visual-editor/src/components/puck/atoms/cta.tsx
jwartofsky-yext: There's a Phone component, but it doesn't have a Link in it

I'll ask in the visual-editor channel if we want to add one.

If so, then we would use this there.

mkilpatrick (approved):

PR #245, opened 2025-02-26
feat: add useTemplateProps

asanehisa:


starter/src/templates/dev.tsx
```diff
@@ -116,7 +116,7 @@ const Dev: Template<TemplateRenderProps> = (props) => {
       </div>
       <div>
         <VisualEditorProvider
-          document={document}
+          templateProps={props}
```
mkilpatrick, line 119: Why did pnpm-lock change?

asanehisa: ah was on accident

PR #247, opened 2025-03-04
fix: prevent loading bar from going backwards

briantstephan:
This fixes an issue that caused the progress bar to go backwards at times. It also adds a new `useProgress` hook that provides `isLoading` and `progress` values when given a set of completion criteria so we can more easily work with the progress bars in the future.

https://github.com/user-attachments/assets/4daf5508-d844-4d31-9c12-eb096bf4f879

### DIFF
--- a/packages/visual-editor/src/components/editor/Editor.tsx
+++ b/packages/visual-editor/src/components/editor/Editor.tsx
@@ -13,6 +13,7 @@ import { useCommonMessageReceivers } from "../../internal/hooks/useMessageReceiv
 import { LayoutEditor } from "../../internal/components/LayoutEditor.tsx";
 import { ThemeEditor } from "../../internal/components/ThemeEditor.tsx";
 import { useCommonMessageSenders } from "../../internal/hooks/useMessageSenders.ts";
+import { useProgress } from "../../internal/hooks/useProgress.ts";
 
 const devLogger = new DevLogger();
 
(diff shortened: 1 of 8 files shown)


packages/visual-editor/src/internal/puck/components/LoadingScreen.tsx
```diff
@@ -8,9 +8,17 @@ export type LoadingScreenProps = {
 };
 
 export function LoadingScreen({ progress }: LoadingScreenProps) {
+  // This ensures that the progress bar never goes backwards.
+  const [maxProgress, setMaxProgress] = React.useState(0);
+  React.useEffect(() => {
```
mkilpatrick, line 13: Would love to figure out the real cause. @benlife5 any ideas?

mkilpatrick (approved):

packages/visual-editor/src/internal/puck/components/LoadingScreen.tsx
jwartofsky-yext: To figure out the real cause, you could add some logging so all of the variables that go into calculating the percentage bar get logged whenever one of them changes.

Then you can see which one is going backwards

briantstephan: I wasn't able to replicate this outside of this prod site, so I couldn't add logging outside of the existing xYextDebug logs. I did get this screen recording with network activity that shoes the loading bar going backwards when `savestate` finishes getting fetched, but I still can't figure out in the code why that would result in negative progress.

https://github.com/user-attachments/assets/35f3214c-ce03-474d-b22e-f6755dc279d6

jwartofsky-yext: I see the backwards loading bar behavior here on dev: https://dev.yext.com/s/1000146856/yextsites/62564/editor#pagesetId=location10&layoutId=location10-default-layout&themeId=jwartofskytheme&entityId=8927722&locale=en

It's currently getting an error fetching document, but I see the bar go backwards first

jwartofsky-yext: A better, working site where the bar goes backwards but then it loads properly:

https://dev.yext.com/s/1000146856/yextsites/63539/editor#pagesetId=jwartofskylocationmarch4&layoutId=jwartofskylocationmarch4-default-layout&themeId=jwartofskygetstartedmarch2024&entityId=8927722

briantstephan: The issue was caused by using `entityFields === null` as a criteria for `isLoading` but also using it to calculate the amount of progress, rather than using `entityFields !== null` for progress. I just made a new `useProgress` hook that takes in a set of success criteria and provides `isLoading` and `progress` values automatically so we don't need to do the math ourselves.

packages/visual-editor/src/components/editor/Editor.tsx
```diff
@@ -104,28 +105,22 @@ export const Editor = ({
...
+      themeDataFetched,
+      entityFields !== null,
+    ],
+  });
 
   return (
     <ErrorBoundary fallback={<></>} onError={logError}>
-      {!isLoading ? (
-        templateMetadata.isThemeMode || forceThemeMode ? (
+      {!isLoading && puckConfig && templateMetadata ? (
```
benlife5, line 122: Why does puckConfig && templateMetadata need to be added here?

benlife5 (commented):
My bad about the entityFields null check. Thanks for looking into this, I think the hook is much nicer

packages/visual-editor/src/components/editor/Editor.tsx
briantstephan: The downside to the hook is TS no longer recognizes that `isLoading` being false means `puckConfig` and `templateMetadata` must not be undefined, since it's abstracted away within `useProgress`. Talked offline and we can add `!` to the variables as we pass them into `ThemeEditor` and `LayoutEditor`, which seems like a cleaner solution.

mkilpatrick (approved):

PR #251, opened 2025-03-05
refactor: localstorage updates

benlife5:
A few small changes to local storage to help dev mode 

1. Use a local variable for sendDevLayoutSaveStateData rather than the state variable. Often the state variable wasn't updated by the time sendDevLayoutSaveStateData was called
2. I think we want to generate the localstorage key per layout not per entity, by default. It becomes weird if you make some dev mode changes to the layout, then go to theme editor and they are only applied to some entities. I've kept the entity key so it can be used for user-based permissions later
3. Where possible rename `themeEntityId` to `themeId`


packages/visual-editor/src/internal/components/ThemeEditor.tsx
```diff
@@ -29,7 +29,7 @@ type ThemeEditorProps = {
   layoutData: Data;
   themeData: ThemeData;
   themeConfig: ThemeConfig | undefined;
-  localDev: boolean;
+  standAloneMode: boolean;
```
mkilpatrick, line 32: So this means you're running locally (npm run dev)? And it was confusing to have "localDev" vs the isDevMode flag from templateMetadata? The new name seems more confusing to me.

benlife5: the `localDev` flag meant running pnpm run dev in the visual-editor repo
`templateMetadata.isDevMode` means you've run npm run dev in your starter and are using dev mode through the platform
I'm fine to keep the old name we just need to remember localDev !== isDevMode

jwartofsky-yext: in this case "localDev" was running the local repo in the visual-editor/starter

isDevMode from templateMetaData is "real local dev"

both are true when running visual-editor/starter

maybe "standAloneMode" could be called "yextDevMode", but I think standAloneMode is also a good name

jwartofsky-yext (dismissed):
lgtm

packages/visual-editor/src/internal/components/ThemeEditor.tsx
mkilpatrick: So it's really localDev vs platformLocalDev. localDev means just running against any starter (which includes the fako starter in this lib). isDevMode signifies the dev toggle is flipped on in platform. Open for discussion amongst the team, but I still find it less confusing before changing the name to standAloneMode.

asanehisa (dismissed):

packages/visual-editor/src/internal/components/ThemeEditor.tsx
asanehisa: Hmmm no big pref from me but some comments / documentation to indicate standAloneMode's meaning would be nice

mkilpatrick (approved):

PR #252, opened 2025-03-06
feat: add yextVisualEditorPlugin()

asanehisa:
On build, it creates the edit and main files in src/templates/ then removes them. 

Future followups: 
- Support devMode
- Replace getMainTemplate() and getEditTemplate() with edit.tsx and main.tsx files whose contents can be copied into the ones generated during build. 
- Support dynamically importing themeConfig and allowing no themeConfig
- Support scope
- Support library upgrades (puck, etc)

### DIFF
--- a/packages/visual-editor/package.json
+++ b/packages/visual-editor/package.json
@@ -26,6 +26,7 @@
   },
   "scripts": {
     "build": "tsc && vite build",
+    "buildAll": "tsc && vite build && vite build --config vite.config.plugin.ts",
     "test": "vitest run",
     "prettier": "prettier --write --cache .",
     "lint": "eslint --cache --fix src/",
--- a/packages/visual-editor/package.json
+++ b/packages/visual-editor/package.json
@@ -25,8 +25,9 @@
     "./style.css": "./dist/style.css"
   },
   "scripts": {
-    "build": "tsc && vite build",
-    "buildAll": "tsc && vite build && vite build --config vite.config.plugin.ts",
+    "build:components": "tsc && vite build",
+    "build:plugin": "tsc && vite build --config vite.config.plugin.ts",
+    "build": "tsc && vite build && vite build --config vite.config.plugin.ts",
     "test": "vitest run",
     "prettier": "prettier --write --cache .",
     "lint": "eslint --cache --fix src/",
--- a/packages/visual-editor/src/vite-plugin/getTemplates.ts
+++ b/packages/visual-editor/src/vite-plugin/getTemplates.ts
@@ -1,0 +1,178 @@
+export const getEditTemplate = (): string => {
+  return `import {
+  applyTheme,
+  Editor,
+  usePlatformBridgeDocument,
+  usePlatformBridgeEntityFields,
+  VisualEditorProvider,
+} from "@yext/visual-editor";
+import { componentRegistry } from "../ve.config";
+import {
+  GetPath,
+  TemplateProps,
+  TemplateConfig,
+  HeadConfig,
+  TemplateRenderProps,
+  GetHeadConfig,
+} from "@yext/pages";
+import { themeConfig } from "../../theme.config";
+import tailwindConfig from "../../tailwind.config";
+
+// Editor is avaliable at /edit
+export const getPath: GetPath<TemplateProps> = () => {
+  return "edit";
+};
+
+export const config: TemplateConfig = {
+  name: "edit",
+};
+
+export const getHeadConfig: GetHeadConfig<TemplateRenderProps> = ({
+  document,
+}): HeadConfig => {
+  return {
+    other: applyTheme(document, themeConfig),
+  };
+};
+
+// Render the editor
+const Edit: () => JSX.Element = () => {
+  const entityDocument = usePlatformBridgeDocument();
+  const entityFields = usePlatformBridgeEntityFields();
+
+  return (
+    <VisualEditorProvider
+      templateProps={{
+        document: entityDocument
+      }}
+      entityFields={entityFields}
+      tailwindConfig={tailwindConfig}
+    >
+      <Editor
+        document={entityDocument}
+        componentRegistry={componentRegistry}
+        themeConfig={themeConfig}
+      />
+    </VisualEditorProvider>
+  );
+};
+
+export default Edit;
+`;
+};
+
+export const getMainTemplate = (): string => {
+  return `import "@yext/visual-editor/style.css";
+import {
+  Template,
+  GetPath,
+  TemplateProps,
+  TemplateRenderProps,
+  GetHeadConfig,
+  HeadConfig,
+  TagType,
+} from "@yext/pages";
+import { Render } from "@measured/puck";
+import { mainConfig } from "../ve.config";
+import {
+  applyTheme,
+  VisualEditorProvider,
+  normalizeSlug,
+  getPageMetadata,
+  applyAnalytics,
+} from "@yext/visual-editor";
+import { themeConfig } from "../../theme.config";
+import { buildSchema } from "../utils/buildSchema.js";
+import { AnalyticsProvider } from "@yext/pages-components";
+
+export const getHeadConfig: GetHeadConfig<TemplateRenderProps> = ({
+  document,
+}): HeadConfig => {
+  const { title, description } = getPageMetadata(document);
+  const faviconUrl = document?._site?.favicon?.url;
+
+  return {
+    title: title,
+    charset: "UTF-8",
+    viewport: "width=device-width, initial-scale=1",
+    tags: [
+      {
+        type: "link",
+        attributes: {
+          rel: "icon",
+          type: "image/x-icon",
+        },
+      },
+      ...(description
+          ? [
+            {
+              type: "meta" as TagType,
+              attributes: {
+                name: "description",
+                content: description,
+              },
+            },
+          ]
+          : []),
+      ...(faviconUrl
+        ? [
+            {
+              type: "link" as TagType,
+              attributes: {
+                rel: "icon",
+                type: "image/x-icon",
+                href: faviconUrl,
+              },
+            },
+          ]
+        : []),
+    ],
+    other: [
+      applyAnalytics(document),
+      applyTheme(document, themeConfig),
+      buildSchema(document),
+    ].join("\\n"),
+  };
+};
+
+export const getPath: GetPath<TemplateProps> = ({ document }) => {
+  if (!document?.__?.layout) {
+    // temporary: guard for generated repo-based static page
+    return \`static-\${Math.floor(Math.random() * (10000 - 1))}\`;
+  }
+
+  if (document.slug) {
+    return document.slug;
+  }
+
+  const localePath = document.locale !== "en" ? \`\${document.locale}/\` : "";
+  const path = document.address
+    ? \`\${localePath}\${document.address.region}/\${document.address.city}/\${document.address.line1}\`
+    : \`\${localePath}\${document.id}\`;
+
+  return normalizeSlug(path);
+};
+
+const Location: Template<TemplateRenderProps> = (props) => {
+  const { document } = props;
+  // temporary: guard for generated repo-based static page
+  if (!document?.__?.layout) {
+    return <></>;
+  }
+
+  return (
+    <AnalyticsProvider
+      apiKey={document?._env?.YEXT_PUBLIC_EVENTS_API_KEY}
+      templateData={props}
+      currency="USD"
+    >
+      <VisualEditorProvider templateProps={props}>
+        <Render config={mainConfig} data={JSON.parse(document.__.layout)} />
+      </VisualEditorProvider>
+    </AnalyticsProvider>
+  );
+};
+
+export default Location;
+`;
+};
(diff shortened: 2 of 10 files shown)


packages/visual-editor/package.json
```diff
@@ -25,7 +25,9 @@
     "./style.css": "./dist/style.css"
   },
   "scripts": {
-    "build": "tsc && vite build",
+    "build:components": "tsc && vite build",
+    "build:plugin": "tsc && vite build --config vite.config.plugin.ts",
```
mkilpatrick, line 29: I don't think the plugin needs tsc, but if it does, how does it know to use tsconfig.plugin.json?

packages/visual-editor/src/vite-plugin/getTemplates.ts
```diff
@@ -0,0 +1,178 @@
+export const getEditTemplate = (): string => {
```
mkilpatrick, line 1: Why functions and not just const strings?

packages/visual-editor/src/vite-plugin/plugin.ts
```diff
@@ -0,0 +1,42 @@
+import path from "node:path";
+import fs from "fs-extra";
+import { Plugin } from "vite";
+import { getEditTemplate, getMainTemplate } from "./getTemplates.ts";
+
+export const yextVisualEditorPlugin = (): Plugin => {
+  if (!process.env.PLUGIN) {
```
mkilpatrick, line 7: We shouldn't need this. Can you explain the purpose again?

packages/visual-editor/src/vite-plugin/plugin.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  return {
+    name: "vite-plugin-yext-visual-editor",
+    apply: "build",
+    buildStart() {
+      console.log("\nGenerating visual editor files...");
+      Object.entries(virtualFiles).forEach(([fileName, content]) => {
+        const filePath = path.join(process.cwd(), fileName);
+        filesToCleanup.push(filePath);
+        fs.ensureFileSync(filePath);
```
mkilpatrick, line 29: Why do we need this? writeFileSync will create or overwrite

packages/visual-editor/src/vite-plugin/plugin.ts
```diff
@@ -0,0 +1,42 @@
...
+  const virtualFiles: Record<string, string> = {
+    "src/templates/main.tsx": getMainTemplate(),
+    "src/templates/edit.tsx": getEditTemplate(),
+  };
+
+  return {
+    name: "vite-plugin-yext-visual-editor",
+    apply: "build",
+    buildStart() {
+      console.log("\nGenerating visual editor files...");
```
mkilpatrick, line 25: Don't log anything, same below

packages/visual-editor/package.json
```diff
@@ -25,7 +25,9 @@
...
+    "build": "tsc && vite build && vite build --config vite.config.plugin.ts",
```
mkilpatrick, line 30: build:components && build:plugin

packages/visual-editor/src/vite-plugin/plugin.ts
```diff
@@ -0,0 +1,42 @@
...
+      Object.entries(virtualFiles).forEach(([fileName, content]) => {
+        const filePath = path.join(process.cwd(), fileName);
+        filesToCleanup.push(filePath);
+        fs.ensureFileSync(filePath);
+        fs.writeFileSync(filePath, content);
+        console.log(`Created: ${filePath}`);
+      });
+    },
+    buildEnd() {
```
mkilpatrick, line 34: How can we clean up if there's an error or the build is interrupted (cmd+c)?

packages/visual-editor/vite.config.plugin.ts
```diff
@@ -0,0 +1,27 @@
+import { defineConfig } from "vite";
+import path from "node:path";
+
+export default defineConfig(() => ({
+  define: {
+    "process.env.PLUGIN": "true",
+  },
+  css: {
+    postcss: "./postcss.config.cts",
```
mkilpatrick, line 9: What's this for?

packages/visual-editor/vite.config.plugin.ts
```diff
@@ -0,0 +1,27 @@
+import { defineConfig } from "vite";
+import path from "node:path";
+
+export default defineConfig(() => ({
+  define: {
+    "process.env.PLUGIN": "true",
```
mkilpatrick, line 6: Shouldn't need

packages/visual-editor/vite.config.plugin.ts
```diff
@@ -0,0 +1,27 @@
...
+  },
+  css: {
+    postcss: "./postcss.config.cts",
+  },
+  build: {
+    emptyOutDir: false,
```
mkilpatrick, line 12: Should be true

packages/visual-editor/vite.config.plugin.ts
```diff
@@ -0,0 +1,27 @@
...
+      entry: {
+        plugin: path.resolve(__dirname, "src/vite-plugin/index.ts"),
+      },
+      name: "visual-editor-vite-plugin",
+      formats: ["es", "cjs"] as ["es", "cjs"], // TypeScript fix
+    },
+    target: "node18",
+    tsconfig: path.resolve(__dirname, "tsconfig.plugin.json"),
+    rollupOptions: {
+      external: ["node:path", "fs-extra"],
```
mkilpatrick, line 24: Was this necessary?

packages/visual-editor/vite.config.ts
```diff
@@ -5,14 +5,27 @@ import type { Plugin } from "vite";
...
+  define: {
+    'process.env.PLUGIN': 'false',
+  },
   plugins: [react(), dts()],
   build: {
     lib: {
-      entry: path.resolve(__dirname, "src/index.ts"),
+      entry: {
+        "visual-editor": path.resolve(__dirname, "src/index.ts"),
+        "plugin/plugin": path.resolve(__dirname, "src/vite-plugin/index.ts"),
```
mkilpatrick, line 16: You have a separate vite config for the plugin. This one shouldn't also be building the plugin.

packages/visual-editor/vite.config.ts
```diff
@@ -5,14 +5,27 @@ import type { Plugin } from "vite";
...
+      },
       name: "visual-editor",
       formats: ["es", "cjs"] as LibraryFormats[], // typescript is unhappy without this forced type definition
     },
     rollupOptions: {
+      onLog(level, log, handler) {
```
mkilpatrick, line 22: I don't think anything about this config should change

packages/visual-editor/package.json
asanehisa: Yup doesn't need rn, removed. If in future we add tsc for plugin, we'd do like tsc --config tsconfig.plugin.json

packages/visual-editor/src/vite-plugin/getTemplates.ts
asanehisa: Fixed

packages/visual-editor/vite.config.plugin.ts
asanehisa: Ah removed.

packages/visual-editor/src/vite-plugin/plugin.ts
asanehisa: fixed. added a SIGINT catcher

packages/visual-editor/src/vite-plugin/plugin.ts
asanehisa: Ah removed.

packages/visual-editor/vite.config.plugin.ts
asanehisa: Oh without it, I get the `[plugin vite:resolve] Module "node:path" has been externalized for browser compatibility... ` warnings, with the external code, it properly uses node target for path and fs.

mkilpatrick: Nvm, this is the new plugin config. But I thought target makes the build context for node which means things don't need to be externalized for browser. Like isn't that the point?

starter/vite.config.js
```diff
@@ -1,7 +1,8 @@
 import { defineConfig } from "vite";
 import react from "@vitejs/plugin-react";
 import yextSSG from "@yext/pages/vite-plugin";
+import { yextVisualEditorPlugin } from "@yext/visual-editor/plugin";
```
mkilpatrick, line 4: I was hoping to have the import at `"@yext/visual-editor` and not subpathed

packages/visual-editor/src/vite-plugin/getTemplates.ts
```diff
@@ -0,0 +1,176 @@
+export const editTemplate = `/** THIS FILE IS AUTOGENERATED AND SHOULD NOT EDITED */
```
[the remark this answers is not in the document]
asanehisa: Hmmm is it at the top of the file?

starter/vite.config.js
asanehisa: Ahh yeah left a mention of it in my PR description but kept running into roadblocks with having it not subpathed due to SSR vs non-SSR mode.

packages/visual-editor/src/vite-plugin/plugin.ts
```diff
@@ -0,0 +1,62 @@
...
+  process.on("SIGINT", () => {
+    cleanupFiles();
+    setTimeout(() => process.exit(0), 100); // without this cleanup doesn't finish before process.exit
+  });
+
+  return {
+    name: "vite-plugin-yext-visual-editor",
+    apply: "serve",
+    configureServer(server) {
+      if (isBuildMode) return;
```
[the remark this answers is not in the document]
asanehisa: Hmm I saw the build hooks running on serve (when running dev mode) so I had to add this or else the files would get generated and deleted before the server closed.

packages/visual-editor/src/vite-plugin/templates/main.tsx
```diff
@@ -1,66 +1,5 @@
...
-    </VisualEditorProvider>
-  );
-};
-
-export default Edit;
-`;
-
-export const mainTemplate = `/** THIS FILE IS AUTOGENERATED AND SHOULD NOT EDITED */
+/** THIS FILE IS AUTOGENERATED AND SHOULD NOT BE EDITED */
+/* eslint-disable */
```
mkilpatrick, line 2: Why is this needed?

packages/visual-editor/src/vite-plugin/getTemplates.ts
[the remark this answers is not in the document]
mkilpatrick: It wasn't, it was inlined after some code and not the first line. Doesn't matter now though.

starter/vite.config.js
mkilpatrick: 😢

packages/visual-editor/src/vite-plugin/plugin.ts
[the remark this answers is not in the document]
mkilpatrick: Maybe something is off? I wouldn't expect that to be the case and the robots tell me
> If you’re needing to check for "build" in configureServer, it’s probably a sign that you’re trying to handle build-specific logic in the wrong hook. Instead, move that logic to hooks like buildStart, transform, or writeBundle, which are designed for the build process. For example:
buildStart: Runs at the start of the build process.
transform: Runs for each module during build.
writeBundle: Runs after the build is complete.

packages/visual-editor/src/vite-plugin/templates/main.tsx
asanehisa: Ah our eslint rn set for "*.{ts,js,tsx,jsx}". that file was failing eslint preventing commit (was failing due to expecting the import paths to end in .ts, but in the actual starter repo we wouldn't want that).

On second thought though, I think I'll change our eslint command to ignore those files instead.

packages/visual-editor/src/vite-plugin/plugin.ts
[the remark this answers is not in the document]
asanehisa: Hmmm ok, will try again + investigate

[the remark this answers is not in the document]
asanehisa: So buildStart, buildEnd, closeBundle all run on serve and build as expected https://vite.dev/guide/api-plugin#universal-hooks

Unfortunately buildEnd and closeBundle are being called before the server is actually closed (when running like pages dev). I believe this is b/c it thinks the server is closed when it is not (possibly due to our custom pagesJS code that starts up server using express?).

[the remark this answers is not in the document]
mkilpatrick: Fair enough

starter/vite.config.js
mkilpatrick: If this is going to cause a lot of strife to get right let's not worry about it. I know that getting it to resolve without the subpath when I was playing around was the biggest issue I had.

mkilpatrick (approved):

PR #254, opened 2025-03-06
feat: remove left toggle for theme manager

colton-demetriou:
Tested with pnpm run dev and verified left toggle/side bar is removed


jwartofsky-yext (approved):
lgtm

PR #256, opened 2025-03-06
feat: convert select fields to autocomplete combobox

briantstephan:
This adds a new `BasicSelector` field that replaces Puck's select fields with autocomplete comboboxes.

Screenshot 2025-03-06 at 6 46 27 PM


packages/visual-editor/src/components/editor/FontSizeSelector.tsx
```diff
@@ -99,25 +100,27 @@ export const FontSizeSelector = (
...
       const tailwindConfig: TailwindConfig = useTailwindConfig();
+      const options = convertDefaultFontSizesToOptions(
+        [
+          { label: "Default", value: "default", px: "" },
+          ...fontSizeOptions(includeLargeSizes),
+        ],
+        tailwindConfig
+      );
+
+      console.log(value);
```
asanehisa, line 111: remove console.log

packages/visual-editor/src/components/puck/Card.tsx
```diff
@@ -91,38 +88,22 @@ const cardFields: Fields<CardProps> = {
...
-      },
+      color: BasicSelector("Font Color", [
+        { label: "Default", value: "default" },
+        { label: "Primary", value: "primary" },
+        { label: "Secondary", value: "secondary" },
+        { label: "Accent", value: "accent" },
+        { label: "Text", value: "text" },
+        { label: "Background", value: "background" },
+      ]),
+      weight: BasicSelector("Font Weight", []),
```
asanehisa, line 99: Oh why is weight here with no options?

packages/visual-editor/src/components/puck/HoursStatus.tsx
```diff
@@ -34,14 +35,10 @@ const hoursStatusWrapperFields: Fields<HoursStatusProps> = {
...
   },
-  timeFormat: {
-    type: "select",
-    label: "Time Format",
-    options: [
-      { label: "12-hour", value: "12h" },
-      { label: "24-hour", value: "24h" },
-    ],
-  },
+  timeFormat: BasicSelector("Time Format", [
```
asanehisa, line 38: not u but i wonder why this is a selector rather than radio if only two options

packages/visual-editor/src/components/puck/Card.tsx
briantstephan: It gets overwritten with the correct options later in `resolveFields`, so it's not strictly necessary that it gets updated to use BasicSelector here but might as well keep it consistent.

packages/visual-editor/src/components/editor/FontSizeSelector.tsx
briantstephan: Done

packages/visual-editor/src/components/puck/HoursStatus.tsx
briantstephan: Radio makes more sense, updated

packages/visual-editor/src/components/editor/BasicSelector.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { Field, FieldLabel } from "@measured/puck";
+import { ChevronDown } from "lucide-react";
+import { Combobox } from "../../internal/puck/ui/Combobox.tsx";
+
+export const BasicSelector = (
```
benlife5, line 6: Can you add this to a index.ts so that it's exported from the package? And then in the components update the import so that it's imported from `"../../index.js"`, which gets replaced by the package name when the components are pulled via CLI

benlife5:
🔥

packages/visual-editor/src/components/editor/BasicSelector.tsx
briantstephan: Good call, done

packages/visual-editor/src/components/puck/Flex.tsx
```diff
@@ -8,6 +8,7 @@ import {
   layoutFields,
   layoutVariants,
 } from "./Layout.tsx";
+import { BasicSelector } from "../editor/index.js";
```
benlife5, line 11: One more to change

packages/visual-editor/src/components/puck/atoms/cta.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from "react";
 import { Link, LinkType } from "@yext/pages-components";
 import { Button, ButtonProps } from "./button.js";
+import { BasicSelector } from "../../editor/index.js";
```
benlife5, line 4: This can be from "../../../index.js"

PR #255, opened 2025-03-06
feat: remove theme left panel, not needed after removing GTM

colton-demetriou:
Tested locally and verified that the panel does not appear.

Screenshot 2025-03-06 at 5 31 44 PM

PR #257, opened 2025-03-07
fix: visual editor fails to build

benlife5:
dispatch was merged twice into ThemeHeader

the themeResolver test started failing in https://github.com/yext/visual-editor/pull/255, I'm not sure why it hadn't failed previously but ultimately console spy needed to be reset for each test. I also moved some of the test data into another file because internalThemeResolver.test.ts imported themeResolver.test.ts which was causing the themeResolver tests to be run twice


jwartofsky-yext (approved):
lgtm

asanehisa (approved):

PR #264, opened 2025-03-18
feat: add FilterSearch Bar for Locator component

Fondryext:
This PR inserts a basic FilterSearch bar into the Locator component. It does not do anything with the results. 
J=WAT-4698
TEST=manual

Ran locally and saw filter searches execute

### DIFF
--- a/starter/src/components/Locator.tsx
+++ b/starter/src/components/Locator.tsx
@@ -10,17 +10,12 @@ import {
 import * as React from "react";
 
 type LocatorProps = {
-  text: string;
   apiKey: string;
 };
 
 const fields: Fields<LocatorProps> = {
-  text: {
-    label: "Text",
-    type: "text",
-  },
   apiKey: {
-    label: "Api Key",
+    label: "API Key",
     type: "text",
   },
 };
@@ -28,7 +23,6 @@ const fields: Fields<LocatorProps> = {
 const LocatorComponent: ComponentConfig<LocatorProps> = {
   fields,
   defaultProps: {
-    text: "Some Placeholder",
     apiKey: "Fake Key",
   },
   label: "Locator",
--- a/starter/src/components/Locator.tsx
+++ b/starter/src/components/Locator.tsx
@@ -8,6 +8,7 @@ import {
   SearchHeadlessProvider,
 } from "@yext/search-headless-react";
 import * as React from "react";
+import { useDocument } from "@yext/visual-editor";
 
 type LocatorProps = {
   apiKey: string;
@@ -29,13 +30,16 @@ const LocatorComponent: ComponentConfig<LocatorProps> = {
   render: (props) => <Locator {...props} />,
 };
 
-const Locator: React.FC<LocatorProps> = (props) => {
+const Locator: React.FC<LocatorProps> = ({ apiKey }) => {
+  const document: {
+    businessId: number;
+  } = useDocument();
   const config = {
-    apiKey: props.apiKey,
+    apiKey,
     experienceKey: "jacob-test",
     locale: "en",
     experienceVersion: "STAGING",
-    businessId: 70452,
+    businessId: document.businessId,
     cloudRegion: CloudRegion.US,
     cloudChoice: CloudChoice.GLOBAL_MULTI,
     environment: Environment.PROD,
--- a/starter/src/components/Locator.tsx
+++ b/starter/src/components/Locator.tsx
@@ -1,12 +1,26 @@
 import { ComponentConfig, Fields } from "@measured/puck";
+import { FilterSearch } from "@yext/search-ui-react";
+import {
+  CloudChoice,
+  CloudRegion,
+  Environment,
+  provideHeadless,
+  SearchHeadlessProvider,
+} from "@yext/search-headless-react";
+import * as React from "react";
 
 type LocatorProps = {
   text: string;
+  apiKey: string;
 };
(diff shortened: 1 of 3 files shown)


starter/src/components/Locator.tsx
```diff
@@ -1,27 +1,62 @@
...
+  CloudRegion,
+  Environment,
+  provideHeadless,
+  SearchHeadlessProvider,
+} from "@yext/search-headless-react";
+import * as React from "react";
 
 type LocatorProps = {
   text: string;
+  apiKey: string;
```
Fondryext, line 14: I made apiKey a prop here at least partially to not check in a real secret. As far as all the others, like businessId, experience, vertical, field, entity, cloudregion, etc, I'm not 100% sure if these make sense best as props or if they should be like pulled from the environment?

mkilpatrick: When the pageset is created we will need to automatically generate the apiKey and add it as an environment variable. You can get the envVar off of the document at `document._env`.

starter/src/components/Locator.tsx
```diff
@@ -1,27 +1,62 @@
...
 };
 
 const fields: Fields<LocatorProps> = {
   text: {
     label: "Text",
     type: "text",
   },
+  apiKey: {
+    label: "Api Key",
```
k-gerner, line 23: nit: Can we make this `API Key` (all caps)

starter/src/components/Locator.tsx
vijay267: For now, we're still figuring out the backend flow to create the page set with a locator with Spruce, so our idea was to hardcode things to prototype the locator component and then handle the environment variables later along with the backend work.

starter/src/components/Locator.tsx
```diff
@@ -1,27 +1,62 @@
...
+    type: "text",
+  },
 };
 
 const LocatorComponent: ComponentConfig<LocatorProps> = {
   fields,
   defaultProps: {
     text: "Some Placeholder",
```
k-gerner, line 31: Feel free to remove the `text` field since we aren't using it anymore

k-gerner (approved):

starter/src/components/Locator.tsx
Fondryext: Gotcha. To be clear, are you all saying that for now I should e.x. attempt to read document._env for the apiKey? Or to keep what I have here until the backend flow is done?

starter/src/components/Locator.tsx
Fondryext: Done

starter/src/components/Locator.tsx
```diff
@@ -1,27 +1,56 @@
...
+const Locator: React.FC<LocatorProps> = (props) => {
+  const config = {
+    apiKey: props.apiKey,
+    experienceKey: "jacob-test",
+    locale: "en",
+    experienceVersion: "STAGING",
+    businessId: 70452,
+    cloudRegion: CloudRegion.US,
+    cloudChoice: CloudChoice.GLOBAL_MULTI,
+    environment: Environment.PROD,
```
asanehisa, line 41: did we want these to all be hardcoded?

asanehisa (approved):

starter/src/components/Locator.tsx
Fondryext: Related to the other comment thread, I think

starter/src/components/Locator.tsx
mkilpatrick: Idk if that question was for me or not, but you can keep it as a prop for now until the backend side is hooked up. That way you can manually enter your key for testing.

starter/src/components/Locator.tsx
```diff
@@ -1,27 +1,56 @@
...
-    text: "Some Placeholder",
+    apiKey: "Fake Key",
   },
   label: "Locator",
   render: (props) => <Locator {...props} />,
 };
 
-const Locator: React.FC<LocatorProps> = ({ text }) => {
-  return <div>{text}</div>;
+const Locator: React.FC<LocatorProps> = (props) => {
```
mkilpatrick, line 32: Destructure the props, either:

inline
`const Locator: React.FC<LocatorProps> = ({apiKey}) => {`

or

```
const Locator: React.FC<LocatorProps> = (props) => {
  const { apiKey, other...} = props;
```

starter/src/components/Locator.tsx
```diff
@@ -1,27 +1,56 @@
...
+  const config = {
+    apiKey: props.apiKey,
+    experienceKey: "jacob-test",
+    locale: "en",
+    experienceVersion: "STAGING",
+    businessId: 70452,
```
mkilpatrick, line 38: This is available on the document at `document.businessId`.

Your component can use the useDocument() hook.

starter/src/components/Locator.tsx
vijay267: I imagine Jacob's question was for me, but like Duval said, let's keep it a prop for now until we have the backend set up.

starter/src/components/Locator.tsx
Fondryext: Thanks!

starter/src/components/Locator.tsx
Fondryext: Thanks all.

PR #266, opened 2025-03-18
fix: add checks for breadcrumbs

asanehisa:
In the case of a stream with data like below, breadcrumbs won't break nor show up at all. This adds a check that document data for dm_directoryParents_example is an array of objects

```
dm_directoryParents_example: [
  345678,
  234533,
]
```


packages/visual-editor/src/components/puck/Breadcrumbs.tsx
```diff
@@ -6,11 +6,16 @@ export type BreadcrumbsProps = {
...
+// getDirectoryParents returns an array of objects. If no dm_directoryParents or children of
+// the directory parent are not objects, returns an empty array.
 const getDirectoryParents = (
   document: Record<string, any>
 ): Array<{ slug: string; name: string }> => {
   for (const key in document) {
-    if (key.startsWith("dm_directoryParents_")) {
+    if (
+      key.startsWith("dm_directoryParents_") &&
+      typeof document[key]?.[0] === "object"
```
mkilpatrick, line 17: We may as well verify that the object shape is what the component expects:
```
{
  name: string,
  slug: string,
}
```
as a list

mkilpatrick (approved):

PR #271, opened 2025-03-20
feat: update body text component

colton-demetriou:
Tested locally, will update font size values after theme updates.
Screenshot 2025-03-20 at 11 23 44 AM

### DIFF
--- a/packages/visual-editor/src/components/puck/BodyText.tsx
+++ b/packages/visual-editor/src/components/puck/BodyText.tsx
@@ -18,7 +18,6 @@ const BodyText = React.forwardRef<HTMLParagraphElement, BodyTextProps>(
   ({ text, variant }, ref) => {
     const document = useDocument();
 
-    // Map variant to fontSize
     const variantToFontSize: Record<
       BodyTextProps["variant"],
       NonNullable<BodyProps["fontSize"]>
--- a/packages/visual-editor/src/components/puck/BodyText.tsx
+++ b/packages/visual-editor/src/components/puck/BodyText.tsx
@@ -9,24 +9,13 @@ import {
   YextEntityFieldSelector,
 } from "../../index.js";
 
-interface BodyTextProps {
+interface BodyTextProps extends BodyProps {
   text: YextEntityField<string>;
-  variant: "small" | "base" | "large";
 }
 
 const BodyText = React.forwardRef<HTMLParagraphElement, BodyTextProps>(
-  ({ text, variant }, ref) => {
+  ({ text, ...bodyProps }, ref) => {
     const document = useDocument();
-
-    const variantToFontSize: Record<
-      BodyTextProps["variant"],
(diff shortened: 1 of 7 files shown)


briantstephan (approved):

packages/visual-editor/src/components/puck/BodyText.tsx
```diff
@@ -7,26 +7,34 @@ import {
...
+    const fontSize = variantToFontSize[variant];
+
     return (
       <EntityField
         displayName="Body"
         fieldId={text.field}
         constantValueEnabled={text.constantValueEnabled}
       >
-        <Body ref={ref} {...bodyProps}>
+        <Body ref={ref} fontSize={fontSize}>
```
benlife5, line 37: Maybe check with Aaron, but I my understanding of the figma these size options should be the only ones on the body.tsx atom (basically all "body" type text will be one of three sizes and one of two weights) and then you'd largely just be passing the props down through this component.

benlife5: The CSS class that will set the size doesn't exist but I think you can go ahead and update bodyVariants in body.tsx to only have the three options

packages/visual-editor/src/components/puck/Card.tsx
```diff
@@ -361,7 +361,7 @@ export const CardComponent: ComponentConfig<CardProps> = {
         constantValue: "Body",
         constantValueEnabled: true,
       },
-      fontSize: "default",
+      fontSize: "base",
       weight: "default",
```
asanehisa, line 365: should weight, color, transform be removed?

asanehisa: Well I guess question is, do we want the usage of <Body/> in other components to match change to <BodyText/>

asanehisa: ah is theme for BodyText changing to match as well? If so, can we update the theme.config.ts in the starter/

benlife5: I think we do want all body text instances to match the new simplified options, but the Card component fields can be updated in it's own PR (I think this just needed to change to satisfy ts)

packages/visual-editor/src/utils/applyAnalytics.ts
```diff
@@ -3,8 +3,9 @@ export const applyAnalytics = (document: Record<string, any>) => {
     return;
   }
 
-  const googleTagManagerId: string = JSON.parse(document.__.visualEditorConfig)
-    ?.siteAttributes?.googleTagManagerId;
+  const googleTagManagerId: string = JSON.parse(
+    document.__.visualEditorConfig
```
mkilpatrick, line 7: This shouldn't be part of this PR

benlife5 (approved):

PR #275, opened 2025-03-20
feat: emails component updates

briantstephan:
This removes the styling fields/props from the Emails component and fixes an issue with empty constant text fields and links.

Screenshot 2025-03-20 at 2 04 53 PM

### DIFF
--- a/packages/visual-editor/src/components/puck/Emails.tsx
+++ b/packages/visual-editor/src/components/puck/Emails.tsx
@@ -15,20 +15,10 @@ import { Link } from "@yext/pages-components";
 const emailsVariants = cva("components list-inside font-body-fontFamily", {
   variants: {
     includeHyperlink: {
-      true: "underline hover:no-underline",
-      false: "",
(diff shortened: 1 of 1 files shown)


packages/visual-editor/src/components/puck/Emails.tsx
```diff
@@ -9,76 +9,17 @@ import {
...
-      accent: "text-palette-accent",
-      text: "text-palette-text",
-      background: "text-palette-background",
-    },
     includeHyperlink: {
       true: "underline hover:no-underline",
       false: "",
     },
   },
-  compoundVariants: [
```
benlife5, line 56: I think we may still want the the compound variants to switch between the body and link styles.
If we don't want to do that switch, the fontSize className applied somewhere else in this component

briantstephan: Ah I see, updated

benlife5 (approved):

PR #277, opened 2025-03-21
feat: add basic result cards

Fondryext:
This PR adds basic result cards, still in need of styling. It also
adjusts how the SearchHeadlessProvider is created, so that the
call to provideHeadless is not inside of a component. This may need
adjusting.

J=WAT-4699
TEST=manual

Results are populated, albeit unstyled. No rerendering issues

### DIFF
--- a/starter/src/components/Locator.tsx
+++ b/starter/src/components/Locator.tsx
@@ -68,7 +68,7 @@ const Locator: React.FC<LocatorProps> = (props) => {
         kind: "fieldValue",
         fieldId: params.newFilter.fieldId,
         value: params.newFilter.value,
-        matcher: Matcher.Equals,
+        matcher: Matcher.Near,
       },
     };
     searchActions.setStaticFilters([locationFilter]);
--- a/starter/src/components/Locator.tsx
+++ b/starter/src/components/Locator.tsx
@@ -1,22 +1,30 @@
 import { ComponentConfig, Fields } from "@measured/puck";
-import { FilterSearch } from "@yext/search-ui-react";
 import {
-  CloudChoice,
-  CloudRegion,
-  Environment,
-  provideHeadless,
-  SearchHeadlessProvider,
+  FilterSearch,
+  OnSelectParams,
+  StandardCard,
+  VerticalResults,
+} from "@yext/search-ui-react";
+import {
+  Matcher,
+  SelectableStaticFilter,
+  useSearchActions,
+  useSearchState,
 } from "@yext/search-headless-react";
 import * as React from "react";
-import { useDocument } from "@yext/visual-editor";
 
 type LocatorProps = {
-  apiKey: string;
+  fieldApiName: string;
+  entityType: string;
 };
 
 const fields: Fields<LocatorProps> = {
-  apiKey: {
-    label: "API Key",
+  fieldApiName: {
+    label: "Field Name",
+    type: "text",
+  },
+  entityType: {
+    label: "Entity Type",
     type: "text",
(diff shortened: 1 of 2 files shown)


starter/src/components/Locator.tsx
```diff
@@ -1,59 +1,82 @@
...
 import * as React from "react";
-import { useDocument } from "@yext/visual-editor";
 
 type LocatorProps = {
-  apiKey: string;
+  fieldApiName: string;
+  entityType: string;
 };
 
 const fields: Fields<LocatorProps> = {
```
benlife5, line 21: I think you can use `YextEntityFieldSelector` here to auto generate a list of fields and then use `fieldApiName.entityField` in the component to get what was selected. We might be able to build something like that for entity type too.
Otherwise, this LGTM

Fondryext: We've decided for now to not allow users to choose a field

starter/src/components/Locator.tsx
```diff
@@ -1,59 +1,104 @@
...
+    },
+  },
+  defaultVariants: {
+    backgroundColor: "default",
   },
+});
+
+type LocatorProps = VariantProps<typeof locatorVariants>;
+
+const locatorFields: Fields<LocatorProps> = {
```
Fondryext, line 39: mostly just adjusted this in the style of other components so that there would be *something* in the props, even if these will definitely be adjusted in the followup changes.

starter/src/components/Locator.tsx
```diff
@@ -1,59 +1,104 @@
...
+  const searchActions = useSearchActions();
+  const handleFilterSelect = (params: OnSelectParams) => {
+    const locationFilter: SelectableStaticFilter = {
+      displayName: params.newDisplayName,
+      selected: true,
+      filter: {
+        kind: "fieldValue",
+        fieldId: params.newFilter.fieldId,
+        value: params.newFilter.value,
+        matcher: Matcher.Equals,
```
k-gerner, line 71: Should this be using the `$near` filter rather than `$eq`?

Fondryext: Yeah probably.

k-gerner:
Looks good to me, but will wait on Sumo for final approval

starter/src/components/Locator.tsx
```diff
@@ -1,59 +1,104 @@
...
+const DEFAULT_FIELD = "builtin.location";
+const DEFAULT_ENTITY_TYPE = "location";
 
-const fields: Fields<LocatorProps> = {
-  apiKey: {
-    label: "API Key",
-    type: "text",
+const locatorVariants = cva("", {
+  variants: {
+    backgroundColor: {
```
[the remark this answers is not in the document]
benlife5: Depends on how you want to reconcile the `dev/watson-locator` branch with the `spring-2025-component-updates` branch -- my change should be merged into the component updates branch soon but it won't necessarily be available here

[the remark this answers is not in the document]
Fondryext: I'm willing to do whatever you all prefer - I don't mind if keeping this as-is means a little extra lift when merging into the `spring-2025-component-updates` branch - as I understand it the other option is waiting for your change and then copying it over here for the time being?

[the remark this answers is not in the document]
benlife5: SGTM, I think it'll probably be easier to make the styling changes all at once rather than piecemeal

benlife5 (approved):

PR #282, opened 2025-03-24
refactor: new theme config

benlife5:
Aligns the colors, headings, body and page sections of the theme config with the document. Also updates of the heading variables.

Usages of colors and body styles will need to be fixed in a future PR following the addition of auto-gen colors and body styles (I will work on that next). The section component (and Flex and Grid) will need some updates as well, since we no longer set gap and have different padding types.

### DIFF
--- a/starter/theme.config.ts
+++ b/starter/theme.config.ts
@@ -34,7 +34,7 @@ const fontWeightOptions = (fontVariable?: string) => {
 
 export const themeConfig: ThemeConfig = {
   palette: {
-    label: "Color Palette",
+    label: "Colors",
     styles: {
       primary: {
         label: "Primary",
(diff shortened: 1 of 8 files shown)


starter/theme.config.ts
```diff
@@ -48,62 +48,56 @@ export const themeConfig: ThemeConfig = {
         default: "#FFFFFF",
```
asanehisa, line 48: do we wanna change the default colors to match the mock? with #CF0A2C, #737B82, #FF7E7E, #000000

asanehisa (commented):
Header and Footer has the background color set to         
default: "var(--colors-palette-background)",

mkilpatrick (approved):

starter/theme.config.ts
benlife5: Good call

mkilpatrick (approved):

PR #284, opened 2025-03-24
feat: autogenerated theme colors

benlife5:
- Creates the tailwind colors based on the theme editor colors
  - palette-primary-light
  - palette-secondary-light
  - palette-tertiary-light
  - palette-quaternary-light
  - palette-primary-dark
  - palette-secondary-dark 
- Overrides tailwind's grays with the 9 grays from the Figma 
- Creates a helper to provide the 7 background options
- Updates header to always use the white background and footer to have the 7 background options

Screenshot 2025-03-24 at 5 02 24 PM

### DIFF
--- a/packages/visual-editor/src/components/puck/Footer.tsx
+++ b/packages/visual-editor/src/components/puck/Footer.tsx
@@ -18,8 +18,8 @@ import {
   FaYoutube,
 } from "react-icons/fa";
 import {
-  BackgroundColorOption,
-  getBackgroundColorOptions,
+  type BackgroundStyle,
+  backgroundColors,
 } from "../../utils/themeConfigOptions.ts";
 
 type socialLink = {
@@ -30,20 +30,20 @@ type socialLink = {
 };
 
 type FooterProps = {
-  backgroundColor?: BackgroundColorOption;
+  backgroundColor?: BackgroundStyle;
 };
 
 const footerFields: Fields<FooterProps> = {
   backgroundColor: BasicSelector(
     "Background Color",
-    getBackgroundColorOptions()
+    Object.values(backgroundColors)
   ),
 };
 
 const Footer: ComponentConfig<FooterProps> = {
   fields: footerFields,
   defaultProps: {
-    backgroundColor: getBackgroundColorOptions()[0].value,
+    backgroundColor: backgroundColors.background1.value,
   },
   label: "Footer",
   render: (props) => <FooterComponent {...props} />,
--- a/packages/visual-editor/src/components/puck/Footer.tsx
+++ b/packages/visual-editor/src/components/puck/Footer.tsx
@@ -8,7 +8,6 @@ import {
   useDocument,
   BasicSelector,
 } from "../../index.ts";
-import { cva, VariantProps } from "class-variance-authority";
 import {
   FaFacebook,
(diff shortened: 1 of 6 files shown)


packages/visual-editor/src/components/puck/Footer.tsx
```diff
@@ -26,37 +29,22 @@ type socialLink = {
...
+  backgroundColor: BasicSelector(
+    "Background Color",
+    getBackgroundColorOptions()
+  ),
 };
 
 const Footer: ComponentConfig<FooterProps> = {
   fields: footerFields,
+  defaultProps: {
+    backgroundColor: getBackgroundColorOptions()[0].value,
```
mkilpatrick, line 46: Can these be keyed instead of arrayed?

packages/visual-editor/src/components/puck/Header.tsx
```diff
@@ -15,23 +14,7 @@ const PLACEHOLDER_IMAGE: ComplexImageType = {
...
-      background: "bg-palette-background",
-    },
-  },
-  defaultVariants: {
-    backgroundColor: "default",
-  },
-});
-
-export type HeaderProps = VariantProps<typeof headerVariants>;
+export type HeaderProps = object;
```
mkilpatrick, line 17: Is this so you don't get TS errors because there are no props?

starter/tailwind.config.ts
```diff
@@ -8,7 +8,35 @@ export default {
     "./node_modules/@yext/visual-editor/dist/**/*.js",
   ],
   theme: {
-    extend: themeResolver({}, themeConfig),
+    extend: themeResolver(
```
mkilpatrick, line 11: Can these be overwritten in themeConfig at all? It'd be unfortunate for the starter to have to define this in case we need to change anything.

benlife5: We could inject these via `themeResolver` -- I think from a pure perspective  `themeResolver` should just be merging the `theme.config` with the `tailwind.config` and all styles should be defined within those. That being said, I get we're mostly going for the no-dev case in which it's best to have as much as possible in the library.

starter/tailwind.config.ts
```diff
@@ -8,7 +8,35 @@ export default {
...
+            "hsl(from var(--colors-palette-primary) h s 98)",
+          "palette-secondary-light":
+            "hsl(from var(--colors-palette-secondary) h s 98)",
+          "palette-tertiary-light":
+            "hsl(from var(--colors-palette-tertiary) h s 98)",
+          "palette-quaternary-light":
+            "hsl(from var(--colors-palette-quaternary) h s 98)",
+          "palette-primary-dark":
+            "hsl(from var(--colors-palette-primary) h s 20)",
+          "palette-secondary-dark":
```
asanehisa, line 24: just wondering, did we want a tertiary and quaternary for dark?

benlife5: Idk the rationale but there's just two dark styles

starter/tailwind.config.ts
mkilpatrick: Could we export another function or const called something like `defaultThemeVariables` or something and use it here?

mkilpatrick (approved):

PR #288, opened 2025-03-25
feat: adjust theme for header and footer

asanehisa:
This is alpha feedback so I have it merging into main. Let me know if this should go into components branch instead. Confirmed new themes affect Header and Footer correctly. After VE is cut, the theme.config.ts in the starter will need the same adjustments made in this visual-editor/starter/theme.config.ts file. 

Screenshot 2025-03-25 at 3 35 14 PM


benlife5:
LGTM -- it will be overwritten by the spring component updates so idk if we want to merge or not

mkilpatrick:
> LGTM -- it will be overwritten by the spring component updates so idk if we want to merge or not

Hmm yea, I didn't think about this conflicting with the 2025 component changes. Seems pointless to do this as it's just going to change.

PR #293, opened 2025-03-26
feat: add additional result card data, and basic styling

Fondryext:
This PR adds a little more styling to the result cards.
It also builds out a variety of information, including:
- name
- address
- basic hours
- distance
- phone number
- slug link
- map link

Some changes will follow - the hours display will use a
forthcoming display util to be added to search-ui-react, and
a later change will pass down styling from the VLE in general
J=WAT-4699
TEST=manual

Screenshot 2025-03-27 at 4 34 25 PM

### DIFF
--- a/starter/src/components/Locator.tsx
+++ b/starter/src/components/Locator.tsx
@@ -18,7 +18,7 @@ import { cva, VariantProps } from "class-variance-authority";
 import { BasicSelector, Button, themeManagerCn } from "@yext/visual-editor";
 import { normalizeSlug } from "@yext/visual-editor";
 import { useEffect, useState } from "react";
-import { HoursStatus } from "@yext/pages-components";
+import { HoursStatus, Link } from "@yext/pages-components";
 import { Address, Hours } from "../types/autogen.ts";
 
 const DEFAULT_FIELD = "builtin.location";
@@ -155,17 +155,17 @@ const LocationCard: CardComponent<Location> = ({
               hours={location.hours}
               timezone={location.timezone}
               className="text-sm"
-            ></HoursStatus>
+            />
           )}
           <div>
-            <a
+            <Link
               target={"_blank"}
               href={getPath(location)}
               className="text-sm py-1000 text-blue-700 cursor-pointer"
               rel="noreferrer"
             >
               <input type="submit" value="View More Information" />
-            </a>
+            </Link>
           </div>
           <div>
             {location.yextDisplayCoordinate && (
--- a/starter/src/components/Locator.tsx
+++ b/starter/src/components/Locator.tsx
@@ -12,34 +12,49 @@ import {
   useSearchState,
 } from "@yext/search-headless-react";
 import * as React from "react";
+import { cva, VariantProps } from "class-variance-authority";
+import { BasicSelector, themeManagerCn } from "@yext/visual-editor";
 
-type LocatorProps = {
-  fieldApiName: string;
-  entityType: string;
-};
+const DEFAULT_FIELD = "builtin.location";
+const DEFAULT_ENTITY_TYPE = "location";
 
-const fields: Fields<LocatorProps> = {
-  fieldApiName: {
-    label: "Field Name",
-    type: "text",
+const locatorVariants = cva("", {
+  variants: {
+    backgroundColor: {
+      default: "bg-locator-backgroundColor",
+      primary: "bg-palette-primary",
+      secondary: "bg-palette-secondary",
+      accent: "bg-palette-accent",
+      text: "bg-palette-text",
+      background: "bg-palette-background",
+    },
   },
-  entityType: {
-    label: "Entity Type",
-    type: "text",
+  defaultVariants: {
+    backgroundColor: "default",
   },
+});
+
+type LocatorProps = VariantProps<typeof locatorVariants>;
+
+const locatorFields: Fields<LocatorProps> = {
+  backgroundColor: BasicSelector("Background Color", [
+    { label: "Default", value: "default" },
+    { label: "Primary", value: "primary" },
+    { label: "Secondary", value: "secondary" },
+    { label: "Accent", value: "accent" },
+    { label: "Text", value: "text" },
+    { label: "Background", value: "background" },
+  ]),
 };
 
 const LocatorComponent: ComponentConfig<LocatorProps> = {
-  fields,
-  defaultProps: {
-    fieldApiName: "builtin.location",
-    entityType: "location",
-  },
+  fields: locatorFields,
   label: "Locator",
   render: (props) => <Locator {...props} />,
 };
 
-const Locator: React.FC<LocatorProps> = ({ fieldApiName, entityType }) => {
+const Locator: React.FC<LocatorProps> = (props) => {
+  const { backgroundColor } = props;
   const resultCount = useSearchState(
     (state) => state.vertical.resultsCount || 0,
   );
@@ -53,7 +68,7 @@ const Locator: React.FC<LocatorProps> = ({ fieldApiName, entityType }) => {
         kind: "fieldValue",
         fieldId: params.newFilter.fieldId,
         value: params.newFilter.value,
-        matcher: Matcher.Equals,
+        matcher: Matcher.Near,
       },
     };
     searchActions.setStaticFilters([locationFilter]);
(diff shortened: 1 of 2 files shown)


starter/src/components/Locator.tsx
```diff
@@ -102,4 +116,207 @@ const Locator: React.FC<LocatorProps> = (props) => {
...
+
+  const distanceInMiles = distance
+    ? (distance / 1609.344).toFixed(1)
+    : undefined;
+
+  return (
+    <div className="flex justify left border-y p-50">
+      <div className="flex p-2.5">
+        <div>
+          <p className="text-lg py-2 text-rose-500">{location.name}</p>
```
asanehisa, line 138: Uncertain if you guys wanted to use the theme (users can edit theme in platform and our components values change as well). If so you'd wanna use text-body-fontSize instead of text-sm and instead of text-lg you'd likely use text-heading1-fontSize or something. Colors can also come from the theme.

starter/src/components/Locator.tsx
```diff
@@ -102,4 +116,207 @@ const Locator: React.FC<LocatorProps> = (props) => {
...
+              target={"_blank"}
+              href={`tel:${location.mainPhone}`}
+              className="text-sm py-5"
+              rel="noreferrer"
+            >
+              {formatPhoneNumber(location.mainPhone)}
+            </a>
+          )}
+          {location.hours && <p className="text-lg">Hours:</p>}
+          {location.hours && (
```
asanehisa, line 152: Oh we have an [hours component](https://github.com/yext/js/tree/main/packages/pages-components/src/components/hours) in js if you'd like to use that?

starter/src/components/Locator.tsx
bryanreed: We definitely do want to use the theme, if the guidance isn't defined in our mocks/requirements for what maps to what I think its ok to push this through and note that we need to gather that and do it on the next pass over

starter/src/components/Locator.tsx
```diff
@@ -102,4 +116,207 @@ const Locator: React.FC<LocatorProps> = (props) => {
...
+        <div className="flex items-start p-5 italic">
+          {distanceInMiles + " mi"}
+        </div>
+      )}
+    </div>
+  );
+};
+
+//reformats a phone number in the familiar (123)-456-7890 style
+function formatPhoneNumber(phoneNumber: string) {
```
asanehisa, line 235: Oh! We have formatPhoneNumber in /Users/asanehisa/visual-editor/packages/visual-editor/src/components/puck/Phone.tsx if thats helpful

starter/src/components/Locator.tsx
```diff
@@ -102,4 +116,207 @@ const Locator: React.FC<LocatorProps> = (props) => {
...
+  // right language populated here.
+  const localePath = "en";
+  const path = location.address
+    ? `${localePath}${location.address.region}/${location.address.city}/${location.address.line1}`
+    : `${localePath}${location.id}`;
+
+  return normalizeSlug(path);
+};
+
+export interface Interval {
```
asanehisa, line 270: I think most of these types are exported from JS as well

starter/src/components/Locator.tsx
vijay267: We have the work to use the theme / styling as a followup item. Our initial goal was just to get something functional until then.

starter/src/components/Locator.tsx
Fondryext: oh god. I should've asked, I've been working on a search hours formatter for the last day 😓 . thank you!

starter/src/components/Locator.tsx
```diff
@@ -102,4 +116,207 @@ const Locator: React.FC<LocatorProps> = (props) => {
...
+              href={getPath(location)}
+              className="text-sm py-1000 text-blue-700 cursor-pointer"
+              rel="noreferrer"
+            >
+              <input type="submit" value="View More Information" />
+            </a>
+          </div>
+          <div>
+            {location.yextDisplayCoordinate && (
+              <button
```
asanehisa, line 203: We have a button atom in visual-editor/packages/visual-editor/src/components/puck/atoms which might be nice to use (then future formatting and matching theme will be easier).

starter/src/components/Locator.tsx
Fondryext: Oh cool. To make sure I understand, are you suggesting I just copy the same logic into this test component for the time being?

Fondryext: There's some import errors as this still exists solely in the starter repo, around importing libphonenumber-js. Maybe this will be an easier adjustment once this code moves out of starter?

starter/src/components/Locator.tsx
Fondryext: oh nice thanks

starter/src/components/Locator.tsx
Fondryext: Good catch - I forgot the search types really are just derived from streams/kg to begin with.

starter/src/components/Locator.tsx
```diff
@@ -102,4 +118,124 @@ const Locator: React.FC<LocatorProps> = (props) => {
...
+              target={"_blank"}
+              href={`tel:${location.mainPhone}`}
+              className="text-sm py-5"
+              rel="noreferrer"
+            >
+              {formatPhoneNumber(location.mainPhone)}
+            </a>
+          )}
+          {location.hours && (
+            <HoursStatus
```
asanehisa, line 154: nit can just be <HoursStatus />

starter/src/components/Locator.tsx
```diff
@@ -102,4 +118,124 @@ const Locator: React.FC<LocatorProps> = (props) => {
...
+              hours={location.hours}
+              timezone={location.timezone}
+              className="text-sm"
+            ></HoursStatus>
+          )}
+          <div>
+            <a
```
asanehisa, line 161: would we be able to use Link here ?

Fondryext: Sure! For the next little while you can safely assume Watson's failures to use pages-components is indicative of our ignorance that they exist 😅

starter/src/components/Locator.tsx
```diff
@@ -102,4 +118,124 @@ const Locator: React.FC<LocatorProps> = (props) => {
...
+
+  return (
+    <div className="flex justify left border-y p-50">
+      <div className="flex p-2.5">
+        <div>
+          <p className="text-lg py-2 text-rose-500">{location.name}</p>
+          <p className="text-sm">{location.address.line1}</p>
+          <p className="text-sm">{`${location.address.city}, ${location.address.region} ${location.address.postalCode}`}</p>
+          {location.mainPhone && (
+            <a
```
mkilpatrick, line 144: Use Link for any a tags. This will allow it to get analytics.

starter/src/components/Locator.tsx
mkilpatrick: I think we just want to export the function so you can use it instead of copying it, especially when this is in the same repo.

starter/src/components/Locator.tsx
```diff
@@ -102,4 +118,124 @@ const Locator: React.FC<LocatorProps> = (props) => {
...
+}
+
+const getPath = (location: Location) => {
+  if (location.slug) {
+    return location.slug;
+  }
+
+  // TODO: in the same way that locale is hardcoded to 'en' in dev.tsx we may need to eventually have the
+  // right language populated here.
+  const localePath = "en";
```
mkilpatrick, line 221: Check out our default in main.tsx. If it's en then we don't prepend the locale.

starter/src/components/Locator.tsx
Fondryext: Thanks for the context.

starter/src/components/Locator.tsx
Fondryext: Thanks for the catch.

starter/src/components/Locator.tsx
benlife5: @mkilpatrick once they finish this component and it's ready to be moved from the `starter` dir to the `packages/visual-editor` dir, they'll be able to do a local import of the same function. For now, I think this is best, unless we want to make the phone formatter globally exported from `@yext/visual-editor` 

(@Fondryext just fyi to do the global export, it has to be exported by one of the files linked in `packages/visual-editor/src/index.ts`)

mkilpatrick: I wonder if we should export it for people to use in their custom components.

Fondryext: LMK if you'd like me to do that export as part of a different change or anything. Otherwise I think this is the last outstanding comment chain, so I'll likely ship this based on Deepak's approval by EOD.

mkilpatrick: For the purposes of this PR you can keep it as is. We need to determine which things we want to expose to users (those creating custom components in a hybrid world). So if we go down that road we can refactor later.

PR #294, opened 2025-03-26
feat: update grid component

briantstephan:
This removes the Maximum Content Width option from the Grid component, and simplifies the behavior of columns. The column field is now a basic number field, and columns wrap to rows on small screen sizes if the Grid is at the top level of the layout, but are fixed to a horizontal alignment at all times if the Grid is at a lower level. This also renames the padding fields and shows the raw px values for spacing rather than the more complicated "[number] ([pixel_value]px)" format.

This screenshot shows two Grids containing CTA buttons on a small mobile screen. The top grid is at the top level of the layout, so it wraps into one column with multiple rows. The bottom grid is within another grid, so the buttons are always horizontally aligned into columns.

Screenshot 2025-03-26 at 11 51 35 AM

### DIFF
--- a/packages/visual-editor/src/components/editor/SpacingSelector.tsx
+++ b/packages/visual-editor/src/components/editor/SpacingSelector.tsx
@@ -71,7 +71,7 @@ const convertDefaultSpacingsToOptions = (
     );
     const spacingPx = customSpacing ? customSpacing.toString() : option.px;
     return {
-      label: option.label + (option.px ? ` (${spacingPx}px)` : ""),
+      label: option.px ? `${spacingPx}px` : option.label,
       value: option.value,
     };
   });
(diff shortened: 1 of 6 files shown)


benlife5:
I think I misread the document and created both topPadding and bottomPadding in the Page Section part of `theme.config` when they should have been the same thing (key:verticalPadding, label: "Top/Bottom Padding")

Would you mind fixing that and making sure the default option in there works with the sectionVariants?

Also in sectionVariants, I don't think horizontalPadding should have a default option any more since it's not in the theme config. I guess just set the defaultProps in the components that use it to 0?

benlife5 (approved):

colton-demetriou (approved):

PR #295, opened 2025-03-26
fix: export heading options

benlife5:
Moves `headingLevelOptions` to themeConfigOptions.tsx for export from the package
Also updates a number of imports that should have been importing through the package alias rather than specific files

### DIFF
--- a/packages/visual-editor/src/components/puck/Flex.tsx
+++ b/packages/visual-editor/src/components/puck/Flex.tsx
@@ -1,15 +1,17 @@
 import * as React from "react";
 import { ComponentConfig, DropZone, Fields } from "@measured/puck";
 import { Section } from "./atoms/section.js";
-import { themeManagerCn } from "../../utils/cn.js";
 import {
   innerLayoutVariants,
   layoutFields,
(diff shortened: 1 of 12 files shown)


packages/visual-editor/src/components/puck/Flex.tsx
```diff
@@ -1,15 +1,17 @@
...
   layoutProps,
   layoutVariants,
 } from "./Layout.tsx";
-import { BasicSelector } from "../../index.js";
-import { backgroundColors } from "../../utils/themeConfigOptions.ts";
+import {
+  BasicSelector,
+  backgroundColors,
+  themeManagerCn,
+} from "../../index.js";
```
mkilpatrick, line 14: These should come from the file rather than index.js, same everywhere else. Some of them acually use .ts instead of .js too.

mkilpatrick: 👍

asanehisa (approved):

jwartofsky-yext (approved):

PR #299, opened 2025-03-28
refactor: button/link/cta component updates

benlife5:
In the button/cta atoms, creates four variants: primary, secondary, link, and directoryLink based on the Figma (directoryLink isn't selectable by the user)

button.tsx includes code to get the background color of the cta that is being rendered and determine what color cta or link should be rendered

main-desktop
main-mobile
directory-card-desktop
directory-card-mobile
directory-list-desktop
directory-list-mobile

https://github.com/user-attachments/assets/6c7dcaeb-76d9-4b9b-be1e-b08ce175904e


packages/visual-editor/src/components/puck/CtaWrapper.tsx
```diff
@@ -23,15 +24,11 @@ const ctaWrapperFields: Fields<CTAWrapperProps> = {
...
-    label: "Variant",
-    type: "radio",
-    options: [
-      { label: "Primary", value: "primary" },
-      { label: "Outline", value: "secondary" },
-      { label: "Link", value: "link" },
-    ],
-  },
+  variant: BasicSelector("Variant", [
+    { label: "Primary", value: "primary" },
```
mkilpatrick, line 28: I wonder if for anything theme related with set options we should expose that as something a [custom component creator] user could import.

packages/visual-editor/src/components/puck/GetDirections.tsx
```diff
@@ -26,7 +26,8 @@ const getDirectionsFields: Fields<GetDirectionsProps> = {
     label: "Variant",
     type: "radio",
     options: [
-      { label: "Button", value: "primary" },
+      { label: "Primary", value: "primary" },
```
mkilpatrick, line 29: Case in point, this is repeated again. Maybe not for this PR but we audit and do this for all of these type of theme options in one swoop.

packages/visual-editor/src/components/puck/atoms/button.tsx
```diff
@@ -64,26 +68,70 @@ export interface ButtonProps
...
+    const handleRef = React.useCallback((node: HTMLButtonElement | null) => {
+      // If the ref is resolved to an HTML element, get the parent color
+      if (node) {
+        // Loop until parent has a background color or there are no more parents
+        let parent: HTMLElement | null = node;
+        while (
+          parent &&
+          (window.getComputedStyle(parent).backgroundColor ===
+            "rgba(0, 0, 0, 0)" || // skip transparent parents
+            parent.className.includes("_DropZone")) // skip Puck wrappers
```
mkilpatrick, line 85: Not sure if this is semi-dangerous if puck changes classnames or adds new ones.

packages/visual-editor/src/components/puck/atoms/cta.tsx
```diff
@@ -24,16 +25,47 @@ const linkTypeFields: Field = BasicSelector("Link Type", [
...
       <Link
         cta={{
           link: link ?? "",
           linkType: linkType,
         }}
+        eventName={eventName}
+        target={target}
       >
         {label}
+        <FaAngleRight
```
mkilpatrick, line 60: Should this just be guarded by caretVisibility? It's more ideal to not include things in the DOM if they'll never be shown.

packages/visual-editor/src/components/puck/atoms/maybeLink.tsx
```diff
@@ -1,19 +1,35 @@
...
-  if (props.href) {
+  const {
+    href,
+    children,
+    className,
+    eventName,
+    alwaysHideCaret,
+    variant = "link",
+  } = props;
+  if (href) {
```
mkilpatrick, line 22: empty line above

packages/visual-editor/src/utils/index.ts
```diff
@@ -23,3 +23,4 @@ export {
 } from "./themeConfigOptions.ts";
 export { applyAnalytics } from "./applyAnalytics.ts";
 export { getPageMetadata } from "./getPageMetadata.ts";
+export { srgbToHSL } from "./colors.ts";
```
mkilpatrick, line 26: Does this need to be exported for consumers?

packages/visual-editor/src/components/puck/CtaWrapper.tsx
benlife5: Yeah, probably should export this for consistency
I think we're probably due for a conversation with Aaron about what expected to be exported from the package vs. what you'd copy be expected to copy from example code/the CLI tool. It's come up a lot this week

packages/visual-editor/src/components/puck/atoms/button.tsx
benlife5: It is in the sense that Puck could change and break it, but that would be a little surprising to me unless it was another full dropzone overhaul
The problem is that these are the divs that create the blue-ish wrapper when you hover so we need some way to ignore them

packages/visual-editor/src/components/puck/atoms/cta.tsx
benlife5: Ideally, but the theme's caretVisibility is set on a CSS variable so it's not easy to access in JS

packages/visual-editor/src/utils/index.ts
benlife5: It's only use is in the button.tsx atom -- so maybe it makes more sense for it to not in the package and instead just at the bottom of that file?

mkilpatrick: Ok let's just put it there for now and if we feel it should be exported later we can do that.

packages/visual-editor/src/components/puck/Promo.tsx
```diff
@@ -109,11 +110,11 @@ const promoFields: Fields<PromoProps> = {
...
         },
       }),
-      variant: BasicSelector("Primary CTA Variant", [
-        { label: "Primary", value: "primary" },
-        { label: "Secondary", value: "secondary" },
-        { label: "Link", value: "link" },
-      ]),
+      variant: {
+        label: "Variant",
+        type: "radio",
```
mkilpatrick, line 115: Did @apav-dev want these as radios vs selectors? I think radio makes sense for 3 or less options but want to confirm.

benlife5: Confirmed that we want radio when there's 2-3 options

asanehisa (approved):
lgtm (but am trusting the math / color logic)

mkilpatrick (approved):

packages/visual-editor/src/components/puck/atoms/button.tsx
```diff
@@ -139,4 +139,60 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
 );
 Button.displayName = "Button";
 
+/**
+ * srgbToHSL converts srgb colors to hsl values.
+ * @param rgb [r, g, b]
+ * @returns {number[] | undefined} [h, s, l] if conversion succeeds
+ */
+export const srgbToHSL = (rgb: number[]): number[] | undefined => {
```
mkilpatrick, line 147: We don't want to lose the tests for this just because it was moved.

mkilpatrick (approved):

benlife5:
Recent updates:
- re-add test
- merge in other updates
- fix issue where sometimes getComputedStyle returns `rgb` rather than `color(srgb)`
- change CTA to use KG's link type, and only expose link type when using a constant value

apav-dev:
@benlife5 I actually only have 1 really nit-picky comment which is that I'm hoping to see the px value when choosing letter spacing for the buttons

asanehisa (approved):

PR #301, opened 2025-03-28
feat: flex component updates

briantstephan:
This replaces the direction, justify content, and wrap selectors to radios. The spacing options have already been updated to only show the px value in the options.

Screenshot 2025-03-28 at 1 49 04 PM

### DIFF
--- a/packages/visual-editor/src/components/puck/Flex.tsx
+++ b/packages/visual-editor/src/components/puck/Flex.tsx
@@ -71,19 +71,31 @@ const FlexContainer = React.forwardRef<HTMLDivElement, FlexProps>(
 FlexContainer.displayName = "Flex";
 
 const flexContainerFields: Fields<FlexProps> = {
-  direction: BasicSelector("Direction", [
-    { value: "row", label: "Horizontal" },
-    { value: "column", label: "Vertical" },
-  ]),
-  justifyContent: BasicSelector("Justify Content", [
-    { value: "start", label: "Start" },
-    { value: "center", label: "Center" },
-    { value: "end", label: "End" },
-  ]),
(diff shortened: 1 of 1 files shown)


benlife5:
This wasn't in the doc, but can you try removing the [wrapper from the dropZone](https://puckeditor.com/docs/integrating-puck/multi-column-layouts#removing-the-wrapper) as part of this update? 

For example if you have two CTAs in a Flex and go to mobile view, they don't expand unless you go into Interactive Mode (cmd+i). On the live page, they'll be full width but the wrapper is messing it up in the editor

https://github.com/user-attachments/assets/ed118c43-a425-4199-886c-51ab82022311

packages/visual-editor/src/components/puck/Flex.tsx
```diff
@@ -71,35 +71,47 @@ const FlexContainer = React.forwardRef<HTMLDivElement, FlexProps>(
...
+    direction: "row",
     justifyContent: "start",
     wrap: "nowrap",
     gap: "0",
     verticalPadding: "default",
     horizontalPadding: "0",
     backgroundColor: backgroundColors.background1.value,
   },
-  resolveFields: (data, params) => {
+  resolveFields: (data: { props: FlexProps }, params: { parent: any }) => {
```
benlife5, line 114: Are these explicit types necessary? I would expect it could get the real type implicitly from being a resolveFields value

briantstephan: Something must be configured wrong with my VS Code because it wasn't able to pick up the types. Removed since it turns out they aren't necessary.

benlife5 (approved):

PR #303, opened 2025-03-28
fix: update BasicSelector to correctly handle objects

briantstephan:
This updates BasicSelector to handle option stringification itself before sending the options to the Combobox. It now correctly compares object values and works as expected. Tested with the Background Color selector.


packages/visual-editor/src/components/editor/BasicSelector.tsx
```diff
@@ -3,10 +3,19 @@ import { Field, FieldLabel } from "@measured/puck";
...
-  label: string,
-  options: { label: string; value: any; color?: string }[]
-): Field => {
+type Option = {
+  label: string;
+  value: any;
+  color?: string;
+};
+
+type StringifiedOption = {
```
mkilpatrick, line 12: Why do we need two types that have the same keys/values?

benlife5 (approved):

jwartofsky-yext (approved):

PR #304, opened 2025-03-31
fix: adjust header

asanehisa:
Header theme config no longer exists so removed those tailwind usages. 

TODO in future: 
- Use Puck Permissions to prevent a user from dragging more than one Header component onto the page.


benlife5:
Ah sorry for the duplication again, #299 includes switching these over to CTAs with the link variant

PR #310, opened 2025-04-01
fix: miscellaneous component updates

benlife5:
Entity Field Tooltips
- rename `useEntityField` => `useEntityTooltips` because we also have `useEntityFields`
- if `useEntityTooltips` is undefined (because it is being called on the actual page, not the editor), return children directly rather than adding wrapping divs
- Set the tooltip div to be `width: fit-content` so the outline boxes match the underlying component

Footer
- Add padding to links to match Figma

CTAs/Buttons
- fix issue where background isn't updated when switched from Primary to Secondary
- eliminate unnecessary color transition

### DIFF
--- a/packages/visual-editor/src/components/editor/EntityField.tsx
+++ b/packages/visual-editor/src/components/editor/EntityField.tsx
@@ -29,11 +29,13 @@ export const EntityField = ({
   constantValueEnabled,
   children,
 }: EntityFieldProps) => {
-  const { tooltipsVisible } = useEntityField();
+  const tooltipsContext = useEntityTooltips();
 
-  if (constantValueEnabled) {
+  if (!tooltipsContext || constantValueEnabled) {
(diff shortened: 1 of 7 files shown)


asanehisa (approved):

packages/visual-editor/src/components/editor/EntityField.tsx
```diff
@@ -45,7 +47,7 @@ export const EntityField = ({
   }
 
   return (
-    <div>
+    <div id="tooltip" className="ve-w-fit">
```
mkilpatrick, line 50: I don't think we want an id on this. ids are supposed to be unique in the dom.

benlife5: ah yep, left over from debugging

mkilpatrick (approved):

PR #311, opened 2025-04-01
feat: add mapbox static map

asanehisa:
width, height, zoom, and mapStyle are hardcoded 
pinColor is based on the hexValue of the primary color (from our tailwind config)


packages/visual-editor/src/components/puck/MapboxStaticMap.tsx
```diff
@@ -0,0 +1,89 @@
...
+  mapStyle = "light-v11",
+}: MapboxStaticProps) => {
+  const document = useDocument();
+  const coordinate = resolveYextEntityField<Coordinate>(
+    document,
+    coordinateField
+  );
+
+  if (!coordinate) {
+    console.warn("yextDisplayCoordinate is not present in the stream");
```
benlife5, line 46: Maybe more generic? something like "coordinate not found"

packages/visual-editor/src/components/puck/MapboxStaticMap.tsx
```diff
@@ -0,0 +1,89 @@
...
+    return <></>;
+  } else if (!apiKey) {
+    console.warn("API Key is required for MapboxStaticMap");
+    return <></>;
+  }
+
+  const iframe = window.document.querySelector("iframe");
```
benlife5, line 53: Ah this is another bummer of the theme only existing as css variables in the editor. This won't work on the live page because there's no iframe, but I think all you might need is to try 
 to get `(JSON.parse(document.__.theme))?.["--colors-palette-primary"]` first and then go to this if that doesn't exist

asanehisa: ooooh thank u for catching this, can't believe i didnt think about live page

benlife5 (approved):

mkilpatrick (approved):

PR #312, opened 2025-04-01
fix: search combobox options by label

briantstephan:
shadcn's combobox component searches by value, but the values we pass in are often completely different from the labels that the user sees. This change passes the options from the BasicSelector into the combobox with the values set to match the labels, since the combobox component doesn't need to know what the actual values are. This way, the combobox searches by the values, it's really searching by labels, which is more intuitive for the user. 

Verified that the search functionality on the BasicSelector works for options with strings, numbers, or objects as values, and it still correctly selects options as expected.

https://github.com/user-attachments/assets/995bf259-a150-42da-bb76-196db2ea152e


benlife5 (approved):
BasicSelector has turned out be anything but basic haha
Thanks for figuring this out!

PR #313, opened 2025-04-01
refactor: button/link/cta color handling

benlife5:
I've switched from finding the actual rendered background color via the DOM to setting background color in context. The original method occurred after render and so would cause flashing on the page. The new method is slightly less ideal in that it requires intentionally setting the context in a parent rather that it working on its own. However, almost everything that sets a background color should use `<Section>` which will handle it for you, and I've exposed BackgroundProvider if you want to do something custom.


packages/visual-editor/src/hooks/README.md
```diff
@@ -108,3 +108,37 @@ const Edit: () => JSX.Element = () => {
...
+const MyComponent = () => {
+  return (
+    <Section background={{bgColor: "bg-white", textColor:"text-black"}}>
+      <Section background={{bgColor: "bg-blue", textColor:"text-red"}}>
+        <ChildComponent /> // prints '{bgColor: "bg-blue", textColor:"text-red"}'
+      </Section>
+      <ChildComponent /> // prints '{bgColor: "bg-white", textColor:"text-black"}'
+      <BackgroundProvider value={{bgColor: "bg-green", textColor:"text-purple"}}>
+        // background setting elements must be wrapped in BackgroundProvider
+        <div style={{backgroundColor: "green", color: "purple"}}>
```
mkilpatrick, line 137: Is this div necessary in the example? Slightly confusing if the color is coming from this or the provider.

packages/visual-editor/src/components/puck/atoms/button.tsx
```diff
@@ -71,60 +70,9 @@ export interface ButtonProps
 const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
   ({ className, variant, asChild = false, ...props }, ref) => {
     const Comp = asChild ? Slot : "button";
-    const [hasDarkBackground, setHasDarkBackground] = React.useState(false);
-
-    // handleRef determines if the active background color is light or dark and passes along the ref
-    const handleRef = React.useCallback((node: HTMLButtonElement | null) => {
```
mkilpatrick, line 77: Nice that all of this can be removed. Is none of the logic inside necessary somewhere else now?

packages/visual-editor/src/components/puck/atoms/button.tsx
```diff
@@ -71,60 +70,9 @@ export interface ButtonProps
...
-      // If the ref is resolved to an HTML element, get the parent color
```
mkilpatrick, line 78: And we don't need srgbToHSL either?

mkilpatrick (approved):

PR #315, opened 2025-04-02
feat: update result card styling to be in-line with mocks

Fondryext:
This PR adjusts the ordering and default styling of the result cards, to match the mocks and so styling choices update the cards as well.

### DIFF
--- a/starter/src/components/Locator.tsx
+++ b/starter/src/components/Locator.tsx
@@ -18,11 +18,17 @@ import {
 } from "@yext/search-headless-react";
 import * as React from "react";
 import { cva, VariantProps } from "class-variance-authority";
-import { BasicSelector, Button, themeManagerCn } from "@yext/visual-editor";
+import {
+  BasicSelector,
+  Button,
+  CTA,
+  Heading,
+  themeManagerCn,
+} from "@yext/visual-editor";
 import { LngLat, LngLatBounds } from "mapbox-gl";
 import { normalizeSlug } from "@yext/visual-editor";
 import { useEffect, useState } from "react";
-import { HoursStatus, Link } from "@yext/pages-components";
+import { HoursStatus } from "@yext/pages-components";
 import { Address, Hours } from "../types/autogen.ts";
 
 const DEFAULT_FIELD = "builtin.location";
@@ -239,12 +245,12 @@ const Locator: React.FC<LocatorProps> = (props) => {
             <Map {...mapProps} />
             {showSearchAreaButton && (
               <div className="absolute bottom-10 left-0 right-0 flex justify-center">
-                <button
+                <Button
                   onClick={handleSearchAreaClick}
-                  className="rounded-2xl border bg-white py-2 px-4 shadow-xl"
+                  className="rounded-2xl border bg-white py-2 px-4 shadow-xl text-black"
                 >
                   <p>Search This Area</p>
-                </button>
+                </Button>
               </div>
             )}
           </div>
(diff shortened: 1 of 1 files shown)


starter/src/components/Locator.tsx
```diff
@@ -305,68 +311,61 @@ const LocationCard: CardComponent<Location> = ({
...
+            className="text-sm"
+          />
+        )}
+        {location.mainPhone && (
+          <CTA
+            label={formatPhoneNumber(location.mainPhone)}
+            link={`tel:${location.mainPhone}`}
+            linkType={"PHONE"}
+            className="text-sm py-3"
+            target={"_blank"}
```
benlife5, line 332: I don't think you need a target for tel: links

starter/src/components/Locator.tsx
```diff
@@ -305,68 +311,61 @@ const LocationCard: CardComponent<Location> = ({
...
+          <HoursStatus
+            hours={location.hours}
+            timezone={location.timezone}
+            className="text-sm"
+          />
+        )}
+        {location.mainPhone && (
+          <CTA
+            label={formatPhoneNumber(location.mainPhone)}
+            link={`tel:${location.mainPhone}`}
```
benlife5, line 329: This works but I think CTA will prepend `tel:` for you with type PHONE

jwartofsky-yext:
Can you add some screenshots before/after?

Otherwise lgtm

starter/src/components/Locator.tsx
```diff
@@ -305,68 +311,61 @@ const LocationCard: CardComponent<Location> = ({
...
-                    "_blank",
-                  )
-                }
-                className="flex flex-col text-sm py-3 bg-rose-500 text-stone-50 rounded-lg drop-shadow-md"
-              >
-                Get Directions
-              </Button>
+    <div className="flex flex-wrap border-y px-4 py-4">
+      <div className="basis-3/4 pb-4">
+        <Heading className="text-lg py-2 text-palette-primary-dark" level={1}>
```
benlife5, line 316: Not sure if you meant to override the font sizes throughout -- for font sizes, there's 3 things you can do:
1. Using Heading/Body/Link with no font size set -- the values from theme will be automatically applied 
2. Apply a class like `text-h6-fontSize` or `text-body-sm-fontSize`/`text-body-lg-fontSize` to override the default value with another theme-based value 
3. Apply an fixed class like you've done here where the size will always be 1.125rem

Totally fine to override things if it seems like that's what the mock called for but wanted to check

starter/src/components/Locator.tsx
```diff
@@ -305,68 +311,61 @@ const LocationCard: CardComponent<Location> = ({
...
+          <CTA
+            label={formatPhoneNumber(location.mainPhone)}
+            link={`tel:${location.mainPhone}`}
+            linkType={"PHONE"}
+            className="text-sm py-3"
+            target={"_blank"}
+            variant="link"
+          />
+        )}
+        <p className="text-sm">{location.address.line1}</p>
```
benlife5, line 336: I'd use Body for these so they get the theme fonts/weights/colors/etc.

starter/src/components/Locator.tsx
Fondryext: Nah you're totally right. Adjusted

starter/src/components/Locator.tsx
Fondryext: Good call

jwartofsky-yext (approved):
+2

PR #316, opened 2025-04-02
refactor: additional layout updates

benlife5:
Spacing Selector
- remove default option from fields that do not have theme-based defaults

Address and HoursTable
- remove padding

Flex and Grid
- add default gap and Vertical Padding
- remove Max Content Width as prop (always uses theme's value)
- fix when dragging between drop zones, reset layout props
- add additional disallowed Page Sections from being dragged into Flex/Grid 

Root 
- by rendering the DropZone with id `"default-zone"`, we can override the root drop zone (@mkilpatrick don't need to open a puck issue because I found this in the Discord)
- disallow content blocks from being dragged directly onto the page

Header and Breadcrumbs
- standardize margin/padding to respect max-width
- add mobile padding padding on mobile 
- fix mobile width overflow

Footer
- standardize margin/padding to respect max-width
- add vertical padding to match mocks
- handle link colors correctly
- push footer to the bottom of the page

https://github.com/user-attachments/assets/b6d65dbb-b701-470f-8047-cdd2cbf3f14c

https://github.com/user-attachments/assets/0f0b1ca6-450b-46bc-a442-e483fda0148e

### DIFF
--- a/packages/visual-editor/src/components/editor/SpacingSelector.tsx
+++ b/packages/visual-editor/src/components/editor/SpacingSelector.tsx
@@ -78,9 +78,9 @@ const convertDefaultSpacingsToOptions = (
 };
 
 export const SpacingSelector = (
+  label: string,
   spacingType: "padding" | "gap",
-  includeDefault: boolean,
-  label?: string
+  includeDefault: boolean
 ): Field => {
   return {
     type: "custom",
@@ -88,10 +88,7 @@ export const SpacingSelector = (
       const tailwindConfig: TailwindConfig = useTailwindConfig();
 
       return (
-        <FieldLabel
-          label={label ?? spacingType}
-          icon={<ChevronDown size={16} />}
-        >
+        <FieldLabel label={label} icon={<ChevronDown size={16} />}>
           <AutoField
             value={value}
             onChange={onChange}
--- a/packages/visual-editor/src/components/editor/SpacingSelector.tsx
+++ b/packages/visual-editor/src/components/editor/SpacingSelector.tsx
@@ -79,6 +79,7 @@ const convertDefaultSpacingsToOptions = (
 
 export const SpacingSelector = (
   spacingType: "padding" | "gap",
+  includeDefault: boolean,
   label?: string
 ): Field => {
   return {
@@ -98,10 +99,12 @@ export const SpacingSelector = (
               type: "select",
               options: convertDefaultSpacingsToOptions(
                 spacingType,
-                [
-                  { label: "Default", value: "default", px: "" },
-                  ...spacingOptions,
-                ],
+                includeDefault
+                  ? [
+                      { label: "Default", value: "default", px: "" },
+                      ...spacingOptions,
+                    ]
+                  : spacingOptions,
                 tailwindConfig
               ),
             }}
--- a/packages/visual-editor/src/components/puck/Card.tsx
+++ b/packages/visual-editor/src/components/puck/Card.tsx
@@ -1,6 +1,7 @@
 import * as React from "react";
 import { ComponentConfig, Fields } from "@measured/puck";
 import {
+  ThemeOptions,
   themeManagerCn,
   useDocument,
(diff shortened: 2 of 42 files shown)


packages/visual-editor/src/components/puck/Card.tsx
```diff
@@ -73,12 +73,15 @@ const cardFields: Fields<CardProps> = {
...
-      transform: BasicSelector("Text Transform", [
-        { value: "none", label: "None" },
-        { value: "lowercase", label: "Lowercase" },
-        { value: "uppercase", label: "Uppercase" },
-        { value: "capitalize", label: "Capitalize" },
-      ]),
+      transform: BasicSelector<CardProps["subheading"]["transform"]>(
+        "Text Transform",
+        [
+          { value: "none", label: "None" },
```
mkilpatrick, line 79: Noting that this will clash with Alexis' PR to simplify the theme options.

packages/visual-editor/src/components/puck/Card.tsx
```diff
@@ -73,12 +73,15 @@ const cardFields: Fields<CardProps> = {
       }),
       fontSize: FontSizeSelector(),
       weight: BasicSelector("Font Weight", []),
-      transform: BasicSelector("Text Transform", [
-        { value: "none", label: "None" },
-        { value: "lowercase", label: "Lowercase" },
-        { value: "uppercase", label: "Uppercase" },
-        { value: "capitalize", label: "Capitalize" },
-      ]),
+      transform: BasicSelector<CardProps["subheading"]["transform"]>(
```
mkilpatrick, line 76: The typing makes it a little more gross to use

packages/visual-editor/src/components/puck/Flex.tsx
```diff
@@ -62,7 +59,7 @@ const FlexContainer = React.forwardRef<HTMLDivElement, FlexProps>(
               flexDirection: direction,
               flexWrap: wrap,
             }}
-            disallow={["Banner"]}
+            disallow={["Banner", "Promo", "Card", "Breadcrumbs"]}
```
mkilpatrick, line 62: Is there a function for each group? If we had one we could use that here so we don't need to remember to update it all the time.

packages/visual-editor/src/components/editor/SpacingSelector.tsx
```diff
@@ -79,6 +79,7 @@ const convertDefaultSpacingsToOptions = (
 
 export const SpacingSelector = (
   spacingType: "padding" | "gap",
+  includeDefault: boolean,
   label?: string
```
mkilpatrick, line 83: I feel like we should make this required instead of defaulting to the spacingType. Then I'd make it the first prop to be consistent with the BasicSelector.

packages/visual-editor/src/components/puck/Card.tsx
benlife5: Hopefully with Alexis's PR we can hide this typing. String union types in props cause issues because the array of here options are just strings.

packages/visual-editor/src/components/puck/Card.tsx
benlife5: Yeahhh, probably going to be a number of conflicts :/ @asanehisa I think it probably makes sense to merge this after yours

packages/visual-editor/src/components/puck/Flex.tsx
benlife5: No, we should add one but need to think if there's any way for this to be dev-configurable

mkilpatrick: I think we can worry about the DXP for that later.

asanehisa (approved):

mkilpatrick (approved):

benlife5:
Removed the BasicSelector typing because it was causing problems

asanehisa (approved):

PR #328, opened 2025-04-04
feat: collections

benlife5:
- refactor NumberOrDefaultField (not currently used) into a more generic OptionalNumberField
    - The generic version might be too confusing and it would be better to just have a LimitField
- add CollectionSection component
- add ExampleRepeatableItem, which is a temporary demo of how to build a card that uses the collection
- update YextEntityFieldSelector and resolveYextEntityField to handle subfields
- will update READMEs once this is more settled

Demo site: https://dev.yext.com/s/1000152098/yextsites/63648/pagesets
c_featuredMenuItems and c_featuredDesserts are populated on that account

- When the parent switches between entity field and constant value, all children are removed
- When the parent changes entity field/limit, the child must be clicked to get the update
- When the parent entity field changes, all child entity fields are cleared

V1 video demo: https://drive.google.com/file/d/16BLmBDCWues5n8c2IaPAqLFS888hh7ha/view?usp=sharing
V2 video demo: https://drive.google.com/file/d/1Cs-T2Z_eQe4RCUvGydAISYdZGwfPhce5/view?usp=drive_link

1. When dragged on to the page normally, the card acts like a normal card
2. CollectionSection lets you pick a title, collection field, and limit
3. When constant value is selected in CollectionSection, the RepeatableItem acts as a single card with all fields available
4. When a list of objects is chosen in CollectionSection, the RepeatableItem allows you to pick subfields that object
5. When a list of non-objects is chosen in CollectionSection, the RepeatableItem allows you to pick that field
6. When constant value is selected in RepeatableItem, all cards get that value
   - I think Aaron wanted this to be an array of constant values, I initially ran into issues with this but can try again


packages/visual-editor/src/components/puck/ExampleRepeatableItem.tsx
```diff
@@ -0,0 +1,168 @@
...
+      {text && <p>{text}</p>}
+    </div>
+  );
+};
+
+export const ExampleRepeatableItemComponent: ComponentConfig<ExampleRepeatableItemProps> =
+  {
+    label: "Repeatable Card",
+    fields: ExampleRepeatableItemFields,
+    resolveFields: (data, params) => {
```
benlife5, line 115: I was hoping to abstract more of this away but the returns fields are very component-specific so it's not possible to make it a one-liner. Might be room for improvement

packages/visual-editor/src/components/puck/contexts/collectionContext.tsx
```diff
@@ -0,0 +1,15 @@
+import { createContext, useContext } from "react";
+import { YextEntityField } from "../../../index.js";
+
+export const CollectionContext = createContext<{
```
benlife5, line 4: Depending on the resolution in the Puck thread about parent in resolveFields/resolveData, I might be able to get rid of this context and just use that

packages/visual-editor/src/components/puck/CollectionSection.tsx
```diff
@@ -0,0 +1,178 @@
...
+      <CollectionContext.Provider
+        value={{
+          parentEntityField: collection.items,
+          limit:
+            typeof collection.limit === "number" ? collection.limit : undefined,
+        }}
+      >
+        <DropZone
+          zone="collection-dropzone"
+          allow={["ExampleRepeatableItemComponent"]}
```
mkilpatrick, line 119: We should turn this into a maintained list

packages/visual-editor/src/components/puck/CollectionSection.tsx
```diff
@@ -0,0 +1,178 @@
...
+          className="flex gap-4 max-w-pageSection-maxWidth flex-wrap"
+        />
+      </CollectionContext.Provider>
+    </Section>
+  );
+};
+
+export const CollectionSection: ComponentConfig<CollectionSectionProps> = {
+  label: "Collection Section",
```
mkilpatrick, line 128: I wonder if this should really be a Section itself or it should really just be a wrapper of items and you'd put the Collection in a Section. I feel like the latter would be more flexible. That's a Product decision though.

packages/visual-editor/src/components/puck/ExampleRepeatableItem.tsx
mkilpatrick, line 115: I think if the parent changes between constant field/entity mapping we should clear out the children. We'd have to ask Aaron. I was able to set collection to constant value, add a repeatable item and configure it, then change the collection to entity field, then I had 4 items in the collection: my first constant one followed by the auto mapped items. Unclear if this is desirable or not.

mkilpatrick: Maybe we can turn it into a function so that lines 116 - 128 don't need to be repeated and you get access to data and parent to return the fields (text and image in this case). The lastFields can be handled automatically, assuming all of that is consistent.

packages/visual-editor/src/internal/utils/getFilteredEntityFields.ts
```diff
@@ -182,8 +192,25 @@ export const getFilteredEntityFields = <T extends Record<string, any>>(
     }
   }
 
+  if (filter?.directChildrenOf) {
```
mkilpatrick, line 195: Add to getFilteredEntityFields.test.ts

packages/visual-editor/src/components/puck/CollectionSection.tsx
```diff
@@ -0,0 +1,178 @@
...
+    type: "object",
+    label: "Collection",
+    objectFields: {
+      items: YextEntityFieldSelector<any, Array<any>>({
+        label: "Items",
+        isCollection: true,
+        filter: {
+          includeListsOnly: true,
+          // Filter hours fields because they have a lot of unused children
+          disallowList: [
```
mkilpatrick, line 69: Maybe we can auto filter hours when including lists

packages/visual-editor/src/utils/resolveYextEntityField.ts
```diff
@@ -40,3 +41,84 @@ export const resolveYextEntityField = <T>(
   console.warn(`The field ${entityField.field} was not found in the document.`);
   return undefined;
 };
+
+export const resolveYextSubfield = <T>(
```
mkilpatrick, line 45: Add tests

packages/visual-editor/src/components/editor/OptionalNumberField.tsx
```diff
@@ -0,0 +1,66 @@
...
+  showNumberFieldRadioLabel: string;
+  defaultCustomValue: number;
+};
+
+type OptionalNumberFieldRenderProps = {
+  value: number | string;
+  onChange: (e: number | string) => void;
+};
+
+export const OptionalNumberField = ({
```
asanehisa, line 17: Hmm I imagine we wanna add this to README (and delete NumberOrDefaultField if it is there)

packages/visual-editor/src/components/editor/YextEntityFieldSelector.tsx
```diff
@@ -212,7 +277,7 @@ const ConstantValueInput = <T extends Record<string, any>>({
   ) : (
     <FieldLabel
       label={constantFieldConfig.label ?? "Value"}
-      className="ve-inline-block ve-pt-4 w-full"
+      className="ve-inline-block w-full"
```
asanehisa, line 280: no u but ve-w-full

packages/visual-editor/src/components/puck/CollectionSection.tsx
```diff
@@ -0,0 +1,178 @@
...
+}
+
+const collectionSectionFields: Fields<CollectionSectionProps> = {
+  styles: {
+    type: "object",
+    label: "Styles",
+    objectFields: {
+      background: BasicSelector(
+        "Background",
+        Object.values(backgroundColors).map(({ label, value }) => ({
```
asanehisa, line 40: ThemeOptions.BACKGROUND_COLOR

packages/visual-editor/src/components/puck/ExampleRepeatableItem.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  const collectionContext = useCollectionContext();
+
+  // If not in a collection, render a single card with normal entity fields
+  if (!isCollection || !collectionContext) {
+    const resolvedText = resolveYextEntityField(document, text);
+    const resolvedImage = resolveYextEntityField(document, image);
+
+    return (
+      <div>
```
asanehisa, line 50: div wrapper seems unnecessary?

packages/visual-editor/src/components/puck/ExampleRepeatableItem.tsx
```diff
@@ -0,0 +1,168 @@
...
+        );
+      })}
+    </div>
+  );
+};
+
+type ExampleRepeatableItemCardProps = {
+  image?: ImageType | ComplexImageType;
+  text?: string;
+};
```
asanehisa, line 81: empty line under

packages/visual-editor/src/components/puck/ExampleRepeatableItem.tsx
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  return (
+    <div className="w-[200px] h-[200px] border-dashed border-black border-4 flex flex-col items-center justify-center">
+      {image?.url && (
+        <div className="mx-8">
+          <Image image={image} layout="fixed" width={100} aspectRatio={1} />
+        </div>
+      )}
+      {text && <p>{text}</p>}
```
asanehisa, line 106: Don't wanna use body?

packages/visual-editor/src/components/puck/contexts/collectionContext.tsx
```diff
@@ -0,0 +1,15 @@
...
+  parentEntityField: YextEntityField<Array<any>>;
+  limit: undefined | number;
+} | null>(null);
+
+export const useCollectionContext = () => {
+  const context = useContext(CollectionContext);
+  if (!context) {
+    return undefined;
```
asanehisa, line 12: warning might be nice

packages/visual-editor/src/internal/utils/getFilteredEntityFields.ts
```diff
@@ -172,7 +179,10 @@ export const getFilteredEntityFields = <T extends Record<string, any>>(
   // Augment to include subfields
   let filteredEntitySubFields: YextSchemaField[] = [];
   for (const yextSchemaField of filteredEntityFields) {
-    const entityFieldNames = getEntityFieldNames(yextSchemaField);
+    const entityFieldNames = getEntityFieldNames(
+      yextSchemaField,
+      Boolean(filter?.directChildrenOf?.length)
```
asanehisa, line 184: Oh any diff from !!(filter?.directChildrenOf?.length)

packages/visual-editor/src/utils/resolveYextEntityField.ts
```diff
@@ -40,3 +41,84 @@ export const resolveYextEntityField = <T>(
...
+
+  const subfieldName = splitName.pop();
+  if (!subfieldName) {
+    return;
+  }
+
+  return parent[subfieldName] as T;
+};
+
+export const handleResolveFieldsForCollections = (
```
benlife5, line 87: Need to add test once updated

benlife5:
- updated README and tests
- eliminated collectionContext
- the child component now only updates when it is clicked on in the editor. I think this is a more predicable solution than the previous strange mix of some updates when the parent updated and other updates and other only when the child was clicked

Am now going to try one more idea for clearing the children and then will update test site and video

packages/visual-editor/src/components/editor/OptionalNumberField.tsx
```diff
@@ -0,0 +1,66 @@
...
+  fieldLabel,
+  hideNumberFieldRadioLabel,
+  hideNumberFieldRadioValue,
+  showNumberFieldRadioLabel,
+  defaultCustomValue,
+}: OptionalNumberFieldProps): CustomField<number | string> => {
+  const showNumberField = showNumberFieldRadioLabel.toLowerCase();
```
asanehisa, line 24: is there a reason u have hideNumberFieldRadioValue ? 
Could it be hideNumberFieldRadioLabel.toLowerCase()

packages/visual-editor/src/utils/resolveYextEntityField.test.ts
```diff
@@ -1,56 +1,294 @@
...
-    ),
-    "potato land"
-  );
+        { field: "address.city", constantValue: "" }
+      ),
+      "potato land"
+    );
+  });
+
+  it("returns empty value when field found in document", async () => {
```
asanehisa, line 23: returns empty value when empty field found in document?

benlife5:
Added clearing the children by messing with Puck history... it's a little dangerous so I need to test some more to make sure this doesn't screw up anything else
The test site has been updated and here is a new video: https://drive.google.com/file/d/1Cs-T2Z_eQe4RCUvGydAISYdZGwfPhce5/view?usp=drive_link

packages/visual-editor/src/components/puck/ExampleRepeatableItem.tsx
```diff
@@ -8,69 +8,61 @@ import {
...
-        const resolvedText = resolveYextSubfield(item, text);
-        return (
-          <ExampleRepeatableItemCard
-            key={i}
-            image={resolvedImage}
-            text={resolvedText}
-          />
-        );
-      })}
+      {resolvedParent
```
mkilpatrick, line 53: Is there a way we can abstract the logic (or create functions) so that when we need to make the individual items work in a collection it's easy to do without repeating all of this stuff? And also when we can update things when Puck supports what we need we only need to update one thing and not every component.

Edit: I think you've done this probably as much as you can already and this is necessary because it's specific to this component. Just checking though.

benlife5: Yeah, I think this is about as abstracted as it can be. We could maybe expose some render logic for whether to do a single card or multiple, but there'd have to be so many config options that I think it's better to supply the render function in full

PR #319, opened 2025-04-03
feat: add Products Section component

asanehisa:
https://github.com/user-attachments/assets/681d09ee-e425-4b50-9975-08a891332adf

adding items: https://jam.dev/c/8bf304e2-1a07-4d8b-baa9-f82078e67c00

PR #329, opened 2025-04-04
feat: add NearbyLocations Component

jwartofsky-yext:
Adds the component "Nearby Locations"

This component gets nearby locations using the Content API

Entity fields had to be omitted due to extra complexity. They are hardcoded to use the fields name, address, hours, and mainPhone.

Screenshot 2025-04-04 at 2 54 12 PM

Mobile
image


packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,293 @@
...
+        {nearbyLocationsStatus === "success" &&
+          !!nearbyLocationsData?.response?.docs && (
+            <div className="grid grid-cols-1 md:grid-cols-3 gap-4 items-stretch">
+              {nearbyLocationsData.response.docs.map(
+                (location: any, index: number) => (
+                  <div
+                    key={index}
+                    className="flex flex-col flew-grow h-full rounded-lg overflow-hidden border"
+                  >
+                    <Section
```
jwartofsky-yext, line 195: I've been having trouble getting this styling to apply to the entire card. h-full and flex-grow don't seem to make this take up the entire space.

If I inverse this and the div above, the rounded corners don't seem to work.

I've also tried replacing the div above with the <Section component and combining the className styling from both, but the rounded corners don't work that way.

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,329 @@
...
+  backgroundColors,
+  ThemeOptions,
+  HeadingLevel,
+} from "../../index.js";
+import { useQuery } from "@tanstack/react-query";
+import { Address, Coordinate, HoursStatus } from "@yext/pages-components";
+import { fetchNearbyLocations } from "../../api/nearbyLocations.tsx";
+import * as React from "react";
+import { formatPhoneNumber } from "./Phone.tsx";
+import { Section } from "./atoms/section.js";
```
asanehisa, line 19: import from "../../index.js"

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,329 @@
...
+      showCurrentStatus: boolean;
+      timeFormat?: "12h" | "24h";
+      dayOfWeekFormat?: "short" | "long";
+      showDayNames?: boolean;
+    };
+  };
+  coordinate: YextEntityField<Coordinate>;
+  radius: number;
+  entityType:
+    | "any"
```
asanehisa, line 43: Hmm we'll wanna this hardcoded as a const somewhere probably

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,329 @@
...
+    filter: { types: ["type.coordinate"] },
+  }),
+  radius: {
+    label: "Radius (Miles)",
+    type: "number",
+  },
+  entityType: {
+    label: "Entity Type",
+    type: "select",
+    options: [
```
asanehisa, line 143: Hmm if we are gonna need this for other components in future, lets add to ThemeOptions

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,329 @@
...
+                            }}
+                            dayOfWeekTemplate={
+                              cards.hours.showDayNames ? undefined : () => <></>
+                            }
+                            className="h-full"
+                          />
+                        </div>
+                      )}
+                      {location.mainPhone && (
+                        <div className="font-body-fontFamily font-body-fontWeight text-body-fontSize-sm">
```
asanehisa, line 266: should this be a CTA?

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+  return (
+    <Section background={styles.backgroundColor} className={`components`}>
+      <div className="space-y-6">
+        {/* Title & CTA Section */}
+        {headingText && (
+          <div className="flex flex-col md:flex-row justify-between items-center md:items-start">
+            <Heading
+              level={heading.level}
+              className="text-2xl font-bold text-center md:text-left"
```
asanehisa, line 214: hmmm is this suppose to have hardcoded classnames? text size etc should be decided by heading level usually

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,329 @@
...
+  } = props;
+  const document = useDocument<any>();
+
+  const coordinate = resolveYextEntityField<Coordinate>(
+    document,
+    coordinateField
+  );
+  const headingText = resolveYextEntityField<string>(document, heading.text);
+
+  const contentApiKey: string = document?._env?.YEXT_CONTENT_API_KEY;
```
asanehisa, line 175: add a warning if no YEXT_CONTENT_API_KEY?

packages/visual-editor/src/api/nearbyLocations.tsx
```diff
@@ -0,0 +1,57 @@
...
+  entityType,
+}: {
+  businessId: number;
+  apiKey: string;
+  longitude: number;
+  latitude: number;
+  radiusMi: number;
+  entityType?: string;
+}): Promise<any> => {
+  const base = `https://streams-dev.yext.com/v2/accounts/${businessId}/content/visualEditorLocations`;
```
mkilpatrick, line 18: Make visualEditorLocations a separate const near the top of the file and comment that this must align with the autogenerated Content API endpoint that's created when a VLE site is created in Storm

packages/visual-editor/src/api/nearbyLocations.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const queryParams: any = {
+    api_key: apiKey,
+    v: getCurrentDateYYYYMMDD(),
```
mkilpatrick, line 21: You need to set an actual date. If you always get the latest date and the api changes then it would be breaking. Date pins to a specific version of the api.

packages/visual-editor/src/api/nearbyLocations.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const response = await fetch(route);
+
+  if (!response.ok) {
+    throw new Error(response.statusText);
+  }
+
+  return await response.json();
+};
+
+export const addQueryParams = (
```
mkilpatrick, line 39: Use what's build into JS

```
let url = new URL("https://example.com/api/resource");
url.searchParams.append("key1", "value1");
```

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,329 @@
...
+    hours: {
+      showCurrentStatus: boolean;
+      timeFormat?: "12h" | "24h";
+      dayOfWeekFormat?: "short" | "long";
+      showDayNames?: boolean;
+    };
+  };
+  coordinate: YextEntityField<Coordinate>;
+  radius: number;
+  entityType:
```
mkilpatrick, line 42: Per the thread can we get this from the pageset somehow instead of making it a prop?

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,329 @@
...
+import * as React from "react";
+import { formatPhoneNumber } from "./Phone.tsx";
+import { Section } from "./atoms/section.js";
+
+export interface NearbyLocationsProps {
+  styles: {
+    backgroundColor?: BackgroundStyle;
+    cardBackgroundColor?: BackgroundStyle;
+  };
+  heading: {
```
mkilpatrick, line 26: I feel like radius should be a user option though we should vet with @apav-dev. We'd have to look into what the api radius limits/options are.

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,329 @@
...
+      { label: "ATM", value: "atm" },
+      { label: "Restaurant", value: "restaraunt" },
+      { label: "Hotel", value: "hotel" },
+      { label: "Provider Facility", value: "providerFacility" },
+      { label: "Financial Professional", value: "financialProfessional" },
+    ],
+  },
+};
+
+const NearbyLocationsWrapper: React.FC<NearbyLocationsProps> = (props) => {
```
mkilpatrick, line 158: NearbyLocationsComponent

packages/visual-editor/src/components/puck/NearbyLocations.tsx
mkilpatrick: Section is the outermost thing that wraps everything inside it.

packages/visual-editor/src/components/puck/NearbyLocations.tsx
jwartofsky-yext: Going to make a Phone atom and we can use that here

packages/visual-editor/src/components/puck/NearbyLocations.tsx
jwartofsky-yext: It lets me query up to 500,000 km and more, so I don't think there's a limit

It's just comparing to the geo field that's returned from the Content endpoint

packages/visual-editor/src/api/nearbyLocations.tsx
jwartofsky-yext: I'll add a map of env to API. The URL returned by the content API is actually incorrect for dev, this is the correct one for us3dev.

I would imagine the other environments are correct though.

Do we want to pass down the Content API base url through an env var instead? That way we wouldn't have to include a const map. We can handle the bad us3dev url in VES until KG fixes it.

mkilpatrick: I was just talking about the last part (the actual endpoint). I actually didn't notice this is hardcoded to dev. How would a map work here? We might need to put the entire url in as an env var.

packages/visual-editor/src/components/puck/NearbyLocations.tsx
jwartofsky-yext: I was using a pattern similar to this:
https://github.com/yext/visual-editor/pull/319/files#diff-85ef3401cc00632e072ab33b722a7fd6bbc41a94cc8ba616613d864f8cae6fc3R106

However Section doesn't seem work with rounded corners properly if it's the outermost component.

mkilpatrick: I think you're looking to use the  <section> html element vs our Section atom. Maybe I'm wrong though?

packages/visual-editor/src/components/puck/NearbyLocations.tsx
jwartofsky-yext: Yes, we can get it from the document. 

meta.entityType.id

Verified that this is available on my other test site

packages/visual-editor/src/api/nearbyLocations.tsx
jwartofsky-yext: Went with putting the entire url in as the env var. We then add v param after.

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,323 @@
...
+      },
+    },
+  },
+  coordinate: YextEntityFieldSelector<any, Coordinate>({
+    label: "Coordinates",
+    filter: { types: ["type.coordinate"] },
+  }),
+  radius: {
+    label: "Radius (Miles)",
+    type: "number",
```
asanehisa, line 138: prob want a min:0

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,323 @@
...
+  const coordinate = resolveYextEntityField<Coordinate>(
+    document,
+    coordinateField
+  );
+  const headingText = resolveYextEntityField<string>(document, heading.text);
+
+  const contentEndpoint: string = document?._env?.YEXT_CONTENT_API_KEY;
+  if (!contentEndpoint) {
+    console.warn(
+      "Missing YEXT_CONTENT_ENDPOINT! Unable to fetch nearby locations."
```
asanehisa, line 167: should it say YEXT_CONTENT_API_KEY rather than YEXT_CONTENT_ENDPOINT?

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,323 @@
...
+    );
+  }
+
+  const entityType: string = document?.meta?.entityType?.id || "locations";
```
asanehisa, line 171: "location" rather than "locations" I think

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,323 @@
...
+                    className="flex flex-col flew-grow h-full rounded-lg overflow-hidden border"
+                  >
+                    <section
+                      className={themeManagerCn(
+                        "components h-full flex-grow",
+                        styles.cardBackgroundColor?.bgColor || "",
+                        styles.cardBackgroundColor?.textColor || ""
+                      )}
+                    >
+                      <Heading level={4} className="text-lg font-bold">
```
asanehisa, line 228: is the text-lg and font-bold purposeful override for heading level?

packages/visual-editor/src/components/puck/NearbyLocations.tsx
jwartofsky-yext: no, it's the whole content endpoint now (including API key)

We need to pass the content endpoint in because it varies by environment

packages/visual-editor/src/components/puck/NearbyLocations.tsx
jwartofsky-yext: Not needed, removed it

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,323 @@
...
+  backgroundColors,
+  ThemeOptions,
+  HeadingLevel,
+  Section,
+  Phone,
+  themeManagerCn,
+} from "../../index.js";
+import { useQuery } from "@tanstack/react-query";
+import { Address, Coordinate, HoursStatus } from "@yext/pages-components";
+import { fetchNearbyLocations } from "../../api/nearbyLocations.tsx";
```
benlife5, line 19: I think function should either be under `components/puck` or exported through the package utils

packages/visual-editor/src/api/nearbyLocations.tsx
```diff
@@ -0,0 +1,38 @@
...
+  limit,
+  entityType,
+}: {
+  contentEndpoint: string;
+  longitude: number;
+  latitude: number;
+  radiusMi: number;
+  limit: number;
+  entityType: string;
+}): Promise<any> => {
```
benlife5, line 17: Promise<Record<string,any>>

starter/package.json
```diff
@@ -26,6 +26,7 @@
     "@radix-ui/react-slot": "^1.0.2",
     "@radix-ui/react-toast": "^1.1.5",
     "@radix-ui/react-tooltip": "^1.1.2",
+    "@tanstack/react-query": "^5.71.5",
```
benlife5, line 29: Does the starter need react-query?

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,323 @@
...
+          <div className="flex flex-col md:flex-row justify-between items-center md:items-start">
+            <Heading level={heading.level} className="text-center md:text-left">
+              {headingText}
+            </Heading>
+          </div>
+        )}
+
+        {nearbyLocationsStatus === "success" &&
+          !!nearbyLocationsData?.response?.docs && (
+            <div className="grid grid-cols-1 md:grid-cols-3 gap-4 items-stretch">
```
benlife5, line 216: Were you able to get all the cards to be the same height?

jwartofsky-yext: Not yet

Content API is down on dev right now so that is making testing more difficult

May try to fake it in the meantime

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -146,6 +145,73 @@ const nearbyLocationsFields: Fields<NearbyLocationsProps> = {
...
+  address,
+  hours,
+  timezone,
+  mainPhone,
+  phoneNumberFormat,
+  phoneNumberLink,
+  backgroundColor,
+}: {
+  key: number;
+  name: string;
```
benlife5, line 160: I think you can use NearbyLocationsProps["cards"] here

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -146,6 +145,73 @@ const nearbyLocationsFields: Fields<NearbyLocationsProps> = {
...
+  hours: any;
+  timezone: string;
+  mainPhone: string;
+  phoneNumberFormat: "domestic" | "international";
+  phoneNumberLink: boolean;
+  backgroundColor?: BackgroundStyle;
+}) => {
+  return (
+    <div
+      key={key}
```
benlife5, line 171: I think key should be consumed when you  `<LocationCard key=` and you don't have to pass it along as a prop

asanehisa (commented):
can u update the components registry too?

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,339 @@
...
+          },
+          showDayNames: {
+            type: "radio",
+            label: "Show Day Names",
+            options: [
+              { label: "Yes", value: true },
+              { label: "No", value: false },
+            ],
+          },
+          dayOfWeekFormat: BasicSelector("Day of Week Format", [
```
mkilpatrick, line 107: Check with @apav-dev if this should be a selector or radio. Our other ones that have 3 or less options seem to be radio.

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,339 @@
...
+  } = props;
+  const document = useDocument<any>();
+
+  const coordinate = resolveYextEntityField<Coordinate>(
+    document,
+    coordinateField
+  );
+  const headingText = resolveYextEntityField<string>(document, heading.text);
+
+  const contentEndpoint: string = document?._env?.YEXT_CONTENT_API_KEY;
```
mkilpatrick, line 227: This is YEXT_CONTENT_ENDPOINT per your updated CR https://gerrit.yext.com/c/alpha/+/289517/1/gocode/src/yext/publish/visualeditor/server/cmd/visualeditorserver/server/site/createVisualEditorSite.go#58

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -0,0 +1,339 @@
...
+  address: any;
+  hours: any;
+  timezone: string;
+  mainPhone: string;
+  backgroundColor?: BackgroundStyle;
+}) => {
+  return (
+    <div className="flex flex-col flew-grow h-full rounded-lg overflow-hidden border">
+      <Section background={backgroundColor} className={`components`}>
+        <Heading level={4}>{name}</Heading>
```
benlife5, line 168: Should this be using the cards.headingLevel field?

packages/visual-editor/src/components/puck/NearbyLocations.tsx
benlife5: Oh I'm sorry I thought these matched 1:1 with NearbyLocationsProps["cards"]. I think the old way is probably better

packages/visual-editor/src/components/puck/NearbyLocations.tsx
jwartofsky-yext: They still have variable height

Going to revisit this once your <Section> component updates are out

packages/visual-editor/src/components/puck/NearbyLocations.tsx
jwartofsky-yext: I think using cards prop as you suggested originally is still good.

The "cards" props are things the user set

The other props come from the nearbyLocations returned from the Content Endpoint

packages/visual-editor/src/components/puck/NearbyLocations.tsx
jwartofsky-yext: Going to revisit when Ben's section component updates are out

Currently I have the location card and the whole component each wrapped in their own Section, which seems like the correct pattern as used by the other components

benlife5 (approved):

benlife5 (approved):

packages/visual-editor/src/components/puck/registry/components.ts
```diff
@@ -180,7 +185,7 @@ export const ui: Registry["items"] = [
   {
     name: "Phone",
```
asanehisa, line 186: i think this is PhoneWrapper now?

packages/visual-editor/src/components/puck/registry/components.ts
```diff
@@ -45,6 +45,11 @@ export const ui: Registry["items"] = [
     type: "registry:component",
     files: [{ path: "atoms/section.tsx", type: "registry:component" }],
   },
+  {
+    name: "phone",
+    type: "registry:component",
```
asanehisa, line 50: registryDependencies: ["body", "cta"],

packages/visual-editor/src/components/puck/registry/components.ts
jwartofsky-yext: Based on how ImageWrapper is setup, it should still be "Phone" here

packages/visual-editor/src/components/puck/registry/components.ts
asanehisa: Hmm does mean we ideally don't want an Phone atom at all / if we do, we want it to not use the body and cta atoms?

jwartofsky-yext: I think for our design the current setup makes the most sense and is easiest to use.

Maybe we can call these "molecules" or something in the future.

I'd be ok with just having this be an "atom" for now. If we have other "molecules" in the future we can think about labelling them differently if we want.

asanehisa (approved):
gonna 100% merge conflict with Core Info Section PR

benlife5 (approved):

PR #333, opened 2025-04-04
feat: add Events component

colton-demetriou:
Screenshot 2025-04-04 at 5 06 51 PM

Tested locally

### DIFF
--- a/packages/visual-editor/src/components/puck/Events.tsx
+++ b/packages/visual-editor/src/components/puck/Events.tsx
@@ -185,7 +185,7 @@ const EventCard: React.FC<EventCardProps> = ({
               link={resolvedCTA?.link || "#"}
               linkType={resolvedCTA?.linkType}
               variant="link"
-              className="text-palette-primary"
+              className="text-palette-primary-dark"
             />
           </EntityField>
         )}
--- a/packages/visual-editor/src/components/puck/Events.tsx
+++ b/packages/visual-editor/src/components/puck/Events.tsx
@@ -1,0 +1,380 @@
+import * as React from "react";
+import { ComponentConfig, Fields, ArrayField } from "@measured/puck";
+import {
+  YextEntityField,
+  YextEntityFieldSelector,
+  useDocument,
+  resolveYextEntityField,
+  BasicSelector,
+  ThemeOptions,
+  Heading,
+  HeadingLevel,
+  Body,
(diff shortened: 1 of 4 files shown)


briantstephan (dismissed):

packages/visual-editor/src/components/puck/Events.tsx
```diff
@@ -0,0 +1,380 @@
...
+            displayName="CTA"
+            fieldId={cta.field}
+            constantValueEnabled={cta.constantValueEnabled}
+          >
+            <CTA
+              label={resolvedCTA?.label}
+              link={resolvedCTA?.link || "#"}
+              linkType={resolvedCTA?.linkType}
+              variant="link"
+              className="text-palette-primary"
```
benlife5, line 188: I don't think we want to set this

packages/visual-editor/src/components/puck/Events.tsx
```diff
@@ -0,0 +1,380 @@
...
+  const resolvedHeading = resolveYextEntityField<string>(
+    document,
+    sectionHeading.text
+  );
+
+  return (
+    <Section background={backgroundColor} className="components">
+      <div className="flex flex-col gap-12 p-8">
+        {resolvedHeading && (
+          <div className="text-center">
```
benlife5, line 209: Can `"text-center"` be applied to either the outer div or the heading and then we can eliminate this div?

packages/visual-editor/src/components/puck/Events.tsx
colton-demetriou: need to for the CTA color to be correct, had to manually set background/text color to b/w to get the image to be the height of the card so needed to override, but should be constant

packages/visual-editor/src/components/puck/Events.tsx
colton-demetriou: no, needed here to correctly format the title, data/time, and description

asanehisa:
Hmmm nit but spacing / padding in mocks looks a bit diff?

colton-demetriou:
> Hmmm nit but spacing / padding in mocks looks a bit diff? specifically for each card (above Event Title)

it looks diff on some of the mocks -- I'll go back to fix if needed

briantstephan (approved):

PR #338, opened 2025-04-07
fix: radio toggle label doesn't work in array fields

benlife5:
Since we added the entity field selector, clicking the labels of the radio toggle didn't work in array fields for some reason, which has become more relevant for the page section components that have an array of cards


asanehisa (approved):
Oh odd

mkilpatrick (approved):

PR #344, opened 2025-04-10
feat: update NearbyLocations to use new Section styling

jwartofsky-yext:
The cards are now all the same height and the corners appear rounded properly.

Latest:
Screenshot 2025-04-10 at 2 40 44 PM

### DIFF
--- a/packages/visual-editor/src/components/puck/NearbyLocations.tsx
+++ b/packages/visual-editor/src/components/puck/NearbyLocations.tsx
@@ -315,7 +315,7 @@ export const NearbyLocations: ComponentConfig<NearbyLocationsProps> = {
   fields: nearbyLocationsFields,
   defaultProps: {
     styles: {
-      backgroundColor: backgroundColors.background2.value,
+      backgroundColor: backgroundColors.background1.value,
       cardBackgroundColor: backgroundColors.background1.value,
     },
     heading: {
--- a/packages/visual-editor/src/components/puck/NearbyLocations.tsx
+++ b/packages/visual-editor/src/components/puck/NearbyLocations.tsx
@@ -238,6 +238,7 @@ const NearbyLocationsComponent: React.FC<NearbyLocationsProps> = (props) => {
     console.warn(
       "Missing YEXT_CONTENT_ENDPOINT! Unable to fetch nearby locations."
     );
+    return <></>;
   }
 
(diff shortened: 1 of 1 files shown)


benlife5:
I think you want to set the Section on line 273 to applyPageLevelStyles and not the one on 170. I think that might automatically apply padding but if not, add some padding to whatever holds the cards

benlife5 (dismissed):

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -228,7 +233,9 @@ const NearbyLocationsComponent: React.FC<NearbyLocationsProps> = (props) => {
   );
   const headingText = resolveYextEntityField<string>(document, heading.text);
 
-  const contentEndpoint: string = document?._env?.YEXT_CONTENT_ENDPOINT;
+  const contentEndpoint: string =
+    document?._env?.YEXT_CONTENT_ENDPOINT ||
+    "https://streams-dev.yext.com/v2/accounts/1000146856/content/visualEditorLocations?api_key=8cd8e7f4a761bc13f2da4cd980fbc1ec";
```
mkilpatrick, line 238: I don't think we should hardcode the fallback url nor put an api in the code

packages/visual-editor/src/components/puck/NearbyLocations.tsx
```diff
@@ -228,7 +233,9 @@ const NearbyLocationsComponent: React.FC<NearbyLocationsProps> = (props) => {
...
   if (!contentEndpoint) {
     console.warn(
       "Missing YEXT_CONTENT_ENDPOINT! Unable to fetch nearby locations."
```
mkilpatrick, line 241: So if this is the case we should probably just return nothing.

packages/visual-editor/src/components/puck/NearbyLocations.tsx
jwartofsky-yext: whoops, that wasn't meant to be there

Removed

packages/visual-editor/src/components/puck/NearbyLocations.tsx
jwartofsky-yext: Done

benlife5 (approved):

mkilpatrick (approved):

PR #360, opened 2025-04-11
fix: basic selector color doesn't update with theme

benlife5:
The theme editor updates the CSS variables for the preview iframe but not the editor page and the selector reads the variables from the editor page

Also converts Banner to use DARK_BACKGROUND_COLOR so that it's dropdown includes the color squares


packages/visual-editor/src/utils/applyTheme.ts
```diff
@@ -90,16 +90,21 @@ export const updateThemeInEditor = async (
   newTheme: ThemeData,
   themeConfig: ThemeConfig
 ) => {
-  devLogger.logFunc("updateThemeInEditor");
```
jwartofsky-yext, line 93: Do we not want devlogger anymore for this?

jwartofsky-yext (approved):

packages/visual-editor/src/utils/applyTheme.ts
benlife5: ah whoops

asanehisa (approved):

coltondemetriou (approved):

colton-demetriou (approved):

PR #361, opened 2025-04-11
fix: adjust constant cards format in collection

asanehisa:
If constantValueEnabled, then user can select between flex or grid for their cards. If flex is clicked, they can choose between flex-col and flex-row. This way person cards can work on grids and event cards can work on flex with flex-col

Else if entityfields are selected, works the same as always

https://github.com/user-attachments/assets/2e032401-ac81-4096-b4de-8b7b72be9d9d


jwartofsky-yext (approved):

benlife5 (approved):
Double check in interactive that this works for the live page, but LGTM

PR #362, opened 2025-04-11
feat: dynamically set mapbox static map size

jwartofsky-yext:
Updates mapbox static map size to fit the container

Inside of a Grid
image

Inside of a Flex
image

Inside of Section Container
image

### DIFF
--- a/packages/visual-editor/src/components/puck/MapboxStaticMap.tsx
+++ b/packages/visual-editor/src/components/puck/MapboxStaticMap.tsx
@@ -64,7 +64,7 @@ export function useGrandparentSize<T extends HTMLElement = HTMLElement>(): [
     height: DEFAULT_HEIGHT,
   });
 
-  React.useLayoutEffect(() => {
+  React.useEffect(() => {
     let node: HTMLElement | null = selfRef.current;
     if (!node) return;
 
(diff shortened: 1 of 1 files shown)


asanehisa:
Hmm I think ideally they wanted it to just "know" how big to be based on the component its in

packages/visual-editor/src/components/puck/MapboxStaticMap.tsx
```diff
@@ -70,6 +82,10 @@ const MapboxStaticMap = ({
     return <></>;
   }
 
+  if (!width || !height) {
```
benlife5, line 85: Yeah I think we want to keep the generated image at 1024x300 but have the image fit it's parent container via object-fit

jwartofsky-yext: If we keep the image at 1024x300 regardless of the parent size, then the image will end up getting stretched and looking strange.

I think we have to dynamically determine the width and height to send to the API so we get the correct size back.

Alternatively we could always get 1024x300 and trim it down to the correct size, but I'm not sure if that's better or not.

asanehisa (approved):
assuming it won't break on live site?

packages/visual-editor/src/components/puck/MapboxStaticMap.tsx
jwartofsky-yext: nvm,

object-cover is better

Updated

asanehisa (dismissed):

benlife5 (dismissed):

colton-demetriou (approved):

asanehisa (dismissed):

asanehisa (approved):

benlife5 (approved):

PR #366, opened 2025-04-13
feat: update mapbox size on various events

jwartofsky-yext:
Update the size of the window 3 times with a 1 second delay after the page loads.
    
Update the size of the window on load, resize, visibilitychange, or scroll.
    
Note that if the size of the parent window hasn't changed, there won't be an additional API call.


asanehisa (commented):
I think you can merge into main now?

packages/visual-editor/src/components/MapboxStaticMap.tsx
```diff
@@ -90,6 +99,30 @@ export function useGrandparentSize<T extends HTMLElement = HTMLElement>(): [
...
     });
+
+    return () => clearTimeout(timer); // Cleanup timeout on unmount or re-render
+  }, [apiCallCount, updateTrigger]);
+
+  const updateEvents: string[] = [
+    "resize",
+    "load",
+    "visibilitychange",
+    "scroll",
```
benlife5, line 110: Is "scroll" needed?

packages/visual-editor/src/components/MapboxStaticMap.tsx
```diff
@@ -90,6 +99,30 @@ export function useGrandparentSize<T extends HTMLElement = HTMLElement>(): [
...
+  ];
+  // Listen for window resize
+  React.useEffect(() => {
+    const handleResize = () => {
+      setUpdateTrigger((prev) => prev + 1); // Trigger re-run
```
benlife5, line 115: Does resize fire constantly throughout you dragging your window? I'm a little concerned that this method will cause the client's mapbox API to get hit dozens of times per page view

packages/visual-editor/src/components/MapboxStaticMap.tsx
jwartofsky-yext: removed

replaced with "focus" and "deviceorientation"

packages/visual-editor/src/components/MapboxStaticMap.tsx
jwartofsky-yext: Added a debounce so it will fire 1 second since the latest update

mkilpatrick (approved):

benlife5 (approved):

PR #367, opened 2025-04-14
chore: add wcag testing

benlife5:


packages/visual-editor/src/components/Address.test.tsx
```diff
@@ -0,0 +1,86 @@
+import * as React from "react";
```
jwartofsky-yext, line 1: Should we put all of these under their own folder src/components/tests?

jwartofsky-yext (approved):

asanehisa (commented):
Just double checking, in the actual starter, the theme.config won't change ? or will it be imported from "@yext/visual-editor"

packages/visual-editor/src/components/Address.test.tsx
benlife5: this way seemed to be the consensus: https://yext.slack.com/archives/C02985ZF96Y/p1744233023627159

benlife5:
> Just double checking, in the actual starter, the theme.config won't change ? or will it be imported from "@yext/visual-editor"

It doesn't need to change, but it could be imported from "@yext/visual-editor", which would mean we wouldn't have to update in multiple places each time

asanehisa (approved):

PR #371, opened 2025-04-15
fix: don't render sample text if field is mapped

colton-demetriou:
Ran locally and verified that the sample text is only rendered when no field is mapped. Otherwise it will be the mapped value, even if empty.

### DIFF
--- a/packages/visual-editor/src/components/TextList.tsx
+++ b/packages/visual-editor/src/components/TextList.tsx
@@ -26,14 +26,12 @@ const TextListComponent: React.FC<TextListProps> = ({
   list: textListField,
 }) => {
   const document = useDocument();
-  let resolvedTextList: any = resolveYextEntityField(document, textListField);
+  let resolvedTextList = resolveYextEntityField(document, textListField);
 
   // When constantValueEnabled is true but no constant values have been set yet, show defaults
   if (
     textListField.constantValueEnabled &&
-    (!textListField.constantValue ||
-      (Array.isArray(textListField.constantValue) &&
-        textListField.constantValue.length === 0))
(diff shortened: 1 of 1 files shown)


packages/visual-editor/src/components/TextList.tsx
```diff
@@ -27,9 +27,17 @@ const TextListComponent: React.FC<TextListProps> = ({
 }) => {
   const document = useDocument();
   let resolvedTextList: any = resolveYextEntityField(document, textListField);
```
benlife5, line 29: remove the `:any` typing here then I think
```
(!textListField.constantValue ||
      (Array.isArray(textListField.constantValue) &&
        textListField.constantValue.length === 0))
```
can become something like
`!textListField.constantValue?.length`

packages/visual-editor/src/components/TextList.tsx
```diff
@@ -39,13 +47,15 @@ const TextListComponent: React.FC<TextListProps> = ({
...
       constantValueEnabled={textListField.constantValueEnabled}
     >
-      <ul className="components list-disc list-inside text-body-fontSize font-body-fontFamily font-body-fontWeight">
-        {resolvedTextList.map((text: any, index: any) => (
-          <li key={index} className="mb-2">
-            {text}
-          </li>
-        ))}
-      </ul>
+      {resolvedTextList && resolvedTextList.length > 0 ? (
```
benlife5, line 50: resolvedTextList?.length

benlife5 (approved):

asanehisa (approved):

briantstephan (approved):

jwartofsky-yext (approved):

PR #377, opened 2025-04-17
fix: fixes outerHTML vulnerability

jwartofsky-yext:
ran `pnpm run test:editor && pnpm run test:components` successfully

### DIFF
--- a/packages/visual-editor/src/components/WCAG/WCAG.setup.ts
+++ b/packages/visual-editor/src/components/WCAG/WCAG.setup.ts
@@ -7,9 +7,16 @@ import "jest-axe/extend-expect";
 
 // Applies the theme variables
 beforeEach(() => {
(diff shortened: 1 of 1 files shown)


packages/visual-editor/src/components/WCAG/WCAG.setup.ts
```diff
@@ -8,9 +8,8 @@ import "jest-axe/extend-expect";
 // Applies the theme variables
 beforeEach(() => {
   const tag = document.createElement("style");
-  const style = applyTheme({}, defaultThemeConfig);
+  tag.textContent = applyTheme({}, defaultThemeConfig);
```
benlife5, line 11: applyTheme returns a bunch of link tags and a style tag so I don't think they can be nested under a style tag. You can check the output screenshots from running locally to see if the theme is being applied

PR #378, opened 2025-04-17
fix: cta undefined 

asanehisa:
`"" ?? "#"` is `""`
`"" || "#"` is `"#"`

`<Link/> `from pages-components requires link for CTAs and empty strings do not count. Fixed and removed unneeded code.


benlife5 (approved):
yay javascript

PR #381, opened 2025-04-18
feat: add liveVisibility prop to Promo

asanehisa:
On show, no changes from normal. On hide, in live page and preview it won't show up at all and in editor it has opacity-30 with an slashed eyeball icon floating.

Screenshot 2025-04-18 at 1 57 10 PM


jwartofsky-yext (dismissed):

colton-demetriou (dismissed):
Nice

PR #391, opened 2025-04-22
fix: handle ComplexImageType in image atom

asanehisa:


jwartofsky-yext (approved):
👍

packages/visual-editor/src/components/atoms/image.tsx
```diff
@@ -41,20 +58,14 @@ export const Image: React.FC<ImageProps> = ({
...
-    typeof resolvedImage === "object" &&
-    "image" in resolvedImage
-  ) {
-    image = resolvedImage.image as ImageType;
-  } else if (resolvedImage) {
-    image = resolvedImage;
-  } else {
-    image = { height: 150, width: 150, url: "" };
+// if image is type YextEntityField<ImageType> | YextEntityField<ComplexImageType>, return the resolvedImage
+const handleUnresolvedImage = (
```
benlife5, line 62: Sorry I'm a little confused... so the issue was that sometimes unresolved images are being passed `<ImageAtom>`? I'm fine with this if it works for both regular and complex images but I'm not sure where those unresolved images would be coming from

asanehisa: Oh no, that wasn't happening. I just added this so the image atom could support more but maybe i won't since image atom shouldn't ever be directly used in that way.

benlife5: ah gotcha. Ok cool, I'm good with whatever you think is best for the prop type

asanehisa: Yaa i think this rn works (unless we see more use cases in future and we wanna change)

PR #394, opened 2025-04-23
fix: fix Collection "grid" mode applying in entity mode

jwartofsky-yext:
When the Collections component is set to "Entity" instead of Constant Value, we hide the options for Grid and Flex.

However they were still being applied.

We only want the Flex behavior while in entity mode.

Using Grid layout in entity mode causes the cards to all get squished.


benlife5:
LGTM. Can you also add Flex and Horizontal as default props since we default to constant value?

benlife5 (approved):

PR #395, opened 2025-04-24
fix: disallow all pointer events and hover states in Theme Editor

asanehisa:
https://jam.dev/c/5e238b4d-ec65-432a-9692-b8a5adb6fe3c


benlife5:
I think the theme editor got set to interactive mode because there was an item to eliminate puck component selection / the blue overlays. I'm wondering if it would be possible to keep the overlays so that components aren't clickable but override the CSS so they're transparent

benlife5:
Double check with Aaron: the item implies that all interactivity including hover states should be blocked, but I personally think it's nice to have stuff like hover states since they're affected by the theme as long as nothing can break the editor

mkilpatrick (approved):

benlife5 (approved):

PR #399, opened 2025-04-24
fix: apply gap inside of grid columns

jwartofsky-yext:
The gap was only being applied to the space between columns.

According to the item, it seems that we expect to have the gap also applied between rows.

Unsure if we want this fix or not. Waiting on feedback.


jwartofsky-yext:
We want this fix instead: https://github.com/yext/visual-editor/pull/406

PR #400, opened 2025-04-24
fix: adjust YextEntityFieldSelector

asanehisa:
When I was playing around with PhotoGallery, I noticed puck's array field (available when const value is toggled) wasn't updating the ui properly. Now fixed.


benlife5 (approved):
Interesting, nice

PR #409, opened 2025-04-28
feat: add directory template

asanehisa:


packages/visual-editor/src/vite-plugin/templates/directory.tsx
```diff
@@ -0,0 +1,104 @@
+import "@yext/visual-editor/style.css";
```
mkilpatrick, line 1: Update the /starter files

asanehisa: ? the /starter only has one template which is dev.tsx

mkilpatrick: Oic. Are we able to test the Directory components in it? (aka is there data in the localData files for it)

asanehisa: Yup there is

mkilpatrick (approved):

jwartofsky-yext (approved):
lgtm

PR #413, opened 2025-04-30
feat: apply headerScript from document

jwartofsky-yext:
Adds a new function "applyHeaderScript" which applies the headerScript from the document.

The headerScript can be set in the platform under "basic settings".


packages/visual-editor/src/utils/applyHeaderScript.ts
```diff
@@ -0,0 +1,13 @@
...
+  if (!document?.__?.visualEditorConfig) {
+    return;
+  }
+
+  const headerScript: string = JSON.parse(
+    document.__.visualEditorConfig
+  )?.headerScript;
+
+  if (headerScript) {
+    return `<script>${headerScript}</script>`;
```
mkilpatrick, line 11: Don't include `<script>`. The user needs to do this in Storm. Just return `headerScript` directly.

packages/visual-editor/src/vite-plugin/templates/main.tsx
```diff
@@ -65,6 +66,7 @@ export const getHeadConfig: GetHeadConfig<TemplateRenderProps> = ({
     ],
     other: [
       applyAnalytics(document),
+      applyHeaderScript(document),
```
mkilpatrick, line 69: Add this to starter's dev.tsx too (and applyAnalytics which we missed)

packages/visual-editor/src/utils/applyHeaderScript.ts
```diff
@@ -0,0 +1,7 @@
+export const applyHeaderScript = (document: Record<string, any>) => {
+  if (!document?.__?.visualEditorConfig) {
+    return;
+  }
+
+  return JSON.parse(document.__.visualEditorConfig)?.headerScript;
```
mkilpatrick, line 6: This will fail if headerScript is undefined

packages/visual-editor/src/vite-plugin/templates/main.tsx
mkilpatrick: bump

asanehisa (dismissed):
we'll def wanna test this out on actual starter site too

mkilpatrick (approved):

asanehisa (approved):

PR #414, opened 2025-04-30
feat: add TeamSection component

asanehisa:
Screenshot 2025-05-08 at 4 28 50 PM

fixed label on Data to Data instead of Section Heading lol

### DIFF
--- a/packages/visual-editor/src/components/pageSections/TeamSection.tsx
+++ b/packages/visual-editor/src/components/pageSections/TeamSection.tsx
@@ -14,16 +14,17 @@ import {
   CTA,
   backgroundColors,
   VisibilityWrapper,
+  PhoneAtom,
 } from "@yext/visual-editor";
-import { ComplexImageType, CTA as CTAType } from "@yext/pages-components";
+import { CTA as CTAType, ImageType } from "@yext/pages-components";
 import { ComponentConfig, Fields } from "@measured/puck";
-import { FaEnvelope, FaPhone } from "react-icons/fa";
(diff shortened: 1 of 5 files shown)


packages/visual-editor/src/components/pageSections/TeamSection.tsx
```diff
@@ -0,0 +1,244 @@
...
+      >
+        <div className="flex flex-col gap-4">
+          {person.phoneNumber && (
+            <div className="flex items-center gap-2">
+              <div className="w-6 h-6 rounded-full bg-palette-primary-light flex items-center justify-center">
+                <FaPhone className="w-3 h-3 text-black" />
+              </div>
+              <CTA
+                link={person.phoneNumber}
+                label={person.phoneNumber}
```
benlife5, line 132: Does this need to be formatted?

asanehisa: Ooh good catch, seems like we didnt for PeopleCards (but maybe that was before we added the PhoneAtom)

packages/visual-editor/src/internal/utils/getFilteredEntityFields.ts
```diff
@@ -43,6 +43,7 @@ export type EntityFieldTypes =
   | "type.coordinate"
   | "type.cta"
   | "type.rich_text_v2"
+  | "type.people"
```
mkilpatrick, line 46: type.team_section

mkilpatrick (approved):

benlife5 (approved):

PR #416, opened 2025-05-01
feat: add TestimonialSection component

asanehisa:


packages/visual-editor/src/internal/utils/getFilteredEntityFields.ts
```diff
@@ -42,6 +42,7 @@ export type EntityFieldTypes =
   | "type.phone"
   | "type.coordinate"
   | "type.cta"
+  | "type.testimonial"
```
mkilpatrick, line 45: type.testimonials_section

mkilpatrick (approved):

PR #417, opened 2025-05-01
feat: remove content, layout, card categories

asanehisa:
Screenshot 2025-05-01 at 10 50 23 AM

### DIFF
--- a/starter/src/ve.config.tsx
+++ b/starter/src/ve.config.tsx
@@ -5,16 +5,7 @@ import {
   PageSectionCategory,
   PageSectionCategoryComponents,
   PageSectionCategoryProps,
-  LayoutBlockCategory,
-  CardCategory,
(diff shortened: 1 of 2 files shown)


starter/src/ve.config.tsx
```diff
@@ -68,11 +41,6 @@ export const mainConfig: Config<MainProps> = {
       return (
         <DropZone
           zone="default-zone"
-          disallow={[
-            ...ContentBlockCategory,
```
jwartofsky-yext, line 72: Is this removing the restrictions we previously had for these components?

Do we want to allow these components to be placed directly in the root now?

asanehisa: they're just being removed completely from components available

jwartofsky-yext (approved):

PR #423, opened 2025-05-02
feat: support fullstory in editor iframe

benlife5:
Adds fullstory capture to the edit page

Testing: Created a branch with a self-packed version of VE. Swapped to that branch on a prod account that is not excluded from reporting. Logged out of admin2 and then into the platform as a user created from the email signup flow. After waiting at least 30 minutes for my fullstory session to populate, checked with Duval (who has access to fullstory) and confirmed the iframed content was visible)

### DIFF
--- a/packages/visual-editor/src/vite-plugin/templates/edit.tsx
+++ b/packages/visual-editor/src/vite-plugin/templates/edit.tsx
@@ -31,7 +31,7 @@ export const getHeadConfig: GetHeadConfig<TemplateRenderProps> = ({
   document,
 }): HeadConfig => {
   return {
-    other: applyTheme(document, themeConfig),
+    other: fullStorySnippet + applyTheme(document, themeConfig),
   };
 };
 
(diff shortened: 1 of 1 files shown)


jwartofsky-yext:
Can you provide some testing details?

packages/visual-editor/src/vite-plugin/templates/edit.tsx
```diff
@@ -58,3 +58,27 @@ const Edit: () => JSX.Element = () => {
...
 export default Edit;
+
+const fullStorySnippet = `<script>
+window['_fs_run_in_iframe'] = true;
+window['_fs_debug'] = false;
+window['_fs_host'] = 'fullstory.com';
+window['_fs_script'] = 'edge.fullstory.com/s/fs.js';
+window['_fs_org'] = 'o-1RWKDJ-na1';
+window['_fs_namespace'] = 'FS';
+!function(m,n,e,t,l,o,g,y){var s,f,a=function(h){
```
mkilpatrick, line 69: Is all of this necessary? I thought we only needed `_fs_run_in_iframe` and maybe the org.

benlife5: > to capture the iframed content, add this special flag to the Fullstory data capture snippet and install the modified snippet on the page that is hosted within the iframe

Yeah unfortunately I think you have to have the full snippet

mkilpatrick: Crazy

mkilpatrick (approved):

asanehisa (approved):

PR #425, opened 2025-05-03
POC: i18n support

mkilpatrick:


jwartofsky-yext:
ik this is a POC, but could you add a description of what exactly this does?

I see it's using i18n to translate the string "Test"

Then it looks like there's also a constant value mapping

PR #429, opened 2025-05-05
fix: generate templates in plugin for dev server

benlife5:
Runs the generate templates stage after writing the VE templates so that the edit static page is including in features.json. Alternative solution to https://github.com/yext/pages/pull/579.

The dev server should be run in the starter with the `--noGenFeatures` flag because the initial generate is run async and can complete after this one.

### DIFF
--- a/packages/visual-editor/src/vite-plugin/plugin.ts
+++ b/packages/visual-editor/src/vite-plugin/plugin.ts
@@ -63,7 +63,6 @@ export const yextVisualEditorPlugin = (): Plugin => {
         return;
       }
       generateFiles();
-
       server.httpServer?.on("close", () => {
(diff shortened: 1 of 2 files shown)


asanehisa (approved):

jwartofsky-yext:
> Runs the generate templates stage after writing the VE templates so that the edit static page is including in features.json. Alternative solution to [yext/pages#579](https://github.com/yext/pages/pull/579).
> 
> The dev server should be run in the starter with the `--noGenFeatures` flag because the initial generate is run async and can complete after this one.

Can you add the flag to the "dev" command in starter/package.json?

benlife5:
Solved by https://github.com/yext/pages/pull/582

PR #430, opened 2025-05-07
feat(dev): use new content endpoints

jwartofsky-yext:
Uses the new content endpoints

Dependent on the other work to be merged first

Also not tested yet since dev is down

### DIFF
--- a/packages/visual-editor/src/components/pageSections/NearbyLocations.tsx
+++ b/packages/visual-editor/src/components/pageSections/NearbyLocations.tsx
@@ -243,15 +243,35 @@ const NearbyLocationsComponent: React.FC<NearbyLocationsSectionProps> = (
   );
   const headingText = resolveYextEntityField<string>(document, heading.text);
 
-  const contentEndpoint: string = document?._env?.YEXT_CONTENT_ENDPOINT;
-  if (!contentEndpoint) {
+  const apiKey: string = document?._env?.YEXT_VISUAL_EDITOR_APP_API_KEY;
+  if (!apiKey) {
     console.warn(
-      "Missing YEXT_CONTENT_ENDPOINT! Unable to fetch nearby locations."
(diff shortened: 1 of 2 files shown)


mkilpatrick (approved):

packages/visual-editor/src/components/pageSections/NearbyLocations.tsx
```diff
@@ -243,41 +243,65 @@ const NearbyLocationsComponent: React.FC<NearbyLocationsSectionProps> = (
   );
   const headingText = resolveYextEntityField<string>(document, heading.text);
 
-  const contentEndpoint: string = document?._env?.YEXT_CONTENT_ENDPOINT;
```
benlife5, line 246: Do we need to continuing supporting the old key because sites created previously will not have the new one set?

packages/visual-editor/src/utils/api/nearbyLocations.tsx
```diff
@@ -36,3 +42,33 @@ export const fetchNearbyLocations = async ({
 
   return await response.json();
 };
+
+function getContentEndpointUrl(
```
mkilpatrick, line 46: Noting that this will change since Spruce is going to add the endpoint to the document.

packages/visual-editor/src/components/pageSections/NearbyLocations.tsx
jwartofsky-yext: The only case where things could end up broken is if someone makes a new VE site after all these changes and then downgrades their VE version

Old sites will still have the old endpoints and keys set so they will continue to work

The new endpoints will be backfilled for all pagesets so all sites will work when upgraded

mkilpatrick: This is fine

jwartofsky-yext:
Replaced with https://github.com/yext/visual-editor/pull/463

PR #432, opened 2025-05-08
feat: add YextStructFieldSelector and update hero

asanehisa:
https://jam.dev/c/0646a03b-d27f-4c8a-89bd-7ea44559d8ce

Made a new YextStructFieldSelector for now rather than attempt to make it with YextEntityFieldSelector. This new functionality with constValue overrides was requested at a recent meeting but would like to wait for us to see if we like it before cleaning up YextEntityFieldSelector and merging them (if we merge).

For HeroSection we wanted 3 main fields, data, styles, and liveVisibility. We also purposefully removed the showHide fields for the ctas and hours.

### DIFF
--- a/packages/visual-editor/src/components/pageSections/HeroSection.tsx
+++ b/packages/visual-editor/src/components/pageSections/HeroSection.tsx
@@ -21,6 +21,7 @@ import {
   YextField,
   VisibilityWrapper,
   CTAProps,
+  resolveYextEntityFieldV2,
 } from "@yext/visual-editor";
 
 const PLACEHOLDER_IMAGE_URL = "https://placehold.co/640x360";
@@ -191,7 +192,7 @@ const HeroSectionWrapper = ({
     document,
     hours.entityField
   );
-  const resolvedHero = resolveYextEntityField(document, hero);
+  const resolvedHero = resolveYextEntityFieldV2(document, hero);
   console.log("resolvedHero", resolvedHero);
 
   const { timezone } = document as {
--- a/packages/visual-editor/src/components/pageSections/HeroSection.tsx
+++ b/packages/visual-editor/src/components/pageSections/HeroSection.tsx
@@ -2,6 +2,10 @@ import * as React from "react";
 import { ComponentConfig, Fields } from "@measured/puck";
 import { HoursStatus, HoursType } from "@yext/pages-components";
 import {
+  YextSelectorV2,
+  YextEntityFieldV2,
+} from "../../editor/UpdatedYextSelector.tsx";
+import {
   HeroSectionType,
   useDocument,
   resolveYextEntityField,
@@ -34,7 +38,7 @@ export interface HeroSectionProps {
     entityField: YextEntityField<HoursType>;
     showHours: boolean;
   };
-  hero: YextEntityField<HeroSectionType>;
+  hero: YextEntityFieldV2<HeroSectionType>;
   primaryCTA: {
     showCTA: boolean;
     variant: CTAProps["variant"];
@@ -101,10 +105,10 @@ const heroSectionFields: Fields<HeroSectionProps> = {
       }),
     },
   }),
-  hero: YextField("Hero", {
-    type: "entityField",
+  hero: YextSelectorV2({
+    label: "Hero",
     filter: {
-      types: ["type.hero_section"],
+      types: ["c_hero"],
     },
   }),
   primaryCTA: YextField("Primary CTA", {
@@ -188,6 +192,7 @@ const HeroSectionWrapper = ({
     hours.entityField
   );
   const resolvedHero = resolveYextEntityField(document, hero);
+  console.log("resolvedHero", resolvedHero);
 
   const { timezone } = document as {
     timezone: string;
--- a/packages/visual-editor/src/components/pageSections/HeroSection.tsx
+++ b/packages/visual-editor/src/components/pageSections/HeroSection.tsx
@@ -19,7 +19,7 @@ import {
   CTAProps,
   resolveYextStructField,
   YextStructFieldSelector,
-  YextEntityFieldV2,
+  YextStructEntityField,
 } from "@yext/visual-editor";
 
 const PLACEHOLDER_IMAGE_URL = "https://placehold.co/640x360";
@@ -29,7 +29,7 @@ export interface HeroSectionProps {
     businessName: YextEntityField<string>;
     localGeoModifier: YextEntityField<string>;
     hours: YextEntityField<HoursType>;
-    hero: YextEntityFieldV2<HeroSectionType>;
+    hero: YextStructEntityField<HeroSectionType>;
   };
   styles: {
     backgroundColor?: BackgroundStyle;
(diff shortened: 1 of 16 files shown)


packages/visual-editor/src/editor/YextSelectorV2.tsx
```diff
@@ -0,0 +1,138 @@
...
+const SubfieldsInput = ({ filter, onChange, value }: InputPropsV2) => {
+  const subfields = getSubfieldsFromType(filter.type);
+  if (!subfields) {
+    return;
+  }
+
+  return (
+    <FieldLabel
+      label="Subfield Assignments"
+      className="ve-inline-block ve-w-full pt-4"
```
benlife5, line 81: ve-pt-4

packages/visual-editor/src/editor/YextSelectorV2.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+type EntityFieldTypesFilter = {
+  type: EntityFieldTypesV2;
+};
+
+// EntityFieldTypesV2 contains new struct types that we support overriding individual subfields for
+export type EntityFieldTypesV2 = "type.hero_section";
+
+// YextEntityFieldV2 keeps track of which fields we are allowing individual overriding for using constantValueOverride
+export type YextEntityFieldV2<T extends Record<string, any> = any> = {
```
benlife5, line 21: If we don't end up integrating YextSelectorV2 in YextEntityFieldSelector, maybe it's called YextStructFieldSelector or similar. I'm not opposed to keeping them separate as long as we can clearly explain their different use cases

packages/visual-editor/src/components/pageSections/HeroSection.tsx
```diff
@@ -17,125 +17,58 @@ import {
...
-        options: [
-          { label: "Show", value: true },
-          { label: "Hide", value: false },
-        ],
-      }),
-      variant: YextField("Button Variant", {
-        type: "radio",
-        options: "CTA_VARIANT",
+      hero: YextSelectorV2({
+        label: "Hero",
```
benlife5, line 68: More of a product thing but I think this is still a bit of an awkward label. At least to me it's not particularly self-evident why these fields are different than the ones above

benlife5 (commented):
I really like this! It seems to be laid out in a very intuitive way and easy to use

packages/visual-editor/src/components/pageSections/HeroSection.tsx
asanehisa: Oh yeah I think I'd add it into YextField at some point (maybe now)

benlife5: Screenshot 2025-05-08 at 10 56 36 AM

Oh sorry I meant like in the editor UI, what does "Hero" as a label mean and why is business name not part of that but the image is

packages/visual-editor/src/utils/resolveYextEntityField.ts
```diff
@@ -42,6 +43,25 @@ export const resolveYextEntityField = <T>(
   return undefined;
 };
 
+export const resolveYextEntityFieldV2 = <T extends Record<string, any>>(
```
jwartofsky-yext, line 46: Can you add some comments explaining the difference between the V2 versions and the V1

It looks like this just uses resolveYextEntityField and adds support for the overriddenValues.

Is it possible to just put the overriddenValues stuff in resolveYextEntityField and use that instead?

packages/visual-editor/src/editor/YextSelectorV2.tsx
```diff
@@ -0,0 +1,138 @@
...
+import "./index.css";
+
+type RenderProps = Parameters<CustomField<any>["render"]>[0];
+
+type EntityFieldTypesFilter = {
+  type: EntityFieldTypesV2;
+};
+
+// EntityFieldTypesV2 contains new struct types that we support overriding individual subfields for
+export type EntityFieldTypesV2 = "type.hero_section";
```
jwartofsky-yext, line 18: Shouldn't this be an array?

asanehisa: it'll be just like 

`export type EntityFieldTypesV2 = "type.hero_section" | "type.promo_section" | "type.blah" `;

asanehisa: Hmmm so rn it is just a string that defines specific string options that are allowed. I don't quite understand the idea behind an array?

jwartofsky-yext: I thought EntityFieldTypes V1 was an array, but it looks like it's also a set of types like this is

My mistake

Your implementation is fine

mkilpatrick: You wouldn't be able to have different types for this usage. It would be pinned to a specific field type, so single option makes sense to me.

packages/visual-editor/src/editor/YextSelectorV2.tsx
mkilpatrick: We can keep it separate for now (though I agree with the name change). We can refactor the two things together later if we want to.

packages/visual-editor/src/editor/YextSelectorV2.tsx
```diff
@@ -0,0 +1,138 @@
...
+// inferred entityValue (from field) being used or use a constantValue.
+const SubfieldsInput = ({ filter, onChange, value }: InputPropsV2) => {
+  const subfields = getSubfieldsFromType(filter.type);
+  if (!subfields) {
+    return;
+  }
+
+  return (
+    <FieldLabel
+      label="Subfield Assignments"
```
mkilpatrick, line 80: "Content Overrides" for now per the thread

packages/visual-editor/src/internal/puck/Subfields.ts
```diff
@@ -0,0 +1,35 @@
...
+}[];
+
+const HERO_SECTION_SUBFIELD: SubFieldProps = [
+  { field: "image", type: "type.image", label: "Image" },
+  { field: "primaryCta", type: "type.cta", label: "Primary CTA" },
+  { field: "secondaryCta", type: "type.cta", label: "Secondary CTA" },
+];
+
+const STRUCT_TYPE_TO_SUBFIELDS: Record<EntityFieldTypesV2, SubFieldProps> = {
+  "type.hero_section": HERO_SECTION_SUBFIELD,
```
mkilpatrick, line 20: Nice and extensible 👍

asanehisa: yay!

packages/visual-editor/src/editor/YextStructFieldSelector.tsx
```diff
@@ -31,8 +31,8 @@ export type SelectorPropsV2 = {
   filter: EntityFieldTypesFilter;
 };
 
-// YextSelectorV2 will be used for new built-in struct and list field types.
-export const YextSelectorV2 = <U extends Record<string, any>>(
+// YextStructFieldSelector will be used for new built-in struct and list field types.
```
mkilpatrick, line 34: Rename YextEntityFieldV2 and any other new things you added that are named like "V2"

PR #433, opened 2025-05-08
feat: create locator template

k-gerner:
Creates the template for locator page sets. This is very similar to the directory template, however it differs in that it uses the search analytics config rather than the pages component, and imports the `locatorConfig` from `ve.config` instead of the `directoryConfig`.

Additionally, we still need to make the necessary update to the starter repo's `ve.config` file, as well as updates to this repo to pull the experience key off of the entity document.

J=[WAT-4782](https://yexttest.atlassian.net/browse/WAT-4782)
TEST=none


packages/visual-editor/src/utils/searchHeadlessConfig.ts
```diff
@@ -58,6 +60,52 @@ export const createSearchHeadlessConfig = (document: any) => {
...
+  }
+  const environment = document?._env?.YEXT_ENVIRONMENT;
+  if (!environment || !isValidEnvironment(environment)) {
+    warnings.push(
+      "Invalid or missing YEXT_ENVIRONMENT! Unable to set up locator analytics."
+    );
+  }
+  const cloudRegion = document?._env?.YEXT_CLOUD_REGION;
+  if (!cloudRegion || !isValidCloudRegion(cloudRegion)) {
+    warnings.push(
```
mkilpatrick, line 81: Why not just console warn directly instead of adding to a list and then iterating over it?

packages/visual-editor/src/vite-plugin/templates/locator.tsx
```diff
@@ -0,0 +1,145 @@
...
+  //   experienceKey: "jacob-test",
+  //   experienceVersion: EXPERIENCE_VERSION,
+  //   region: 'US',
+  //   env: 'PRODUCTION',
+  // }
+  // const searcher = provideHeadless(config);
+
+  return (
+    <SearchHeadlessProvider searcher={searcher}>
+      <AnalyticsProvider {...searchAnalyticsConfig}>
```
mkilpatrick, line 133: Does this also end up recording a pageview?

packages/visual-editor/src/utils/searchHeadlessConfig.ts
```diff
@@ -58,6 +64,56 @@ export const createSearchHeadlessConfig = (document: any) => {
...
+    warnings.forEach((msg) => console.warn(msg));
+    return;
+  }
+
+  // from @yext/analytics EnvironmentEnum
+  const analyticsEnvironment =
+    environment === Environment.SANDBOX ? "SANDBOX" : "PRODUCTION";
+  // from @yext/analytics RegionEnum
+  const analyticsRegion = cloudRegion.toUpperCase();
+  // experienceKey will eventually be on the entity document, waiting on Spruce
```
vijay267, line 103: Spruce already did this change, so anything else we're waiting on here?

packages/visual-editor/src/utils/searchHeadlessConfig.ts
k-gerner: So that we can check the length of it to determine if we should return `undefined` rather than a long if statement or separate booleans

packages/visual-editor/src/vite-plugin/templates/locator.tsx
k-gerner: Out of the box, no. However a lot of search analytics is in flux right now so maybe we could add support for it later

packages/visual-editor/src/utils/searchHeadlessConfig.ts
```diff
@@ -19,7 +25,7 @@ export const createSearchHeadlessConfig = (document: any) => {
   if (!mapboxApiKey) {
     warnings.push("Missing YEXT_MAPBOX_API_KEY! Unable to set up locator.");
   }
-  const cloudRegion = document?._env?.YEXT_CLOUD_REGION;
+  const cloudRegion = document?._env?.YEXT_CLOUD_REGION?.toLowerCase();
```
jwartofsky-yext, line 28: is "toLowerCase()" neccessary here?

packages/visual-editor/src/utils/searchHeadlessConfig.ts
```diff
@@ -31,7 +37,7 @@ export const createSearchHeadlessConfig = (document: any) => {
       "Invalid or missing YEXT_CLOUD_CHOICE! Unable to set up locator."
     );
   }
-  const environment = document?._env?.YEXT_ENVIRONMENT;
+  const environment = document?._env?.YEXT_ENVIRONMENT?.toLowerCase();
```
jwartofsky-yext, line 40: same

Shouldn't the env always be lowercase?

packages/visual-editor/src/utils/searchHeadlessConfig.ts
```diff
@@ -58,6 +64,56 @@ export const createSearchHeadlessConfig = (document: any) => {
...
+    // document may not exist while loading the page - no need to warn
+    return;
+  }
+  const warnings = [];
+  const businessId = document?.businessId;
+  if (!businessId) {
+    warnings.push("Missing businessId! Unable to set up locator analytics.");
+  }
+  const environment = document?._env?.YEXT_ENVIRONMENT?.toLowerCase();
+  if (!environment || !isValidEnvironment(environment)) {
```
jwartofsky-yext, line 82: isValidEnvironment should just return false if environment is null and then you won't need the first nullcheck

packages/visual-editor/src/utils/searchHeadlessConfig.ts
```diff
@@ -65,18 +69,22 @@ export const createSearchHeadlessConfig = (document: any) => {
  * @param document the entity document
  */
 export const createSearchAnalyticsConfig = (document: any) => {
+  if (!document) {
```
mkilpatrick, line 72: This is dubious. I saw the error with Kyle that the document is not found but there's probably a different reason.

packages/visual-editor/src/utils/searchHeadlessConfig.ts
k-gerner: Yes because the region etc used in `search-core` uses the lowercase version while the version we get from alpha code is uppercase

packages/visual-editor/src/utils/searchHeadlessConfig.ts
k-gerner: Do you mind if I leave it in for now then?

packages/visual-editor/src/utils/searchHeadlessConfig.ts
k-gerner: The env from `src/com/yext/config/Env.java` which we access via `YextConfig.YEXT_SITE.getEnv().name()` in our code gives the uppercase enums. This takes in lowercase.

An argument could be made as to where the lowercasing logic should live (alpha code or VE code) but since the code in alpha was already checked in I figured I'd just do it here

jwartofsky-yext (approved):

PR #443, opened 2025-05-12
feat: insight section const value support

colton-demetriou:
Screenshot 2025-05-12 at 3 25 56 PM

### DIFF
--- a/packages/visual-editor/src/components/pageSections/InsightSection.tsx
+++ b/packages/visual-editor/src/components/pageSections/InsightSection.tsx
@@ -17,9 +17,9 @@ import {
   InsightSectionType,
   InsightStruct,
   Timestamp,
+  MaybeRTF,
 } from "@yext/visual-editor";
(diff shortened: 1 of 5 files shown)


packages/visual-editor/src/components/pageSections/InsightSection.tsx
```diff
@@ -109,11 +109,9 @@ const InsightCard = ({
             </div>
           )}
           {insight.name && <Heading level={3}>{insight.name}</Heading>}
-          {insight.description?.json && (
+          {insight.description && (
             <Body>
```
benlife5, line 113: MaybeRTF includes it's own Body and undefined check so it can be used directly

benlife5 (approved):

briantstephan (approved):

asanehisa (commented):
ah late but can adjust InsightSection in the registry.tsx since it is using an additional atom now?

PR #447, opened 2025-05-13
feat: add const support for faqs section

colton-demetriou:
Screenshot 2025-05-13 at 12 22 34 PM

### DIFF
--- a/packages/visual-editor/src/components/pageSections/FAQsSection.tsx
+++ b/packages/visual-editor/src/components/pageSections/FAQsSection.tsx
@@ -14,6 +14,7 @@ import {
   VisibilityWrapper,
   EntityField,
   FAQSectionType,
+  MaybeRTF,
 } from "@yext/visual-editor";
 import {
   Accordion,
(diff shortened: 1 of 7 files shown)


packages/visual-editor/src/components/pageSections/FAQsSection.tsx
```diff
@@ -148,7 +140,6 @@ export const FAQSection: ComponentConfig<FAQSectionProps> = {
         constantValue: {
           faqs: [],
```
asanehisa, line 141: I think we might want the constValue faqs to have values now (so if they switch to constValue it isn't blank). Like mocks here: https://www.figma.com/design/sX3bkzkEpQ3g2mJrfSSdmd/Quickstart-Template--2025-?node-id=4787-28266&t=rGaJ8Tx6TsXvnxFv-0

colton-demetriou: I was doing that before but seemed like it was no longer desired -- if it is i'll go back and update all the components I've worked on in a separate pr to have default values

benlife5 (approved):

PR #448, opened 2025-05-13
feat: support const values for team section

colton-demetriou:
Screenshot 2025-05-13 at 1 35 15 PM

### DIFF
--- a/packages/visual-editor/src/internal/puck/constant-value-fields/TeamSection.tsx
+++ b/packages/visual-editor/src/internal/puck/constant-value-fields/TeamSection.tsx
@@ -41,14 +41,6 @@ const PersonStructArrayField: ArrayField<PersonStruct[]> = {
           label: "URL",
           type: "text",
         },
-        width: {
-          label: "Width",
-          type: "number",
-        },
-        height: {
-          label: "Height",
-          type: "number",
-        },
       },
     },
     name: {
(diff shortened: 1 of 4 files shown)


packages/visual-editor/src/internal/puck/constant-value-fields/TeamSection.tsx
```diff
@@ -0,0 +1,73 @@
...
+  arrayFields: {
+    headshot: {
+      type: "object",
+      label: "Headshot",
+      objectFields: {
+        url: {
+          label: "URL",
+          type: "text",
+        },
+        width: {
```
benlife5, line 44: I think we just want URL when it's an image constant value, right @asanehisa?

asanehisa: Ah yeah no width / height. Just use image and keep to aspect-ratio of the image

asanehisa: see PromoSection or EventSection

packages/visual-editor/src/internal/puck/constant-value-fields/TeamSection.tsx
```diff
@@ -0,0 +1,65 @@
...
+            onChange({ people: newValue }, uiState)
+          }
+        />
+      </div>
+    );
+  },
+};
+
+const PersonStructArrayField: ArrayField<PersonStruct[]> = {
+  label: "Array Field",
```
asanehisa, line 33: I don't know if u want this labeled Array Field

colton-demetriou: it's called that in EventSection as well, should be fine

benlife5 (approved):

PR #449, opened 2025-05-13
feat: add const value support for testimonials section

colton-demetriou:
Screenshot 2025-05-13 at 2 11 49 PM

### DIFF
--- a/packages/visual-editor/src/components/atoms/timestamp.tsx
+++ b/packages/visual-editor/src/components/atoms/timestamp.tsx
@@ -2,6 +2,7 @@ const format1: Intl.DateTimeFormatOptions = {
   month: "short",
   day: "numeric",
   year: "numeric",
+  timeZone: "UTC",
 };
 
 const format2: Intl.DateTimeFormatOptions = {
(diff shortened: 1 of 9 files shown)


benlife5 (approved):

packages/visual-editor/src/internal/puck/constant-value-fields/TestimonialSection.tsx
```diff
@@ -0,0 +1,58 @@
...
+    description: {
+      type: "textarea",
+      label: "Description",
+    },
+    contributorName: {
+      type: "text",
+      label: "Contributor Name",
+    },
+    contributionDate: {
+      type: "text",
```
asanehisa, line 52: Oh this is dateTime format I think?

asanehisa: like no need time but it is processed that way in Timestamp?

colton-demetriou: should i just make it DateTimeSelector?

benlife5: I'd say to make a new DateSelector. I think it can just be DateTimeSelector with input type="date"

asanehisa: Oh don't think u quite did what Ben mentioned (and I agree w/)

packages/visual-editor/src/components/atoms/timestamp.tsx
```diff
@@ -98,8 +99,15 @@ export function Timestamp({
 }: TimestampProps): JSX.Element {
   let timestamp;
   try {
-    const startDate = new Date(date);
-    const formattedEndDate = new Date(endDate);
+    // For date-only strings (YYYY-MM-DD), create the date in UTC
+    const startDate =
+      date.length === 10 ? new Date(date + "T12:00:00.000Z") : new Date(date);
```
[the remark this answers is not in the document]
colton-demetriou: doesn't really feel like that adds much value, shouldn't need to do this anywhere else

asanehisa (approved):

benlife5 (approved):

PR #450, opened 2025-05-13
feat: merge demo changes into main locator branch

k-gerner:
Main changes in here include:
- export locator components correctly
- move search / analytics providers into locator component directly
- remove locator components from starter config, since they won't work with location entities
- correctly extract experience key from entity document


starter/src/components/Locator.tsx
Fondryext, line 1: Do we wanna keep this around or just delete it outright?

k-gerner: Fair, I'll delete

Fondryext (approved):

PR #452, opened 2025-05-13
feat: data migrations

benlife5:
Adds a `migrate` function that can transform layout data. Applied in the templates and editor.

Demo video:
The first tab is running with the current components. The second tab has changes expressed by following migration:
```
  {
    CoreInfoSection: {
      action: "removed",
    },
    BannerSection: {
      action: "renamed",
      newName: "BasicBanner",
    },
    HeroSection: {
      action: "updated",
      propTransformation: ({
        data: { businessName, ...otherData },
        ...props
      }) => ({ ...props, data: { ...otherData }, businessName }),
    },
  }
```

https://github.com/user-attachments/assets/5dd6df53-18ba-4196-922d-a7f59e9e6cd3

### DIFF
--- a/packages/visual-editor/src/components/migrations/migrationRegistry.ts
+++ b/packages/visual-editor/src/components/migrations/migrationRegistry.ts
@@ -1,3 +1,8 @@
-import { MigrationRegistry } from "@yext/visual-editor";
+import { MigrationRegistry } from "../../utils/migrate.ts";
 
+// To add a migration:
+// Create a new file in this directory that exports a Migration
+// Import it in this file and add it to this array.
+// The migrations are run in the order of this array
+// and the data's version number indicates the last applied index.
 export const migrationRegistry: MigrationRegistry = [];
--- a/packages/visual-editor/src/utils/README.md
+++ b/packages/visual-editor/src/utils/README.md
@@ -788,7 +788,7 @@ const myComponentFields: Fields<MyComponentProps> = {
 `migrate` transforms Puck layout data to handle updates to the Puck version and to `visual-editor` components.
 It is run when data is loaded into the editor (both published and save state). It should also be
 run before using `<Render>` in a template. It does not currently handle dropzones but will be updated
-in future version to handle slots.
+in a future version to handle slots.
 
 `migrate` first runs Puck's `migrate` function to handle Puck migrations and then applies
 the migrations specified in `components/migrations/migrationRegistry.ts`.


packages/visual-editor/src/internal/puck/components/LayoutHeader.tsx
```diff
@@ -55,6 +55,35 @@ export const LayoutHeader = (props: LayoutHeaderProps) => {
       <div className="header-left">
         <UIButtonsToggle showLeft={true} />
         <EntityFieldsToggle />
+        {localDev && (
```
asanehisa, line 58: Oh neat this only pops up on `npm run dev` ?

packages/visual-editor/src/utils/migrate.ts
```diff
@@ -0,0 +1,103 @@
...
+  Config,
+  Data,
+  DefaultComponentProps,
+  DefaultRootProps,
+  migrate as migratePuck,
+  transformProps,
+} from "@measured/puck";
+import { migrationRegistry as commonMigrationRegistry } from "../components/migrations/migrationRegistry.ts";
+
+export type MigrationAction =
```
asanehisa, line 11: is this the formatting after prettier?

packages/visual-editor/src/internal/puck/components/LayoutHeader.tsx
benlife5: Yep! Only when in `visual-editor/starter`

packages/visual-editor/src/utils/migrate.ts
benlife5: Yeah, always seems off to me too haha

packages/visual-editor/src/internal/puck/components/LayoutHeader.tsx
asanehisa: Oooh so have we 2 development flows: 
localDev in visual-editor/starter/
devMode in pages-visual-editor-starter

packages/visual-editor/src/components/migrations/migrationRegistry.ts
```diff
@@ -0,0 +1,3 @@
+import { MigrationRegistry } from "@yext/visual-editor";
```
mkilpatrick, line 1: I'm always not clear if something should be imported from the library itself for the registry stuff or if you can use the file directly.

packages/visual-editor/src/internal/hooks/useMessageReceivers.ts
```diff
@@ -42,11 +44,18 @@ export const useCommonMessageReceivers = (
       setTemplateMetadata(devMetadata);
       const puckConfig = componentRegistry.get(devMetadata.templateId);
       setPuckConfig(puckConfig);
-      setLayoutData({
-        root: {},
-        content: [],
-        zones: {},
-      });
+      // applies current migration version to empty data
```
mkilpatrick, line 47: Do we want to do this elsewhere (that's not localDev)? Also don't we want to set the version in the default templates? Obviously that data doesn't live in this lib yet but will.

packages/visual-editor/src/utils/README.md
```diff
@@ -782,3 +782,43 @@ const myComponentFields: Fields<MyComponentProps> = {
     },
   },
 ```
+
+## migrate
+
+`migrate` transforms Puck layout data to handle updates to the Puck version and to `visual-editor` components.
+It is run when data is loaded into the editor (both published and save state). It should also be
+run before using `<Render>` in a template. It does not currently handle dropzones but will be updated
+in future version to handle slots.
```
mkilpatrick, line 791: in a future version

packages/visual-editor/src/utils/README.md
```diff
@@ -782,3 +782,43 @@ const myComponentFields: Fields<MyComponentProps> = {
...
+    // This component will be removed from all layouts
+  },
+  ComponentName: {
+    action: "renamed"
+    newName: string;
+    // This component will be renamed in all layouts
+    // Updates Puck's "type" property in data
+  },
+  ComponentName: {
+    action: "updated"
```
mkilpatrick, line 818: What if a component is both renamed and updated?

packages/visual-editor/src/utils/migrate.ts
```diff
@@ -0,0 +1,103 @@
...
+      //   })
+      // })
+      data = applyMigrationAction(data, componentName, migrationAction);
+    });
+  });
+
+  if (!data.root.props) {
+    data.root.props = {};
+  }
+  data.root.props.version = migrationRegistry.length;
```
mkilpatrick, line 68: This means that if we update a component in one PR and then another component in another PR then those would be two separate registry entries and each would cause a new version, even though we could've combined them into one registry entry if we did them together. Just trying to think about when we have massive changes like now and it would be nice to have those all as one migration version and not 15.

packages/visual-editor/src/utils/migrate.ts
```diff
@@ -0,0 +1,103 @@
...
+  // Apply puck migrations
+  data = migratePuck(data);
+
+  const migrationsToApply = migrationRegistry.slice(version);
+  if (migrationsToApply.length === 0) {
+    return data;
+  }
+
+  migrationsToApply.forEach((migration) => {
+    Object.entries(migration).forEach(([componentName, migrationAction]) => {
```
mkilpatrick, line 50: Do we want to skip migrations that have already happened before? If there are 3 and we already ran through 2 in the past, we can skip those now right?

packages/visual-editor/src/internal/puck/components/LayoutHeader.tsx
benlife5: Yep!

packages/visual-editor/src/utils/README.md
benlife5: You could have one migration for rename and then one migration for update. I could create an action for that if we wanted though

packages/visual-editor/src/utils/migrate.ts
benlife5: Yeah that's fair. I was thinking the version number should work this way so that it's specific about what migrations have applied as opposed to it being ambiguous if a migration version had changed what migrations it includes. But maybe that's not a problem since sites are only updated when we cut releases.

packages/visual-editor/src/internal/hooks/useMessageReceivers.ts
benlife5: Yes, we'll want the default layout data to have a version number. Version is also applied when the data is received from the db so if we've set empty data in VES it'll get a version number when it loads in the editor

packages/visual-editor/src/components/migrations/migrationRegistry.ts
benlife5: My rule of thumb has been that files in the `components` dir should import from the library and files elsewhere should do direct imports. This file wouldn't ever be pulled by the CLI tool so I think it could be direct

mkilpatrick: Ok let's change to direct then

packages/visual-editor/src/utils/migrate.ts
mkilpatrick: I guess it doesn't actually matter. Whether a "migration" is a single change or a group of changes, they get run regardless. So even if version is 234534 it really doesn't matter. But that's a good point about the releases. There could be a PR that moves all the previous migrations into one final version. Though that's more work on our side to keep it "clean." The previous migrations before cleaning are necessary for the tests to pass.

Just rambling here. I'm not really providing any specific guidance.

packages/visual-editor/src/components/migrations/migrationRegistry.ts
```diff
@@ -0,0 +1,3 @@
...
+
+export const migrationRegistry: MigrationRegistry = [];
```
jwartofsky-yext, line 3: Is the end goal that we just continually update this file with all of the latest migrations?

Could you add a comment explaining that?

benlife5: Done

mkilpatrick (approved):

asanehisa (approved):

PR #455, opened 2025-05-14
fix: react warnings and RTF styling

benlife5:
I missed these earlier...
`value` for an `input` should not be undefined and then become defined
RTFs can have lots of elements so it should not a child of `<p>`
`text-h2-color`, etc. are no longer classes and can be removed


packages/visual-editor/src/components/atoms/maybeRTF.tsx
```diff
@@ -22,11 +22,7 @@ export const MaybeRTF = ({ data, ...props }: MaybeRTFProps) => {
     }
 
     if ("json" in data && data.json) {
-      return (
-        <Body {...props}>
-          <LexicalRichText serializedAST={JSON.stringify(data.json)} />
-        </Body>
-      );
+      return <LexicalRichText serializedAST={JSON.stringify(data.json)} />;
```
asanehisa, line 25: Hmm maybe we can still wrap in div with the fontFamily, etc props? That way if they aren't set in the LexicalRichText it'll use the theme stuff? (I assume)

benlife5: LexicalRichText applies it's own styles but I've updated to apply the theme stuff.

We're going to need to loop back to this for the swap to HTML because I know the idea was "if you mess up accessibility in the RTF fields, that's on you" but it's weird that all these RTF components don't apply the theme, even when you haven't done a bunch of crazy RTF stuff

benlife5: I'll start a thread in the proj channel to ask more specifically about what we're expecting here

asanehisa: Ah gotcha. if we do html, then it won't apply font unless user wanted a specific font in their RTF?

asanehisa (approved):

packages/visual-editor/src/components/atoms/maybeRTF.tsx
```diff
@@ -23,9 +23,31 @@ export const MaybeRTF = ({ data, ...props }: MaybeRTFProps) => {
...
     if ("json" in data && data.json) {
       return (
-        <Body {...props}>
-          <LexicalRichText serializedAST={JSON.stringify(data.json)} />
-        </Body>
+        <LexicalRichText
+          serializedAST={JSON.stringify(data.json)}
+          nodeClassNames={{
+            paragraph:
+              "components font-body-fontFamily font-body-fontWeight text-body-fontSize",
```
asanehisa, line 30: actually i don't think we wanna apply weight

asanehisa: Since ppl purposefully use RTFs to do weight stuff, so just family and size

mkilpatrick (dismissed):

mkilpatrick (approved):

jwartofsky-yext (approved):

PR #458, opened 2025-05-15
feat: support boolean and option with YextEntityField

benlife5:
Adds support for `type.boolean`

`type.option` returns for `type.option` and `type.string` unless the `expandOptionFields` has been applied, in which case it only returns for `type.option`

Multi-option has the same behavior but with `includeListsOnly:true`

https://github.com/user-attachments/assets/f785f5b5-0fa0-4272-b235-f49a8e663998

### DIFF
--- a/packages/visual-editor/src/internal/utils/getFilteredEntityFields.ts
+++ b/packages/visual-editor/src/internal/utils/getFilteredEntityFields.ts
@@ -42,6 +42,7 @@ export type EntityFieldTypes =
   | "type.phone"
   | "type.coordinate"
   | "type.cta"
+  | "type.boolean"
   | "type.faq_section"
   | "type.testimonials_section"
   | "type.products_section"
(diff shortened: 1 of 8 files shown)


asanehisa (commented):
Hmmm I wonder if its ever possible we'd wanna add "type.option" to supported if ppl wanna filter only by type.option (i imagine it must have a slightly diff use case from type.string?)

jwartofsky-yext (dismissed):

packages/visual-editor/src/internal/utils/getFilteredEntityFields.ts
```diff
@@ -144,6 +146,14 @@ const getEntityTypeToFieldNames = (
       const typeName = getTypeFromSchemaField(fieldToSchema.schemaField);
       if (typeName) {
         prev = appendToMapList(prev, typeName, fieldToSchema.name);
+
```
asanehisa, line 149: hmmm so multioption is just type.option that is a list, so does it always get used as type.string and type.option?

asanehisa: Oh jk I get it, multiption is type.list with list but follows same logic

> type.option returns for type.option and type.string unless the expandOptionFields has been applied, in which case it only returns for type.option

asanehisa (approved):

mkilpatrick (approved):

benlife5:
Confirmed the documents till has strings with the `replaceOptionValuesWithDisplayNames` transformation

PR #459, opened 2025-05-15
chore: improve automated tests

benlife5:
Each of the PageSections now has 4 tests (run for mobile and desktop):
- default props with empty document
- default props with data in the document
- version 0 props that use as many entity values as possible
- version 0 props that use as many constant values as possible
The rendered component is then tested for WCAG and that each piece of expected data appears on the page.

Migrations are run before each test so the a component's test should fail the v0 tests if it has prop changes that are not properly migration. (Conversely, updating `expect` for a version 0 tests indicates we are making breaking changes).


asanehisa (approved):
🔥🔥🔥🔥🔥 love these tests, we'll def need documentation / guidance on how to add ur own (for devs making their own components)

jwartofsky-yext (approved):
nice!

mkilpatrick (commented):
1. Where does the migrate function get called for the tests?
2. If we change a component then we'd add another another set of 4 ComponentTests?

benlife5:
> Where does the migrate function get called for the tests?

The page section tests each have something like
```
      const data = migrate(
        {
          root: { version },
          content: [
            {
              type: "CoreInfoSection",
              props: props,
            },
          ],
        },
        migrationRegistry,
        puckConfig
      );
      const { container } = reactRender(
        <VisualEditorProvider templateProps={{ document }}>
          <Render config={puckConfig} data={data} />
        </VisualEditorProvider>
      );
```

> If we change a component then we'd add another another set of 4 ComponentTests?

Depends but probably just 2 -- I think the defaultProps tests don't need to be versioned, they'll just run against the current component. So it would be adding a test that uses entity fields and a test that uses constant values with new version

mkilpatrick (approved):

jwartofsky-yext (approved):

asanehisa (approved):

PR #460, opened 2025-05-15
feat: expand font options

jwartofsky-yext:
Adds a lot more font options from the Google Fonts API

Searchbar to come in another PR


packages/visual-editor/src/utils/applyTheme.test.ts
```diff
@@ -172,9 +172,4 @@ const themeConfig: ThemeConfig = {
...
...00&family=Vollkorn:ital,wght@0,400..900;1,400..900&family=Vollkorn+SC:wght@400..900&family=Wavefont:wght@100..900&family=WindSong:wght@400..500&family=Winky+Rough:ital,wght@0,300..900;1,300..900&family=Winky+Sans:ital,wght@0,300..900;1,300..900&family=Wittgenstein:ital,wght@0,400..900;1,400..900&family=Wix+Madefor+Display:wght@400..800&family=Wix+Madefor+Text:ital,wght@0,400..800;1,400..800&family=Work+Sans:ital,wght@0,100..900;1,100..900&family=Xanh+Mono:ital,wght@0,400..400;1,400..400&family=Yaldevi:wght@200..700&family=Yanone+Kaffeesatz:wght@200..700&family=Yantramanav:wght@100..900&family=Yrsa:ital,wght@0,300..700;1,300..700&family=Ysabeau:ital,wght@0,100..900;1,100..900&family=Ysabeau+Infant:ital,wght@0,100..900;1,100..900&family=Ysabeau+Office:ital,wght@0,100..900;1,100..900&family=Ysabeau+SC:wght@100..900&family=Zain:ital,wght@0,200..900;1,200..900&family=Zen+Kaku+Gothic+Antique:wght@300..900&family=Zen+Kaku+Gothic+New:wght@300..900&family=Zen+Loop:ital,wght@0,400..400;1,400..400&family=Zen+Maru+Gothic:wght@300..900&family=Zen+Old+Mincho:wght@400..900&family=Zilla+Slab:ital,wght@0,300..700;1,300..700&family=Zilla+Slab+Highlight:wght@400..700&display=swap" rel="stylesheet">';
```
benlife5, line 175: I think this URL is too long, I guess it's going to need to be broken up into multiple `<Link>` tags

packages/visual-editor/src/utils/visualEditorFonts.ts
```diff
@@ -6,31 +6,831 @@ type FontSpecification = {
...
+  Aleo: { italics: true, minWeight: 100, maxWeight: 900, fallback: "serif" },
+  Alexandria: { italics: false, minWeight: 100, maxWeight: 900, fallback: "sans-serif" },
+  Alkatra: { italics: false, minWeight: 400, maxWeight: 700, fallback: "sans-serif" },
+  Allan: { italics: false, minWeight: 400, maxWeight: 700, fallback: "sans-serif" },
+  Almarai: { italics: false, minWeight: 300, maxWeight: 800, fallback: "sans-serif" },
+  Almendra: { italics: true, minWeight: 400, maxWeight: 700, fallback: "serif" },
+  "Alumni Sans": { italics: true, minWeight: 100, maxWeight: 900, fallback: "sans-serif" },
+  "Alumni Sans Collegiate One": { italics: true, minWeight: 400, maxWeight: 400, fallback: "sans-serif" },
+  "Alumni Sans Inline One": { italics: true, minWeight: 400, maxWeight: 400, fallback: "sans-serif" },
+  "Alumni Sans Pinstripe": { italics: true, minWeight: 400, maxWeight: 400, fallback: "sans-serif" },
```
benlife5, line 40: I don't think the code I wrote to transform this object to a URL handles fonts with the same minWeight and maxWeight correctly, sorry

packages/visual-editor/src/utils/visualEditorFonts.ts
```diff
@@ -6,31 +6,831 @@ type FontSpecification = {
...
+  "Ysabeau Office": { italics: true, minWeight: 100, maxWeight: 900, fallback: "sans-serif" },
+  "Ysabeau SC": { italics: false, minWeight: 100, maxWeight: 900, fallback: "sans-serif" },
+  Zain: { italics: true, minWeight: 200, maxWeight: 900, fallback: "sans-serif" },
+  "Zen Kaku Gothic Antique": { italics: false, minWeight: 300, maxWeight: 900, fallback: "sans-serif" },
+  "Zen Kaku Gothic New": { italics: false, minWeight: 300, maxWeight: 900, fallback: "sans-serif" },
+  "Zen Loop": { italics: true, minWeight: 400, maxWeight: 400, fallback: "sans-serif" },
+  "Zen Maru Gothic": { italics: false, minWeight: 300, maxWeight: 900, fallback: "sans-serif" },
+  "Zen Old Mincho": { italics: false, minWeight: 400, maxWeight: 900, fallback: "serif" },
+  "Zilla Slab": { italics: true, minWeight: 300, maxWeight: 700, fallback: "serif" },
+  "Zilla Slab Highlight": { italics: false, minWeight: 400, maxWeight: 700, fallback: "serif" },
```
mkilpatrick, line 833: add your script as part of this PR

packages/visual-editor/src/utils/visualEditorFonts.ts
jwartofsky-yext: It seems to handle it correctly

It gets set to the weight and there's only one option in the selector for those fonts

packages/visual-editor/src/utils/applyTheme.test.ts
jwartofsky-yext: Done

packages/visual-editor/src/utils/visualEditorFonts.ts
```diff
@@ -51,24 +32,34 @@ export const constructFontSelectOptions = (fonts: FontRegistry) => {
  * Note: Google does not return font file URLs if the params
  * fall outside of the font's supported values
  */
-const constructGoogleFontLinkTags = (fonts: FontRegistry) => {
+const constructGoogleFontLinkTags = (fonts: FontRegistry): string => {
   const prefix = `<link rel="preconnect" href="https://fonts.googleapis.com">
```
benlife5, line 36: We should only need to include the preconnect tags once

packages/visual-editor/src/utils/visualEditorFonts.ts
jwartofsky-yext: Fixed

Will now just have 1 weight if the max and min are the same

benlife5:
This LGTM. My only note is that I think we need to test to see if there's an impact on page load time for live pages. it should be fine/necessary for the editor but for the generated pages, it's loading a lot more files than it did previously. It doesn't load the actual font file unless its used, so maybe it's ok because it's just a few thousand lines of css text from a Google CDN, but we should check.

jwartofsky-yext:
> This LGTM. My only note is that I think we need to test to see if there's an impact on page load time for live pages. it should be fine/necessary for the editor but for the generated pages, it's loading a lot more files than it did previously. It doesn't load the actual font file unless its used, so maybe it's ok because it's just a few thousand lines of css text from a Google CDN, but we should check.

Live page loads fine

I think this should be good to go 👍

benlife5 (approved):

PR #461, opened 2025-05-16
test github action

benlife5:
Do not merge, just testing the action


.github/workflows/check-migrations.yml
```diff
@@ -0,0 +1,43 @@
...
+
+      - name: Comment on PR
+        if: env.components_updated == 'true' && env.migrations_updated == 'false'
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const prNumber = context.payload.pull_request.number;
+            const commentBody = "Warning: Component files have been updated but no migrations have been added.";
+
+            await github.rest.issues.createComment({
```
mkilpatrick, line 39: Can this fail the PR instead of just adding a warning comment? It might already do that and new actions just don't run in the PR they're added.

benlife5: It could, but I don't think all component changes actually need migrations (like if it was a minor styling fix) so it really is more of a warning

asanehisa: Oooh when we have documentation, the action can also link to a place where u can see how to decide if u need to add a migration and how to actually add it

mkilpatrick: You could create an empty markdown file now and link to that in Github. We probably wouldn't want to link to the live doc site here.

PR #462, opened 2025-05-16
feat: add migration check

benlife5:
Adds a github action that checks if component files have been updated. If they have and migrations have not, a comment is added to the PR with a warning. Doesn't actually check for anything other than files being updated so it's far from comprehensive but may be a helpful nudge?

See example here: #461 (the first commit had component changes but no migration, the second commit component+migration changes, and the third no component changes)

### DIFF
--- a/packages/visual-editor/src/components/migrations/README.md
+++ b/packages/visual-editor/src/components/migrations/README.md
@@ -1,8 +1,8 @@
 # Migrations
 
 Migrations ensure backwards compatibility for component updates. When a site is updated, it immediately begins
-using the new components however, the published and save state data is old. This data is run through
-`migrate` prior to being rendered so that any migrations will be applied and the data will match with the components.
+using the new components; however, the published and save state data may not reflect the schema of these component updates.
+This data is run through `migrate` prior to being rendered so that any migrations will be applied and the data will match with the components.
 
 ## When do I need to add a migrations?
 
@@ -21,8 +21,7 @@ If you update some fields, such as select, the new options may not match the old
 
 4. A component is removed (optional)
 
-A removed component will automatically disappear from the live page, however adding a
-removed migration will also remove the "No Configuration for" message from the editor.
+A removed component will automatically disappear from the live page as well as showing a "No Configuration for" message in the editor. Add a removed migration to handle this.
 
 ## How do I create a migration?
 
@@ -30,10 +29,11 @@ In the `migrations` directory, create a new file of the form `1234_example_migra
 the number increases sequentially and the rest of the name is a brief description of the migration.
 
 In that file, export an object of type `Migration`, which is object mapping component names to
-`MigrationActions`. The component name should be the name the component is registered as in
-`ve.config.ts` or `_componentCategories.ts`.
+`MigrationActions` (see [migrate.ts](https://github.com/yext/visual-editor/blob/main/packages/visual-editor/src/utils/migrate.ts)).
+The component name should be the name the component is registered as in
+`ve.config.ts` or [`_componentCategories.ts`](https://github.com/yext/visual-editor/blob/main/packages/visual-editor/src/components/_componentCategories.ts).
 
-There are three `MigrationActions`:
+There are three [`MigrationActions`](https://github.com/yext/visual-editor/blob/1210ee5bae73bff1456563b57506ff163fa59cb6/packages/visual-editor/src/utils/migrate.ts#L11):
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


jwartofsky-yext:
This seems like it could be helpful, but I think there would be a lot of cases where the component files are updated but nothing needs to change in the migration.

This would mean that the warning would be misguided most of the time, and I think people would just ignore it.

Is there a way we could check if the props were changed instead?

Maybe we would have to run a script using ts-morph or something like that

benlife5:
Yeahh, I go back and forth on whether this would actually be helpful. I'm not sure there's a good way to actually know whether changes need a migration or not. Probably could check the prop type, but the problem is there's other breaking things you could do without modifying the props, like changing puck fields or not rendering certain pieces of data.

asanehisa (dismissed):
I think the warning is nice as long as yeah ppl don't start to ignore it

benlife5:
Went ahead and wrote a basic README for migrations. Also updated the check to look specifically at `migrationRegistry.ts`

packages/visual-editor/src/components/migrations/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+Without a migration, new props will not have any data and data contained in old props will not be used.
+
+3. Component fields are updated
+
+If you update some fields, such as select, the new options may not match the old data.
+
+4. A component is removed (optional)
+
+A removed component will automatically disappear from the live page, however adding a
```
mkilpatrick, line 24: A removed component will automatically disappear from the live page as well as showing a "No Configuration for" message in the editor. Add a removed migration to handle this.

packages/visual-editor/src/components/migrations/README.md
```diff
@@ -0,0 +1,75 @@
...
+removed migration will also remove the "No Configuration for" message from the editor.
+
+## How do I create a migration?
+
+In the `migrations` directory, create a new file of the form `1234_example_migration.ts` where
+the number increases sequentially and the rest of the name is a brief description of the migration.
+
+In that file, export an object of type `Migration`, which is object mapping component names to
```
mkilpatrick, line 32: You could link the functions/types to the github code.

packages/visual-editor/src/components/migrations/README.md
```diff
@@ -0,0 +1,75 @@
...
+`MigrationActions`. The component name should be the name the component is registered as in
+`ve.config.ts` or `_componentCategories.ts`.
+
+There are three `MigrationActions`:
```
mkilpatrick, line 36: Same

packages/visual-editor/src/components/migrations/README.md
```diff
@@ -0,0 +1,75 @@
...
+Renames all existing usages of a component to `newName`.
+This name should be the name of the component as it is registered as in `ve.config.ts` or `_componentCategories.ts`.
+
+```ts
+{ action: "renamed", newName: string }
+```
+
+### Updated
+
+Transforms the existing props to the new set of props.
```
mkilpatrick, line 59: Maybe show a simple example implementation.

packages/visual-editor/src/components/migrations/README.md
```diff
@@ -0,0 +1,75 @@
...
+```ts
+{
+  action: "updated";
+  propTransformation: (oldProps: Record<string, any>) => Record<string, any>;
+}
+```
+
+In `migrationRegistry.ts`, import your migration and append it to the array. Migrations are run in order.
+
+## How do I test a migration?
```
mkilpatrick, line 71: Should we talk about adding new test cases too?

mkilpatrick (approved):

packages/visual-editor/src/components/migrations/README.md
```diff
@@ -0,0 +1,99 @@
...
+  propTransformation: ({ heading, ...props }) => ({ title: heading, ...props });
+}
+```
+
+In `migrationRegistry.ts`, import your migration and append it to the array. Migrations are run in order.
+
+## How do I test a migration?
+
+In the local dev starter, create a layout with the existing components. Log the layout data
+and save it somewhere. Then, make your component updates. In the editor, set the layout data
```
jwartofsky-yext, line 81: What's your method for setting the data in the editor to the old data?

Are you putting it into browser memory so it loads that way?

benlife5: In one of the other PRs, I added a way to get/set layout data in localDev 
Screenshot 2025-05-19 at 10 39 24 AM

jwartofsky-yext: Ah I see, very nice

PR #464, opened 2025-05-19
fix: remove text-body-fontSize-sm

benlife5:
`text-body-fontSize-sm` isn't a valid class name (`text-body-sm-fontSize` is). These two components aren't using the `sm` variant in the mocks anyway

### DIFF
--- a/packages/visual-editor/src/components/Directory.tsx
+++ b/packages/visual-editor/src/components/Directory.tsx
@@ -86,7 +86,7 @@ const DirectoryCard = ({
         />
       )}
       {profile.address && (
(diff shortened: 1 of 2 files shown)


jwartofsky-yext (commented):
Does changing from an invalid classname to a valid one change how the page looks?

Otherwise lgtm

benlife5:
Good point -- it could change how the page looks because right now they're the a constant font size and after this they will be responsive to the theme, but I think that's ok

jwartofsky-yext (approved):

mkilpatrick (approved):

PR #465, opened 2025-05-19
feat: use app API key for analytics

jwartofsky-yext:
Uses YEXT_PUBLIC_VISUAL_EDITOR_APP_API_KEY instead of YEXT_PUBLIC_EVENTS_API_KEY

Both keys just point to the app which has all permissions needed for VE.

### DIFF
--- a/packages/visual-editor/src/vite-plugin/templates/main.tsx
+++ b/packages/visual-editor/src/vite-plugin/templates/main.tsx
@@ -93,7 +93,7 @@ const Location: Template<TemplateRenderProps> = (props) => {
 
   return (
     <AnalyticsProvider
-      apiKey={document?._env?.YEXT_PUBLIC_EVENTS_API_KEY}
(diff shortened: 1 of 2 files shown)


packages/visual-editor/src/vite-plugin/templates/main.tsx
```diff
@@ -93,7 +93,7 @@ const Location: Template<TemplateRenderProps> = (props) => {
 
   return (
     <AnalyticsProvider
-      apiKey={document?._env?.YEXT_PUBLIC_EVENTS_API_KEY}
+      apiKey={document?._env?.YEXT_PUBLIC_VISUAL_EDITOR_APP_API_KEY}
```
mkilpatrick, line 96: You'll need to make sure this gets updated in Watson's locator code too

mkilpatrick (approved):

packages/visual-editor/src/vite-plugin/templates/main.tsx
jwartofsky-yext: will do, thanks for the callout

PR #467, opened 2025-05-19
fix: fix image warnings

jwartofsky-yext:
Fixes warnings that appear when trying to use fixed size without height and width.

If height and width are not present, fallback to the intrinsic size.


packages/visual-editor/src/components/pageSections/EventSection.tsx
```diff
@@ -98,11 +98,7 @@ const EventCard = ({
       <div className="lg:w-[45%] w-full h-full">
         {event.image && (
           <div className="h-full">
-            <Image
-              image={event.image}
-              layout="auto"
-              aspectRatio={event.image.width / event.image.height}
-            />
+            <Image image={event.image} layout="auto" />
```
benlife5, line 101: I think we still want to pass an aspect ratio here for entity values

packages/visual-editor/src/components/atoms/image.tsx
```diff
@@ -32,15 +32,30 @@ export const Image: React.FC<ImageProps> = ({
...
           className="object-cover"
         />
+      ) : (
+        <img
+          src={isComplexImageType(image) ? image.image.url : image.url}
+          alt={
+            isComplexImageType(image)
+              ? (image.image.alternateText ?? "")
+              : (image.alternateText ?? "")
+          }
```
benlife5, line 50: Do we want `object-cover` on this image too?

jwartofsky-yext: Done

benlife5 (approved):

mkilpatrick (approved):

PR #469, opened 2025-05-20
docs: add hybrid implementation docs

benlife5:


HybridDevelopment.md
```diff
@@ -0,0 +1,120 @@
+# Hybrid Site Development Guide
```
benlife5, line 1: @mkilpatrick where should this file live?

HybridDevelopment.md
```diff
@@ -0,0 +1,120 @@
...
+
+1. Create a new repository using https://github.com/YextSolutions/pages-visual-editor-starter
+2. In the Yext Platform, navigate to Pages > All Sites in the sidebar and create a new site
+   via the "Add New Site" button. Use the repo you created in step 1.
+3. In Knowledge Graph, create a Site entity.
+   It is standard to use your business name as the site entity's "name" field.
+4. Configure your [site stream](https://hitchhikers.yext.com/docs/pages/site-configuration/?target=site-stream)
+   by updating your `config.yaml` to use the site entity's id in place of `entity-id`.
+   Also add any alternate profiles you wish to use with your site entity.
+5. Set up content endpoints?
```
benlife5, line 13: @jwartofsky-yext the content endpoints only get created via the get started flow? 
If so, @mkilpatrick what should the recommendation be here?

jwartofsky-yext: Content Endpoints are now created during pageset creation. There is 1 content endpoint/pageset

So they should be created for this flow when the user creates the pageset

benlife5: Ah cool ok. Do you know if that works with repo page set creation too? Or just in-platform?

jwartofsky-yext: It should work for both

If they're creating the pageset through the Pagesets page in YextSitesStorm, it will work

HybridDevelopment.md
mkilpatrick: packages/visual-editor/src/docs/hybrid-development.md

HybridDevelopment.md
mkilpatrick: You don't create a pageset for in-repo. They already exist based on what's in the repo and the details come from the deploy. They are not a traditional "pageset" resource. They only need to set up content endpoints if they plan to use the nearby locations component.

They will need to set up an app with the events api for analytics though.

HybridDevelopment.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Types of Page Groups
+
+### In-Platform VLE Page Groups
+
+In-Platform Page Groups can be created on the Page Groups page in the Yext platform.
+They use a pre-set stream with all entity fields that are not linked-entity fields included.
+Currently, all In-Platform Page Groups use the `main` template. There can be multiple
```
mkilpatrick, line 21: Idk if in this section, but we should describe how an in-repo template works with in-platform pagesets.

A template effectively just defines the components available to the VLE pageset.
The pageset is created through the Storm UI and specifically uses the main.tsx template, with the exception for the other built-in pageset types for directory and locator. 

For the hybrid case, if you want the ability to create in-platform pagesets you can define the components in the registry in ve.config.tsx. If you don't want in-platform pagesets created and only want the in-repo pagesets to show up, remove the "main" registry entry (or don't include it since we removed the file from the starter).

HybridDevelopment.md
```diff
@@ -0,0 +1,120 @@
...
+
+At the top level of the `TemplateConfig` include:
+
+```ts
+additionalProperties: {
+  isVETemplate: true;
+}
+```
+
+It is required to include an entity type filter, even if also using a saved filter.
```
mkilpatrick, line 57: I wonder if we can make TS require it when isVETemplate is set.

HybridDevelopment.md
```diff
@@ -0,0 +1,120 @@
...
+which defines the components available to that template. To associate a config with a template:
+
+1. Use it in the template's Render component `<Render config={} ...`
+2. Register it in the `componentRegistry` of `ve.config.tsx` so it can be used in the editor
+
+```ts
+export const componentRegistry = new Map<string, Config<any>>([
+  ["main", mainConfig],
+  ["directory", directoryConfig],
+  // templateName must match the file name of the template (excluding the file extension)
```
mkilpatrick, line 116: If the TemplateConfig's name is set then it needs to match that. The order of precedence is TemplateConfig.name > filename.

HybridDevelopment.md
jwartofsky-yext: I think we will want to go with an approach where we use Site _env vars as a fallback if the _pageset data is not available

https://yext.slack.com/archives/C06A06BCUUF/p1747943146412829?thread_ts=1747844077.461129&cid=C06A06BCUUF

HybridDevelopment.md
benlife5: We don't really have a way to remove the ability to create in-platform page sets currently. If you remove main from ve.config.tsx, they will fail to generate but the create button still exists. If you remove main from template-manifest, the main template is still injected by the plugin and tries (and fails) to generate as a static page.

mkilpatrick (approved):

packages/visual-editor/src/docs/hybrid-development.md
```diff
@@ -0,0 +1,136 @@
+# Hybrid Site Development Guide
```
mkilpatrick, line 1: Add this to the beginning

```
---
title: Hybrid Site Development Guide
outline: deep
---
```

mkilpatrick (approved):

PR #473, opened 2025-05-22
refactor: handle missing data in Insights and Products

benlife5:
Thread: https://yext.slack.com/archives/C06A06BCUUF/p1747766568494449?thread_ts=1747680405.497999&cid=C06A06BCUUF

- Standardizes image size and adds a placeholder if the image doesn't exist on desktop
- Pushes the CTA to the end of the card 

This isn't perfect / I'm sure there will be more changes at some point, but I think this handles it better

https://github.com/user-attachments/assets/54835842-2d4c-455c-8aba-ab0ef9d88f0d

### DIFF
--- a/packages/visual-editor/src/components/pageSections/ProductSection.tsx
+++ b/packages/visual-editor/src/components/pageSections/ProductSection.tsx
@@ -90,17 +90,20 @@ const ProductCard = ({
 }) => {
   return (
     <Background
-      className="flex flex-col justify-between rounded-lg overflow-hidden border h-full"
+      className="flex flex-col rounded-lg overflow-hidden border h-full"
       background={backgroundColor}
     >
-      {product.image && (
+      {product.image ? (
         <Image
           image={product.image}
           layout={"auto"}
-          aspectRatio={product.image.width / product.image.height}
+          aspectRatio={1.778} // 16:9
+          className="h-[200px]"
         />
+      ) : (
+        <div className="sm:h-[200px]" />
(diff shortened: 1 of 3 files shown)


mkilpatrick (approved):

asanehisa (approved):

packages/visual-editor/src/components/pageSections/ProductSection.tsx
```diff
@@ -90,17 +90,20 @@ const ProductCard = ({
...
+      className="flex flex-col rounded-lg overflow-hidden border h-full"
       background={backgroundColor}
     >
-      {product.image && (
+      {product.image ? (
         <Image
           image={product.image}
           layout={"auto"}
-          aspectRatio={product.image.width / product.image.height}
+          aspectRatio={1.778} // 16:9
```
jwartofsky-yext, line 100: Would it be better to do something like `!!product?.image?.width ? (product.image.width / product.image.height) : 1.778`?

benlife5: It's a little unclear from the mocks but my interpretation was that we always want these images to be 16:9

jwartofsky-yext (approved):

packages/visual-editor/src/components/pageSections/ProductSection.tsx
jwartofsky-yext: ack

PR #483, opened 2025-05-29
feat: extract strings for translation

jwartofsky-yext:
Adds two scripts to extract strings for translation and populate the default translations file.

1. i18n:insert - modifies components in src/components to add i18n() wrapping
   * there is also an i18n:insert:dry variant

2. i18n:scan - uses i18n-scan to populate locales/en

Moves generateFontRegistry.js to the new /scripts folder

Adds an i18n wrapper function under utils/i18n.ts. This will use the i18n translation if present and configured and returns the unmodified string otherwise.

Includes the modifications to src/components/** as done by i18n:insert

Includes the populated locales/en/visual-editor.json as generated by i18n:scan


packages/visual-editor/i18next-scanner.config.cjs
```diff
@@ -0,0 +1,29 @@
...
+  output: "",
+
+  options: {
+    debug: true,
+    defaultValue: (lng, ns, key) => key,
+    func: {
+      list: ["t", "i18next.t", "i18n"],
+      extensions: [".ts", ".tsx"],
+    },
+    lngs: ["en"],
```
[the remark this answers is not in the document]
jwartofsky-yext: This is to generate the initial file only. The rest are generated by another script which will be in another PR.

packages/visual-editor/package.json
```diff
@@ -116,15 +121,19 @@
...
     "gray-matter": "^4.0.3",
     "husky": "^9.1.4",
+    "i18next-scanner": "^4.6.0",
     "jest-axe": "^10.0.0",
     "jsdom": "^24.1.1",
     "libphonenumber-js": "^1.11.19",
     "lint-staged": "^15.2.7",
     "minimist": "^1.2.8",
+    "node-fetch": "^3.3.2",
+    "p-limit": "^6.2.0",
```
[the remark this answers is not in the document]
jwartofsky-yext: oops, those are meant to be in the other PR 😢

packages/visual-editor/package.json
```diff
@@ -116,15 +121,19 @@
...
     "playwright": "^1.51.1",
     "react-icons": "^5.2.1",
     "semver": "^7.6.3",
     "tailwindcss": "^3.4.6",
+    "ts-morph": "^26.0.0",
```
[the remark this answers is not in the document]
jwartofsky-yext: This is used for the insertI18n.js script to modify src/components

packages/visual-editor/i18next-scanner.config.cjs
```diff
@@ -0,0 +1,29 @@
+module.exports = {
```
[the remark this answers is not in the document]
jwartofsky-yext: I tried originally with regular .js, but I wasn't able to get it to work

Changing to .cjs was able to resolve the issue however.

I can try to find what the original error was

packages/visual-editor/i18next-scanner.config.cjs
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   input: [
-    "src/**/*.{ts,tsx}",
+    "src/components/**/*.{ts,tsx}",
+    "src/editor/**/*.{ts,tsx}",
     "!**/__screenshots__/**", // exclude this folder and all files inside it
```
mkilpatrick, line 5: Probably don't need this one now with the other folder inputs.

packages/visual-editor/package.json
```diff
@@ -73,12 +73,10 @@
     "lz-string": "1.5.0",
     "next-themes": "^0.3.0",
     "picocolors": "^1.0.1",
-    "prettier": "^3.3.3",
     "prompts": "^2.4.2",
     "pure-react-carousel": "^1.32.0",
     "react-color": "^2.19.3",
     "react-error-boundary": "^5.0.0",
-    "react-i18next": "^15.5.2",
```
mkilpatrick: I'm surprised we don't need this

packages/visual-editor/scripts/insertI18n.ts
```diff
@@ -0,0 +1,141 @@
+import { Project, SyntaxKind } from "ts-morph";
+import prettier from "prettier";
+import fs from "fs";
+
+// CLI flag
+const isDryRun = process.argv.includes("--dry-run");
+
+/**
+ * Runs through everything under src/components and adds i18n("...") to all appropriate strings
```
mkilpatrick, line 9: and src/editor

packages/visual-editor/scripts/insertI18n.ts
```diff
@@ -0,0 +1,141 @@
...
+ */
```
mkilpatrick, line 10: Does this wrap all strings or only those in YextField or label/title/description elements? Does it catch all React code like
```
<div>Some string</div>
```

packages/visual-editor/src/utils/i18n.ts
```diff
@@ -10,17 +10,9 @@ const fallbackT = (key: string) => key;
  *
  * @param key translation key
  */
-export const i18n = (key: string) => {
+export const i18n = (key: string): string => {
```
mkilpatrick, line 13: How are we handling the complexities of i18n? We need to support plural, context, etc.

packages/visual-editor/i18next-scanner.config.cjs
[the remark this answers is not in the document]
jwartofsky-yext: i18next-scanner expects commonJS and gets internal errors if I try to change it to mjs

I think we should keep this as is

packages/visual-editor/i18next-scanner.config.cjs
jwartofsky-yext: There's screenshot folders inside the component folders, ex:

src/components/contentBlocks/__screenshots

packages/visual-editor/package.json
jwartofsky-yext: It's possible for us to use i18n outside of react context and use something like useI18n inside the react context.

Parsing out what is a react component vs what isn't programmatically is complicated.. but possible

react-i18next will update when the language changes. 

i18next will only update if the page is refreshed

packages/visual-editor/i18next-scanner.config.cjs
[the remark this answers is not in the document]
jwartofsky-yext: Changed to include the other translation files

packages/visual-editor/package.json
jwartofsky-yext: So we do want to use react-i18next where we can

we can use the hook useI18n in components inside a react context

To translate stuff in the editor/config, we will likely need to make a field override. I looked into this a good bit today, but didn't quite finish

https://discord.com/channels/1153376562259951687/1153376562830397563/1371140919717269644

Will continue on Monday

I included translations for the strings inside the configs in the "visual-editor.json" translation files, but did not keep the i18n in the code as it doesn't work there.

asanehisa (approved):

packages/visual-editor/i18next-scanner.config.cjs
```diff
@@ -0,0 +1,63 @@
...
+    lngs: [
+      "en",
+      "zh_CN",
+      "zh_TW",
+      "hr_HR",
+      "cs_CZ",
+      "da_DK",
+      "nl",
+      "en_GB",
+      "en_US",
```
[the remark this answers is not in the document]
jwartofsky-yext: Perhaps not, but en_US is listed in storm (and not en).

I think it would be ok to use en instead of en_US

packages/visual-editor/src/components/pageSections/HeroSection.tsx
```diff
@@ -236,11 +253,15 @@ const HeroSectionWrapper = ({ data, styles }: HeroSectionProps) => {
...
-          displayName="Image"
+          displayName={i18n("image", { defaultValue: "Image" })}
           fieldId={data.hero.field}
           constantValueEnabled={data.hero.constantValueOverride.image}
         >
-          <div className="w-full" role="region" aria-label="Hero Image">
+          <div
+            className="w-full"
+            role="region"
+            aria-label={i18n("heroImage", { defaultValue: "Hero Image" })}
```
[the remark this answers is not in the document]
mkilpatrick: Thanks for the detail. A couple of comments:
1. I'm pretty sure you can set up react-i18next with a default namespace in its config so you don't need to pass it every time.
2. Since we extract all values and non-translated strings are just empty string, this would already handle a bomb-out scenario, right? We will want the locale translations (for constant value user input) to work the same way as profile fields where they are language-specific, so if there isn't a value it doesn't just automatically fall back to a different language (en). This could be different for the Editor translations, but that's why we'll have an action to fail the PR if there are missing translations. All that being said, does it make sense to still keep this wrapper?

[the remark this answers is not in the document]
jwartofsky-yext: No, a bombout scenario without the wrapper does not use the default value. It will use the key (and also seems to lowercase it for some reason). Bombout as in i18n is not initialized. If i18n is initialized it should use the default value as a fallback without the wrapper.

The wrapper handles the bomb-out to make sure the default value is used and the key is never shown on the page.

I'll test with the default value namespace and see if removing the wrapper is a good idea.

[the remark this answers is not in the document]
jwartofsky-yext: Removed the i18n wrapper

briantstephan (approved):

packages/visual-editor/src/editor/YextEntityFieldSelector.tsx
```diff
@@ -372,9 +372,7 @@ export const EntityFieldInput = <T extends Record<string, any>>({
     return BasicSelector(templateMetadata.entityTypeDisplayName + " Field", [
       {
         value: "",
-        label: t("selectAContentField", {
-          defaultValue: "Select a Content field",
-        }),
+        label: t("basicSelectorDefaultContentLabel", "Select a Content field"),
```
mkilpatrick, line 375: Just basicSelectorContentLabel (remove default)

packages/visual-editor/scripts/insertI18n.ts
```diff
@@ -168,9 +187,7 @@ const isDryRun = process.argv.includes("--dry-run");
     // Now modify all collected nodes and track functions to add t()
     for (const { attr, raw, funcAncestor } of jsxAttrsToModify) {
       const key = toCamelCase(raw);
-      attr.setInitializer(
-        `{t("${key}", { defaultValue: "${escapeString(raw)}" })}`
-      );
+      attr.setInitializer(`{t("${key}", "${escapeString(raw)}")}`);
```
mkilpatrick, line 190: Do we care that the key is the full text? The dev would need to manually adjust but I guess that's ok and this is a best effort.

jwartofsky-yext: We could maybe try to figure out what the key should be programmatically, but that may be complicated

I think this is a good start

A lot of the keys are probably fine this way too

PR #489, opened 2025-05-30
feat: show entity type name in field label

briantstephan:
This sets the label above entity field selectors to "[entity_type_display_name] Field" using the display name provided by the template metadata, when it would previously always show "Entity Field". This also adds a new `useTemplateMetadata` hook to facilitate it.

### DIFF
--- a/packages/visual-editor/src/editor/Editor.tsx
+++ b/packages/visual-editor/src/editor/Editor.tsx
@@ -116,7 +116,7 @@ export const Editor = ({
   });
 
   return (
-    <TemplateMetadataContext.Provider value={templateMetadata}>
+    <TemplateMetadataContext.Provider value={templateMetadata!}>
       <ErrorBoundary fallback={<></>} onError={logError}>
         {!isLoading ? (
           templateMetadata?.isThemeMode || forceThemeMode ? (
--- a/packages/visual-editor/src/editor/YextEntityFieldSelector.tsx
+++ b/packages/visual-editor/src/editor/YextEntityFieldSelector.tsx
@@ -366,11 +366,7 @@ export const EntityFieldInput = <T extends Record<string, any>>({
       });
     }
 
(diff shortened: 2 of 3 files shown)


packages/visual-editor/src/editor/Editor.tsx
```diff
@@ -113,31 +116,33 @@ export const Editor = ({
...
-            themeConfig={themeConfig}
-            localDev={!!localDev}
-          />
+    <TemplateMetadataContext.Provider value={templateMetadata!}>
+      <ErrorBoundary fallback={<></>} onError={logError}>
+        {!isLoading ? (
+          templateMetadata?.isThemeMode || forceThemeMode ? (
+            <ThemeEditor
+              puckConfig={puckConfig!}
+              templateMetadata={templateMetadata!}
```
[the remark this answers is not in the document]
briantstephan: We're not prop drilling here since `templateMetadata` seems to be used in every child component in the chain, so I think it's fine to leave it. Each removal of the prop would require an addition of the hook, so it wouldn't be any simpler.

packages/visual-editor/src/editor/YextEntityFieldSelector.tsx
```diff
@@ -364,7 +366,7 @@ export const EntityFieldInput = <T extends Record<string, any>>({
       });
     }
 
-    return BasicSelector("Entity Field", [
+    return BasicSelector(templateMetadata.entityTypeDisplayName + " Field", [
```
jwartofsky-yext, line 369: Is there ever a case where templateMetadata.entityTypeDisplayName could be null?

If so, we could set a backup '?? "Entity"'

briantstephan: I don't think that's possible. `templateMetadata` will never be null at this point, and we always set the display name on it in storm before it's sent down.

mkilpatrick (approved):

PR #495, opened 2025-06-02
feat: use YEXT_EDIT_LAYOUT_MODE_MAPBOX_API_KEY when in VLE if present

k-gerner:
We will use an environment-wide Mapbox key (in the `YEXT_EDIT_LAYOUT_MODE_MAPBOX_API_KEY` env var) when in the VLE iframe in order to get around the URL restrictions by the site-specific API key. If this key is not present for some reason, we will fall back to the site-specific API key. 

J=[WAT-4873](https://yexttest.atlassian.net/browse/WAT-4873)
TEST=manual

### DIFF
--- a/packages/visual-editor/src/components/Locator.tsx
+++ b/packages/visual-editor/src/components/Locator.tsx
@@ -432,8 +432,6 @@ const Map: React.FC<MapProps> = ({
     typeof document === "undefined"
       ? undefined
       : (document.getElementById("preview-frame") as HTMLIFrameElement);
-  const entityDocument: any = useDocument();
-  const mapboxApiKey = entityDocument._env?.YEXT_MAPBOX_API_KEY;
   //@ts-expect-error MapboxGL is not loaded in the iframe content window
   if (iframe?.contentDocument && !iframe.contentWindow?.mapboxgl) {
     // We are in an iframe, and mapboxgl is not loaded in yet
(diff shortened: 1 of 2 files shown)


packages/visual-editor/src/utils/searchHeadlessConfig.ts
```diff
@@ -48,6 +48,17 @@ export const createSearchHeadlessConfig = (document: any) => {
...
   }
 
+  // The iframe'd locator will need a non-URL-restricted Mapbox API key in order to display the map
+  // properly. Without this key, we will fall back to the site-specific Mapbox API key, which is not
+  // guaranteed to work in the iframe.
+  const iframeMapboxApiKey =
+    document?._env?.YEXT_EDIT_LAYOUT_MODE_MAPBOX_API_KEY;
+  if (!iframeMapboxApiKey) {
+    console.warn(
+      "Missing YEXT_EDIT_LAYOUT_MODE_MAPBOX_API_KEY! Some map behavior may be unavailable in the layout editor."
```
jwartofsky-yext, line 58: This message is incorrect if we were looking for the other key

packages/visual-editor/src/components/Locator.tsx
```diff
@@ -448,6 +446,16 @@ const Map: React.FC<MapProps> = ({
     );
   }
 
+  const entityDocument: any = useDocument();
+  let mapboxApiKey = entityDocument._env?.YEXT_MAPBOX_API_KEY;
```
jwartofsky-yext, line 450: Couldn't you just do something like `mapboxApiKey = entityDocument._env?.YEXT_MAPBOX_API_KEY ?? entityDocument?._env?.YEXT_EDIT_LAYOUT_MODE_MAPBOX_API_KEY; `?

packages/visual-editor/src/utils/searchHeadlessConfig.ts
k-gerner: Sorry which part is incorrect?

PR #497, opened 2025-06-03
feat: resize tooltips based on viewport

benlife5:
Updates the entity field tooltip size based on the viewport size/zoom. Uses similar logic to how puck scales the component name in the action bar.

https://github.com/user-attachments/assets/8b1a1b11-0c3c-4018-9a0c-55bf9e09d329

### DIFF
--- a/packages/visual-editor/src/internal/puck/ui/Tooltip.tsx
+++ b/packages/visual-editor/src/internal/puck/ui/Tooltip.tsx
@@ -16,13 +16,24 @@ interface TooltipContentProps
   zoomWithViewport?: boolean;
 }
 
+// Safe fallback if tooltip renders outside of puck context
+const tryUsePuck = () => {
+  try {
+    const puck = usePuck();
+    return puck;
+  } catch {
+    return undefined;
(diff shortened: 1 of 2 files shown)


mkilpatrick (dismissed):
Nice

packages/visual-editor/src/internal/puck/ui/Tooltip.tsx
```diff
@@ -16,13 +16,24 @@ interface TooltipContentProps
   zoomWithViewport?: boolean;
 }
 
+// Safe fallback if tooltip renders outside of puck context
```
mkilpatrick, line 19: Was this to avoid a blowup on the published page?

benlife5: It was blowing up the test file. I tested published pages and they were fine since this is only in-editor but I think this fallback is good just in case

asanehisa (approved):

jwartofsky-yext (approved):
lgtm

PR #500, opened 2025-06-03
fix: hide email list length when using constant value

briantstephan:
This hides the email list length option when the field is set to a constant value since it is redundant.

### DIFF
--- a/packages/visual-editor/src/components/contentBlocks/Emails.tsx
+++ b/packages/visual-editor/src/components/contentBlocks/Emails.tsx
@@ -13,7 +13,7 @@ import {
 
 export interface EmailsProps {
   list: YextEntityField<string[]>;
-  listLength: number;
+  listLength?: number;
   includeHyperlink: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


packages/visual-editor/src/components/contentBlocks/Emails.tsx
```diff
@@ -85,6 +80,20 @@ const EmailsComponent: React.FC<EmailsProps> = ({
 export const Emails: ComponentConfig<EmailsProps> = {
   label: "Emails",
   fields: EmailsFields,
+  resolveFields: (data, { fields }) => {
+    if (data.props.list.constantValueEnabled) {
+      return fields;
```
benlife5, line 85: Does this remove the field if you flip from constant -> entity -> constant?

briantstephan: Tested manually and verified that the option is hidden if the emails are set to a constant value, then will show up if switched to an entity field, then will be hidden again when switching back to constant, and so on. The setting also persists when switching between the two, so nothing is lost.

PR #503, opened 2025-06-04
feat: set translation languages

benlife5:
Creates two i18n Instances:
1. `i18nComponents` is set based on the document locale . A normal `useTranslation` hook call returns this instance's `t`,
2. `i18nPlatform` is set based on the storm user's language. `usePlatformTranslation` returns the instance's `t`.

Added buttons to the localDev mode to test locales. Will look into cleaning up the header in a separate PR.

https://github.com/user-attachments/assets/424eaef2-b55a-4300-b931-a2a2b3773666

https://github.com/user-attachments/assets/7a3862b6-e80c-4d34-84d4-616fc2cecd90


packages/visual-editor/src/editor/Editor.tsx
```diff
@@ -103,6 +104,14 @@ export const Editor = ({
     }
   }, [templateMetadata?.isDevMode, devPageSets]);
 
+  useEffect(() => {
+    if (templateMetadata?.platformLocale) {
+      i18nPlatformInstance.changeLanguage(
+        templateMetadata?.platformLocale.replace("-", "_")
```
mkilpatrick, line 110: Why do we want the underscore locale? That's non-standard.

packages/visual-editor/src/utils/i18nComponents.ts
```diff
@@ -0,0 +1,57 @@
+import i18next from "i18next";
+import { initReactI18next } from "react-i18next";
+
+const NAMESPACE = "visual-editor";
+
+const i18nComponentsInstance = i18next.createInstance();
+
+const resources: Record<string, any> = {};
+const modules = import.meta.glob("../../locales/*/visual-editor.json", {
```
mkilpatrick, line 9: Not you, but does glob properly handle paths on other OSes? If not we should use path to construct the path.

packages/visual-editor/src/utils/i18nComponents.ts
```diff
@@ -0,0 +1,57 @@
...
+      [NAMESPACE]: (modules[path] as any).default,
+    };
+  }
+}
+
+i18nComponentsInstance.use(initReactI18next).init({
+  // TODO: replace these fallbacks once we remove the dialects
+  // from the stored translations
+  fallbackLng: {
+    cs: ["cs_CZ"],
```
mkilpatrick, line 28: Also not you, but it's unclear to me why we would need these fallbacks.

packages/visual-editor/src/utils/i18nComponents.ts
```diff
@@ -0,0 +1,57 @@
...
+  if (match) {
+    const lang = match[1];
+    resources[lang] = {
+      [NAMESPACE]: (modules[path] as any).default,
+    };
+  }
+}
+
+i18nComponentsInstance.use(initReactI18next).init({
+  // TODO: replace these fallbacks once we remove the dialects
```
mkilpatrick, line 25: Noting that we SHOULD have dialects in our stored translations.

packages/visual-editor/src/utils/i18nPlatform.ts
```diff
@@ -0,0 +1,36 @@
+import i18next from "i18next";
+import { initReactI18next, useTranslation } from "react-i18next";
+
+const NAMESPACE = "visual-editor";
```
mkilpatrick, line 4: Should we use a different namespace? Otherwise these two are the same code just creating 2 different instances, right?

packages/visual-editor/src/editor/Editor.tsx
benlife5: I think Intl and google take `-` but the Yext platform spits out `_` so right now our translation files also use `_`. We'll need to convert somewhere but this should be handled centrally rather than ad hoc

packages/visual-editor/src/utils/i18nComponents.ts
benlife5: cc @jwartofsky-yext but my understanding is we'll need a real translation service to do that

packages/visual-editor/src/utils/i18nComponents.ts
benlife5: For example, if the document locale is `es`, it does not automatically fallback to `es_ES`. We need to map the supported document locales to the languages we have translations for.

jwartofsky-yext: Right now, we only have translation folders for various dialects

i18n will automatically fallback from a dialect to the more generic language (ex "es_ES" to "es"), but not the other way

Since the translations we have from Google don't actually use dialects, I'm thinking we should just change them to the generic language (or at least also add the generic languages).

Then we won't need the fallbacks here since we will have "es" to fallback to

packages/visual-editor/src/utils/i18nPlatform.ts
jwartofsky-yext: It's probably fine for them to use the same namespace. That will just make them use the same translation files.

The reason for two instances is that we can have two languages selected at once

One for the platform and one for the document

packages/visual-editor/src/editor/Editor.tsx
jwartofsky-yext: I made another PR which changes everything to kebab-case, so I think we should go with kebab everywhere https://github.com/yext/visual-editor/pull/506

packages/visual-editor/src/utils/i18nComponents.ts
benlife5: Yeah according to Google, import.meta.glob is cross platform since it's a vite feature

PR #509, opened 2025-06-05
feat: add platform translations part 1

benlife5:
Adds translations for the editor text that will use the Storm user locale. Didn't update the insert script because a lot of these were one-off situations or required a restructure.

The field labels are still set in the components and require a more complicated translation setup. Looking into that now.

https://github.com/user-attachments/assets/132c528c-53f4-485f-8a87-c0af8e1744a9


packages/visual-editor/src/internal/puck/components/LoadingScreen.tsx
```diff
@@ -2,16 +2,24 @@ import { Progress } from "../ui/Progress.tsx";
...
+  progress,
+  platformLanguageIsSet,
+}: LoadingScreenProps) {
+  const { t } = usePlatformTranslation();
   return (
     <div className="ve-flex ve-h-screen ve-loading-wrapper ve-w-screen ve-flex-col ve-items-center ve-justify-center">
       <Progress className="ve-w-1/3 ve-loading-progress" value={progress} />
-      <div>Loading Visual Editor...</div>
+      {platformLanguageIsSet && (
+        <div>{t("loadingVE", "Loading Visual Editor")}...</div>
```
mkilpatrick, line 21: The ellipses should probably be part of the translation too. For instance, for langs that read RTL. You could google, but that's probably the case.

packages/visual-editor/src/internal/puck/components/ThemeHeader.tsx
```diff
@@ -185,7 +191,10 @@ export const ThemeHeader = (props: ThemeHeaderProps) => {
               await onPublishTheme();
             }}
           >
-            {`Update ${totalEntityCount} ${totalEntityCount === 1 ? "Page" : "Pages"}`}
+            {
+              // TODO: translation concatenation
```
mkilpatrick, line 195: What do you mean by this?

packages/visual-editor/locales/ja/visual-editor.json
```diff
@@ -1,6 +1,29 @@
 {
+  "Category": "カテゴリ",
+  "FAQ": "よくある質問",
+  "FontSize": "フォントサイズ",
+  "Name": "名前",
+  "Title": "タイトル",
   "address": "住所",
+  "addressInput": {
+    "city": "市",
```
asanehisa, line 9: Hmmm so just a note, but for addresses in Japan. This isn't the input structure used in general. But I'm not sure if our JS Address component easily works with different countries address systems. 

pic of how KG asks for address if country = japan (but also even if language is japanese doesn't mean the entities are in japan sooo)
Screenshot 2025-06-05 at 2 55 24 PM

benlife5: @mkilpatrick Is this worth working on now or should we wait until it's requested?

mkilpatrick: We probably don't even want core info to allow static values anymore to be honest. We can check on that with Aaron. But to the point of the question, we can ignore for now until it's requested.

mkilpatrick (approved):

jwartofsky-yext (dismissed):

packages/visual-editor/src/editor/FontSizeSelector.tsx
```diff
@@ -1,16 +1,17 @@
...
 import { Combobox } from "../internal/puck/ui/Combobox.tsx";
+import { usePlatformTranslation } from "../utils/i18nPlatform.ts";
 
 export const fontSizeOptions = (includeLargeSizes = true) => {
+  const { t } = usePlatformTranslation();
   const fontSizeOptions = [
     { label: "XS", value: "xs", px: "12" },
     { label: "SM", value: "sm", px: "14" },
-    { label: "Base", value: "base", px: "16" },
+    { label: t("base", "Base"), value: "base", px: "16" },
```
jwartofsky-yext, line 14: This is probably fine, but might be good to add a context here

You can do so like this: t("base", { defaultValue: "Base", context: "font" }

This would help if there's any case where there's a language where "base" is not appropriate here

jwartofsky-yext (approved):

packages/visual-editor/src/editor/FontSizeSelector.tsx
benlife5: ok cool, I can come back through and add if we hookup context to the translate script

PR #533, opened 2025-06-13
chore: hide translation diffs

benlife5:
Screenshot 2025-06-13 at 11 06 54 AM

I'm not sure if this is better but we might want can auto-hide the translation diffs? I wish they appeared smaller

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+starter/localData/** linguist-generated=true
+starter/src/dev.config.ts linguist-generated=true
+packages/visual-editor/locales/** linguist-generated=true
+packages/visual-editor/locales/en/visual-editor.json linguist-generated=false


mkilpatrick (approved):

.gitattributes
```diff
@@ -0,0 +1,4 @@
+starter/localData/** linguist-generated=true
+starter/src/dev.config.ts linguist-generated=true
+packages/visual-editor/locales/** linguist-generated=true
```
mkilpatrick, line 3: Do you want **/* so all the translation files are collapsed?

mkilpatrick (approved):

.gitattributes
benlife5: `locales/**` should match all children

PR #537, opened 2025-06-13
feat: add heading align for testimonial component

colton-demetriou:
Missed this in the first heading align pr, adding now.

### DIFF
--- a/packages/visual-editor/src/components/migrations/0002_add_heading_alignment.ts
+++ b/packages/visual-editor/src/components/migrations/0002_add_heading_alignment.ts
@@ -59,4 +59,8 @@ export const addHeadingAlignmentMigration: Migration = {
     action: "updated",
     propTransformation: updateHeadingStructure,
   },
+  TestimonialSection: {
+    action: "updated",
+    propTransformation: updateHeadingStructure,
+  },
 };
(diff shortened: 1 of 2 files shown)


packages/visual-editor/src/components/pageSections/TestimonialSection.tsx
```diff
@@ -188,7 +211,7 @@ const TestimonialSectionWrapper = ({
                 key={index}
                 testimonial={testimonial}
                 backgroundColor={styles.cardBackgroundColor}
-                sectionHeadingLevel={styles.headingLevel}
+                sectionHeadingLevel={styles?.heading?.level ?? 2}
```
asanehisa, line 214: hmm shouldn't need` ?? 2 `considering u have the migration. and i think its ok to add this to the older migration since no sites are using that change yet (since we didn't cut a release)

packages/visual-editor/src/components/migrations/0002_add_heading_alignment.ts
```diff
@@ -59,4 +59,8 @@ export const addHeadingAlignmentMigration: Migration = {
     action: "updated",
     propTransformation: updateHeadingStructure,
   },
+  TestimonialSection: {
+    action: "updated",
+    propTransformation: updateHeadingStructure,
```
asanehisa, line 64: Oh actually when u add migrations, ur suppose to update tests too so u can test the migrations per https://github.com/yext/visual-editor/blob/88fc8a13150567979753800b800d37c1cfda265d/packages/visual-editor/src/components/migrations/README.md#adding-test-cases

PR #538, opened 2025-06-13
feat: edit constant value rtf fields in drawer

benlife5:
Pops a drawer in Storm that will be used to edit RTF. For now, these fields just use a simple string.

Although RTF could have been done without storm, the mocks show the drawer taking the full screen height which can't be done in the iframe, we get access to Mana, and we're going to need to do this for image assets anyway.

To Do (probably in separate PRs): 
- Build RTF editor in storm
- Handle storing/passing/rendering RTF
- Handle or potentially remove non translatable RTF?

https://github.com/user-attachments/assets/2bd5cfa1-aca9-49a6-af44-5c96b8676f1d

### DIFF
--- a/packages/visual-editor/src/editor/TranslatableRichTextField.tsx
+++ b/packages/visual-editor/src/editor/TranslatableRichTextField.tsx
@@ -2,47 +2,82 @@ import { TranslatableRichText } from "../types/types.ts";
 import { useDocument } from "../hooks/useDocument.tsx";
 import { usePlatformTranslation } from "../utils/i18nPlatform.ts";
 import { Translation } from "../internal/types/translation.ts";
-import { AutoField, CustomField, FieldLabel } from "@measured/puck";
+import { CustomField, FieldLabel } from "@measured/puck";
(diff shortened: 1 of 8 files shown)


packages/visual-editor/src/editor/TranslatableRichTextField.tsx
```diff
@@ -2,47 +2,82 @@ import { TranslatableRichText } from "../types/types.ts";
...
+            type: "RTF",
+            value: resolvedValue,
+            id: messageId,
+            fieldName: fieldLabel,
+          },
+        });
+
+        // localDev
+        if (
+          window.location.href.includes("http://localhost:5173/dev-location")
```
mkilpatrick, line 63: Does this work in proxied dev mode in Storm?

mkilpatrick (dismissed):

packages/visual-editor/src/editor/TranslatableRichTextField.tsx
benlife5: The real drawer comes up in storm dev mode

PR #546, opened 2025-06-18
feat: add hasLocalizedValue to translation struct 

asanehisa:
Screenshot 2025-06-18 at 10 04 39 AM

confirmed it doesn't break my site and the layout data looks like this 

Adding this so we have a way to check if a static value should expect values for each locale.

### DIFF
--- a/packages/visual-editor/src/editor/TranslatableRichTextField.tsx
+++ b/packages/visual-editor/src/editor/TranslatableRichTextField.tsx
@@ -70,7 +70,8 @@ export function TranslatableRichTextField<
         onChange({
           ...(typeof value === "object" && !Array.isArray(value) ? value : {}),
           [locale]: newValue,
-        } as T);
+          hasLocalizedValue: "true",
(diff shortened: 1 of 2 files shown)


packages/visual-editor/src/editor/TranslatableRichTextField.tsx
```diff
@@ -70,7 +70,8 @@ export function TranslatableRichTextField<
         onChange({
           ...(typeof value === "object" && !Array.isArray(value) ? value : {}),
           [locale]: newValue,
-        } as T);
+          hasLocalizedValue: "true",
```
benlife5, line 73: "true" is a string because it needs to be Record<string, string>?

benlife5 (commented):
LGTM, will need to add to all the defaultProps too

benlife5 (approved):

mkilpatrick (approved):

benlife5 (approved):

PR #549, opened 2025-06-18
feat: add rtf drawer functionality

colton-demetriou:
Tested in local dev, verified that RTF styling works and is locale specific.

https://www.loom.com/share/15269807f812418b96c9da7b47aa89ba?sid=1ad487cd-6c6a-46b5-b819-9b000f51c5c6

### DIFF
--- a/packages/visual-editor/THIRD-PARTY-NOTICES
+++ b/packages/visual-editor/THIRD-PARTY-NOTICES
@@ -233,7 +233,7 @@ SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 The following npm package may be included in this product:
 
- - eslint@9.25.1
+ - eslint@9.29.0
 
 This package contains the following license:
 
@@ -261,7 +261,7 @@ THE SOFTWARE.
 
 The following npm package may be included in this product:
 
- - dompurify@3.2.5
+ - dompurify@3.2.6
 
 This package contains the following license:
 
@@ -886,7 +886,6 @@ OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 The following npm packages may be included in this product:
 
  - @measured/puck@0.18.3
- - @radix-ui/react-slot@1.2.0
  - next-themes@0.3.0
 
 These packages each contain the following license:
@@ -1017,7 +1016,7 @@ SOFTWARE.
 
 The following npm package may be included in this product:
 
- - @tanstack/react-query@5.74.7
+ - @tanstack/react-query@5.80.7
 
 This package contains the following license:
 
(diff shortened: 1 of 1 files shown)


packages/visual-editor/src/components/TranslatableRichTextDrawer.tsx
```diff
@@ -0,0 +1,229 @@
...
+    <Dialog.Root open={open} onOpenChange={setOpen}>
+      <Dialog.Trigger asChild>
+        <div
+          className={`ve-p-2 ve-border ve-rounded ve-cursor-pointer ve-bg-white hover:ve-bg-gray-50 ${styles["ve-multiline-ellipsis"]}`}
+          tabIndex={0}
+          role="button"
+          aria-label="Edit rich text"
+        >
+          {renderPreview() || (
+            <span className="ve-text-gray-400">Click to edit...</span>
```
benlife5, line 66: Add a platform translation for these strings

packages/visual-editor/src/components/TranslatableLexicalEditor.tsx
```diff
@@ -0,0 +1,87 @@
...
+  value,
+  onChange,
+  translations = {},
+  onTranslationChange,
+  currentLocale = "en",
+  availableLocales = ["en"],
+}: TranslatableLexicalEditorProps) {
+  const [selectedLocale, setSelectedLocale] = useState(currentLocale);
+  const { appState } = usePuck();
+  const isDrawerOpen = appState?.ui?.rightSideBarVisible;
```
benlife5, line 25: This shouldn't be related to puck's sidebar?

packages/visual-editor/src/components/LexicalEditor.tsx
```diff
@@ -0,0 +1,488 @@
...
+import { Pencil } from "lucide-react";
+import React from "react";
+
+interface LexicalEditorProps {
+  value: string | RichText;
+  onChange: (value: string | RichText) => void;
+  showToolbar?: boolean;
+}
+
+const theme = {
```
benlife5, line 50: I _think_ this should be based on lexical-theme.module.scss in alpha, although we'll need to be make some modifications to support theme

benlife5 (commented):
Oh this reimplements the drawer in VE rather than storm but that's probably fine

mkilpatrick:
> Oh this reimplements the drawer in VE rather than storm but that's probably fine
Hmm, I think we want the Storm drawer version so we can do mana-ish things, though it is nicer to have fully in VE because then it's usable in proxied and true local dev modes.

mkilpatrick:
> > Oh this reimplements the drawer in VE rather than storm but that's probably fine
> 
> Hmm, I think we want the Storm drawer version so we can do mana-ish things, though it is nicer to have fully in VE because then it's usable in proxied and true local dev modes.

Talked offline. We need to do most things in Storm related to RTF and the messages to/from VE/Storm should be:
VE --> Storm : locale, rtfJson, and whatever causes the modal to pop in RTF mode
Storm --> VE : locale, rtfJson, rtfHtml

PR #550, opened 2025-06-18
fix: limit mapbox size

briantstephan:
This limits the mapbox height and width to the size limits defined here: https://docs.mapbox.com/api/maps/static-images/#retrieve-a-static-map-from-a-style


benlife5 (approved):
ha I didn't realized we already had a max defined, nice

asanehisa (approved):

mkilpatrick (approved):

PR #551, opened 2025-06-18
feat: add defaultProps to arrays 

asanehisa:
https://jam.dev/c/84d96b29-565d-470b-87fb-d596c518785e

Wasn't there before but now when u add new "cards", they look nicer / not empty.

(and i need hasLocalizedValues)

### DIFF
--- a/packages/visual-editor/src/internal/puck/constant-value-fields/FAQsSection.tsx
+++ b/packages/visual-editor/src/internal/puck/constant-value-fields/FAQsSection.tsx
@@ -60,6 +60,16 @@ const FAQStructArrayField = (): ArrayField<FAQStruct[]> => {
       question: questionField,
       answer: answerField,
     },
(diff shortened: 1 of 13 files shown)


benlife5 (dismissed):

packages/visual-editor/src/internal/puck/constant-value-fields/FAQsSection.tsx
```diff
@@ -60,6 +60,16 @@ const FAQStructArrayField = (): ArrayField<FAQStruct[]> => {
       question: questionField,
       answer: answerField,
     },
+    defaultItemProps: {
```
mkilpatrick, line 63: Make as const in the internal package and export so it can be imported into pageSections to avoid duplication. Though it does seem backwards. Thanks shadcli.

mkilpatrick (approved):

PR #554, opened 2025-06-23
feat: add links to nearby location cards

briantstephan:


packages/visual-editor/src/components/pageSections/NearbyLocations.tsx
```diff
@@ -220,14 +224,23 @@ const LocationCard = ({
...
+  slug?: string;
 }) => {
   return (
     <Background
       background={styles?.cardBackgroundColor}
       className="flex flex-col flew-grow h-full rounded-lg overflow-hidden border p-6 sm:p-8"
       as="section"
     >
-      <Heading level={styles?.cardHeadingLevel}>{name}</Heading>
+      <MaybeLink
```
benlife5, line 236: This renders with the Link styles rather than the Heading styles, right? Might be worth checking with product about that, since having a cardHeadingLevel doesn't make sense if it doesn't render as a heading

briantstephan: It should render the Heading styles too, since the heading is passed directly into the CTA atom as the label when the link exists. Confirmed with Aaron that we still want the card heading level prop.

benlife5: Oh I see, my bad. Sounds good

asanehisa (approved):

mkilpatrick (approved):

PR #555, opened 2025-06-23
fix: add missing locale labels

benlife5:
We didn't have locale labels for basic string constant values because they don't have a sublabel

https://github.com/user-attachments/assets/0251a211-807c-428c-9396-dff5fb8d6c49


packages/visual-editor/src/editor/YextEntityFieldSelector.tsx
```diff
@@ -258,7 +267,9 @@ export const ConstantValueModeToggler = ({
         </TooltipProvider>
       )}
       <p className="ve-self-center ve-text-sm ve-text-gray-800 ve-font-semibold">
-        {pt(label)}
+        {showLocale && constantValueEnabled
+          ? `${pt(label)} (${locale})`
+          : `${pt(label)}`}
```
mkilpatrick, line 272: Does this actually work (the pt part) when the string is a variable and not an actual value?

benlife5: It will translate it if a translation exists. Since it's a variable, the possible options need to be wrapped in `msg` wherever they are defined for the scanner to create a translation for them.

PR #559, opened 2025-06-24
feat: update i18n for context

asanehisa:
Appears to be working for contexts such as "direction", "text size", and "phone number" based on my limited understanding of spanish / japanese.

This basically passes it into google api like "Right [[direction']]" and then removes the context afterwards so we get "Derecha" for spanish's right (direction) instead of "Bien" which is more like right (correct).


benlife5 (approved):
Nice!!

packages/visual-editor/scripts/generateTranslations.ts
```diff
@@ -23,6 +23,44 @@ type TranslationSegment = [
...
 
+// Context separator used by i18next-scanner
+const CONTEXT_SEPARATOR = "_";
+
+// Unique markers for embedding context into the text for translation
+const CONTEXT_MARKER_START = "[[";
+const CONTEXT_MARKER_END = "]]";
+
+// Google Translate API very rarely removes/adds [ or ]
+// Matches one or more opening "[", then any characters (non-greedy), then one or more closing "]"
```
mkilpatrick, line 34: Shouldn't we look specifically for two brackets instead of one or more?

mkilpatrick (approved):

packages/visual-editor/scripts/generateTranslations.ts
mkilpatrick: Very odd.

PR #560, opened 2025-06-24
feat: add rtf functionality

asanehisa:
Adjusted code to work with YSS RTF changes. This should be released after YSS is shipped. This also handles multiple locales + having a diff RTF static value for each locale. 

https://github.com/user-attachments/assets/7798edc0-f0db-4b3f-a0b5-9e19767d2dff


packages/visual-editor/src/utils/resolveTranslatableString.tsx
```diff
@@ -42,63 +42,58 @@ export const resolveTranslatableRichText = (
...
-  return "";
 };
 
 /**
- * Takes a TranslatableString and a locale and returns the value to be displayed in the editor input
- * @param translatableString a TranslatableString
- * @param locale "en" or other locale value
- * @return string to be displayed in the editor input
+ * Safely renders a TranslatableRichText value, ensuring it never returns an object
+ * @param translatableRichText
```
benlife5, line 75: descriptions for these

packages/visual-editor/src/utils/resolveTranslatableString.tsx
```diff
@@ -42,63 +42,58 @@ export const resolveTranslatableRichText = (
...
-    return richTextToString(translatableString);
-  }
-
-  const localizedValue: string | RichText = translatableString[locale];
+export const safeRenderTranslatableRichText = (
+  translatableRichText?: TranslatableRichText,
+  locale: string = "en"
+): React.ReactElement | string => {
+  try {
+    const result = resolveTranslatableRichText(translatableRichText, locale);
```
benlife5, line 83: I'm not sure we need this function, shouldn't `resolveTranslatableRichText` be guaranteed to return a valid value?

packages/visual-editor/src/editor/TranslatableRichTextField.tsx
```diff
@@ -22,8 +22,11 @@ export function TranslatableRichTextField<
     render: ({ onChange, value }) => {
       const { i18n } = useTranslation();
       const locale = i18n.language;
-      const resolvedValue = getDisplayValue(value, locale);
-      const fieldLabel = (label && pt(label)) + ` (${locale})`;
+      const resolvedValue = safeRenderTranslatableRichText(value, locale);
+      const fieldLabel = label ? (label && pt(label)) + ` (${locale})` : "";
+
+      // Ensure fieldLabel is a string
+      const safeFieldLabel = typeof fieldLabel === "string" ? fieldLabel : "";
```
benlife5, line 29: Shouldn't need this type check

packages/visual-editor/src/editor/TranslatableStringField.tsx
```diff
@@ -38,9 +44,11 @@ export function TranslatableStringField<
         return <div className={"ve-pt-3"}>{autoField}</div>;
       }
 
-      return (
-        <FieldLabel label={pt(label) + ` (${locale})`}>{autoField}</FieldLabel>
-      );
+      const fieldLabel = pt(label) + ` (${locale})`;
+      // Ensure fieldLabel is a string
+      const safeFieldLabel = typeof fieldLabel === "string" ? fieldLabel : "";
```
benlife5, line 49: Same, make fieldLabel guaranteed to be a string via TS

packages/visual-editor/src/utils/resolveTranslatableString.tsx
asanehisa: done cuz deleted

mkilpatrick:
Does this work with translations? Aka if you change locale then you can have a different rtf?

packages/visual-editor/src/editor/TranslatableRichTextField.tsx
```diff
@@ -38,41 +41,82 @@ export function TranslatableRichTextField<
...
         }
       );
 
       const handleClick = () => {
         const messageId = `RichText-${Date.now()}`;
         setPendingMessageId(messageId);
+
+        // Extract the appropriate value to send to Storm
+        let valueToSend = resolvedValue;
+        if (
```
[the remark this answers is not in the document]
asanehisa: Yup fixed

packages/visual-editor/src/editor/TranslatableRichTextField.tsx
```diff
@@ -38,35 +38,57 @@ export function TranslatableRichTextField<
...
+              handleNewValue(payload.value || "", locale);
+            }
           }
         }
       );
 
       const handleClick = () => {
         const messageId = `RichText-${Date.now()}`;
         setPendingMessageId(messageId);
+        const valueForLocale = (value as Record<string, any>)[locale];
```
benlife5, line 55: Could this be combined with the logic on line 60 to make it a little clearer what is happening? I don't love the `as` typecast even though it's safe-ish here

packages/visual-editor/src/editor/TranslatableRichTextField.tsx
```diff
@@ -38,35 +38,57 @@ export function TranslatableRichTextField<
...
+            locale: locale,
           },
         });
 
         // localDev
         if (
           window.location.href.includes("http://localhost:5173/dev-location")
         ) {
-          handleNewValue(prompt("Enter text:") ?? "");
+          handleNewValue(prompt("Enter text:") ?? "", locale);
```
benlife5, line 73: What form of data does this take in? The goal here was to have some semi-easy way to render RTF in local dev without the full editor

packages/visual-editor/src/internal/puck/constant-value-fields/TextList.tsx
```diff
@@ -176,7 +176,10 @@ export const TRANSLATABLE_TEXT_LIST_CONSTANT_CONFIG: CustomField<
               key={locale}
               field={{ type: "text" }}
               id={`${id}-value-${index}`}
-              value={getDisplayValue(item, locale)}
+              value={(() => {
+                const resolvedValue = resolveTranslatableRichText(item, locale);
```
benlife5, line 180: Can you add a comment explaining what this is doing?

packages/visual-editor/src/editor/TranslatableRichTextField.tsx
asanehisa: Fixed so it takes in HTML and if not HTML then it treats as string. JSON was a bit funkier cuz I'd have to detect if user had stringified JSON or just plain JSON

benlife5 (approved):

packages/visual-editor/src/editor/TranslatableRichTextField.tsx
```diff
@@ -22,51 +22,80 @@ export function TranslatableRichTextField<
...
         "constantValueEditorClosed",
         TARGET_ORIGINS,
         (_, payload) => {
           if (pendingMessageId && pendingMessageId === payload?.id) {
-            handleNewValue(payload.value);
+            // Handle the new Storm payload structure with locale, rtfJson, and rtfHtml
+            if (payload.locale && payload.rtfJson) {
+              handleNewValue(payload.rtfJson, payload.locale, payload.rtfHtml);
+            } else {
+              // Fallback for backward compatibility
```
mkilpatrick, line 46: Once Storm is deployed it will always send the new payload right? If so then you shouldn't need any fallback.

packages/visual-editor/src/utils/resolveTranslatableString.tsx
```diff
@@ -34,33 +34,48 @@ export const resolveTranslatableString = (
...
  */
 export const resolveTranslatableRichText = (
   translatableRichText?: TranslatableRichText,
   locale: string = "en"
 ): string | React.ReactElement => {
-  if (!translatableRichText) return "";
+  try {
+    if (!translatableRichText) return "";
+
+    if (
```
mkilpatrick, line 48: There's this if condition, the one below, and then the logic in isRichText which also checks if the type is an object like on line 55. Can this be simplified to just check if translatableRichText is a map, grab the locale, and then always call toStringOrElement which happens in both of these conditions?

packages/visual-editor/src/utils/resolveTranslatableString.tsx
```diff
@@ -118,5 +126,12 @@ function toStringOrElement(
   if (isRichText(value)) {
     return <MaybeRTF data={value} />;
   }
-  return value ?? "";
+
+  // Ensure we only return strings
+  if (typeof value === "string") {
```
mkilpatrick, line 131: Hmm, we can't migrate the strings to rtf because we don't know the formatting and that happens in Storm, right? But now the rtf fields from Storm that we're going to render are always the html version so can't we just always pretend it's a string we need to render? Or is this to support the json version from the document because Spruce hasn't switch it over yet?

Basically I'm trying to see if we can simplify a lot of the crazy logic in here. Is that possible? If not now, will it be possible at some point, and if so, what is needed to do that?

asanehisa: No I think ur right, a lot of this logic can be simplified now. Just did. Sorry about that

PR #566, opened 2025-06-30
feat: hide banner if missing data

colton-demetriou:
Screenshot 2025-06-30 at 12 30 24 PM
Screenshot 2025-06-30 at 12 30 35 PM

### DIFF
--- a/packages/visual-editor/src/components/pageSections/Banner.tsx
+++ b/packages/visual-editor/src/components/pageSections/Banner.tsx
@@ -86,19 +86,21 @@ const BannerComponent = ({ data, styles }: BannerSectionProps) => {
   }[styles.textAlignment];
 
   return (
-    <PageSection
-      background={styles.backgroundColor}
-      verticalPadding="sm"
-      className={`flex ${justifyClass} items-center`}
-    >
-      <EntityField
-        displayName={pt("fields.bannerText", "Banner Text")}
-        fieldId={data.text.field}
-        constantValueEnabled={data.text.constantValueEnabled}
+    resolvedText && (
+      <PageSection
+        background={styles.backgroundColor}
+        verticalPadding="sm"
+        className={`flex ${justifyClass} items-center`}
       >
-        <Body>{resolvedText}</Body>
-      </EntityField>
-    </PageSection>
+        <EntityField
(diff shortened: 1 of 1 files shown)


packages/visual-editor/src/components/pageSections/Banner.tsx
```diff
@@ -86,19 +86,21 @@ const BannerComponent = ({ data, styles }: BannerSectionProps) => {
...
-    <PageSection
-      background={styles.backgroundColor}
-      verticalPadding="sm"
-      className={`flex ${justifyClass} items-center`}
-    >
-      <EntityField
-        displayName={pt("fields.bannerText", "Banner Text")}
-        fieldId={data.text.field}
-        constantValueEnabled={data.text.constantValueEnabled}
+    resolvedText && (
```
benlife5, line 89: This works but I'm wondering if we should use the VisibilityWrapper functionality because if there's no data this will also be invisible in the editor drag-and-drop

mkilpatrick: Not a bad idea. But it wouldn't be 100% clear that it's invisible due to the visibility prop vs there's no data on the particular entity. Start a thread please.

benlife5: Oh true, I forgot that the visibility prop applies to all instances

benlife5 (dismissed):

packages/visual-editor/src/components/pageSections/Banner.tsx
colton-demetriou: https://yext.slack.com/archives/C06A06BCUUF/p1751309460125069?thread_ts=1751304107.414609&cid=C06A06BCUUF

mkilpatrick, line 89: Might be simpler to do

```
if (!resolvedText) { return <></>; }
return <PageSection...
```

Then we can do the same in a bunch of other places eventually. Though maybe that changes if we still need to return the full thing wrapped in some Invisible component

mkilpatrick (approved):

benlife5 (approved):

briantstephan (approved):

PR #569, opened 2025-07-01
feat: add Open Now button to locator

Fondryext:
This PR adds the ability to add an 'Open Now' button to the
Locator component. If enabled, and if clicked, a filter is
applied to only show the currently open results.

To accomplish this, a new component atom, toggle, has also been added. For now it has
no props, and always uses the secondary color when toggled.
J=WAT-4908
TEST=manual

Ran in interactive dev mode and saw all expected behavior.
Reference the Jam video
https://jam.dev/c/bb97c575-2e20-4d9e-83c0-0e9594e584c0

### DIFF
--- a/packages/visual-editor/package.json
+++ b/packages/visual-editor/package.json
@@ -77,7 +77,6 @@
     "lucide-react": "^0.414.0",
     "lz-string": "1.5.0",
     "next-themes": "^0.3.0",
-    "pack": "^2.2.0",
     "picocolors": "^1.0.1",
     "prompts": "^2.4.2",
     "pure-react-carousel": "^1.32.0",
--- a/packages/visual-editor/package.json
+++ b/packages/visual-editor/package.json
@@ -66,6 +66,7 @@
     "@radix-ui/react-separator": "^1.1.7",
     "@radix-ui/react-slot": "^1.1.0",
     "@radix-ui/react-switch": "^1.1.0",
+    "@radix-ui/react-toggle": "^1.1.9",
     "@radix-ui/react-tooltip": "^1.1.2",
     "@tanstack/react-query": "^5.71.5",
     "class-variance-authority": "^0.7.0",
@@ -76,6 +77,7 @@
     "lucide-react": "^0.414.0",
     "lz-string": "1.5.0",
     "next-themes": "^0.3.0",
+    "pack": "^2.2.0",
     "picocolors": "^1.0.1",
     "prompts": "^2.4.2",
     "pure-react-carousel": "^1.32.0",
@@ -141,8 +143,8 @@
   },
   "peerDependencies": {
     "@yext/pages-components": "^1.1.4",
-    "@yext/search-headless-react": "^2.5.2",
-    "@yext/search-ui-react": "^1.8.13",
+    "@yext/search-headless-react": "^2.5.3",
+    "@yext/search-ui-react": "^1.8.15",
     "mapbox-gl": "^2.9.2",
     "react": "^17.0.2 || ^18.2.0",
     "react-dom": "^17.0.2 || ^18.2.0"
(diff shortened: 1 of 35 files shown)


packages/visual-editor/src/components/Locator.tsx
```diff
@@ -47,14 +50,17 @@ import {
 import { MapPinIcon } from "./MapPinIcon.js";
 import { FaAngleRight } from "react-icons/fa";
 import { formatPhoneNumber } from "./atoms/phone.js";
+import { YextField } from "../editor";
```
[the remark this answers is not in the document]
Fondryext: oops auto import got me

packages/visual-editor/src/components/Locator.tsx
```diff
@@ -305,6 +320,41 @@ const LocatorInternal: React.FC<LocatorProps> = (props) => {
...
+  const handleOpenNowClick = (selected: boolean) => {
+    searchActions.setFilterOption({
+      filter: {
+        kind: "fieldValue",
+        fieldId: HOURS_FIELD,
+        matcher: Matcher.OpenAt,
+        value: "now",
+      },
+      selected,
+      displayName: "Open Now",
```
benlife5, line 333: Not sure what this is used for, but might need a translation?

packages/visual-editor/src/components/atoms/toggle.tsx
```diff
@@ -0,0 +1,26 @@
...
+import * as TogglePrimitive from "@radix-ui/react-toggle";
+import { themeManagerCn } from "@yext/visual-editor";
+
+function Toggle({
+  className,
+  ...props
+}: React.ComponentProps<typeof TogglePrimitive.Root>) {
+  return (
+    <TogglePrimitive.Root
+      data-slot="toggle"
```
benlife5, line 13: probably want to respect textTransform too, unfortunately it can't be set via tailwind
```
        style={{
          // @ts-expect-error ts(2322) the css variable here resolves to a valid enum value
          textTransform: "var(--textTransform-button-textTransform)",
        }}
```

packages/visual-editor/src/components/Locator.tsx
```diff
@@ -89,6 +94,16 @@ const locatorFields: Fields<LocatorProps> = {
       },
     ],
   }),
+  openNowButton: YextField(
+    msg("fields.options.includeOpenNow", "Include Open Now Button"),
+    {
+      type: "radio",
+      options: [
+        { label: msg("yes", "Yes"), value: true },
```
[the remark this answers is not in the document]
Fondryext: Oh, gotcha! I did it like this because the same lines are in NearbyLocations and DefaultThemeConfig. Would you like me to update those as well, or are they different for some other reason?

packages/visual-editor/package.json
```diff
@@ -76,6 +77,7 @@
     "lucide-react": "^0.414.0",
     "lz-string": "1.5.0",
     "next-themes": "^0.3.0",
+    "pack": "^2.2.0",
```
benlife5, line 80: Where does pack get used?

packages/visual-editor/src/components/atoms/toggle.tsx
Fondryext: Thanks!

packages/visual-editor/src/components/Locator.tsx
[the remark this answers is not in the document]
Fondryext: Done!

packages/visual-editor/package.json
Fondryext: That's really odd, I have no recollection of adding this. Maybe I ran something wrong with pack while testing. Removed!

asanehisa (approved):

benlife5 (approved):

asanehisa (changes requested):
Btw we do have migrations code for adding new fields to components.
See https://github.com/yext/visual-editor/tree/main/packages/visual-editor/src/components/migrations

edit: oops didn't mean to make this block the PR
Just a btw without adding migrations, the value for openNowButton is "undefined" for existing users. Default props would only apply for new Locators. (well I guess u guys didn't set any defaultProps)

asanehisa (approved):

benlife5:
Seems like you can't add a scope and use it in the same PR, I've removed the scope for now but it will be usable going forward

PR #571, opened 2025-07-01
fix: reduce fonts imported into generated pages

asanehisa:
Before:
Screenshot 2025-07-01 at 2 26 27 PM
Would inject every single font option

Now: 
Screenshot 2025-07-01 at 2 26 17 PM
Only injects fonts being used in the theme

Confirmed on site [here](https://www.yext.com/s/4412098/yextsites/158910/pagesets) that live pages have correct style tags and Layout Editor / Theme Editor work. Theme Editor also has all the font options still and changing fonts works as expected.

### DIFF
--- a/packages/visual-editor/src/utils/applyTheme.ts
+++ b/packages/visual-editor/src/utils/applyTheme.ts
@@ -27,16 +27,15 @@ const devLogger = new DevLogger();
 export const applyTheme = (
   document: Document,
   themeConfig: ThemeConfig,
-  base?: string,
-  inEditor?: boolean
+  base?: string
 ): string => {
   devLogger.logFunc("applyTheme");
 
   const publishedTheme = document?.__?.theme;
   const overrides = publishedTheme ? JSON.parse(publishedTheme) : undefined;
 
-  // In the editor inject all the google font tags, else only inject the in-use font tags
-  const fontLinkTags = inEditor
+  // In the editor with no published theme, inject all the google font tags, else only inject the in-use font tags
+  const fontLinkTags = !publishedTheme
     ? googleFontLinkTags
     : constructGoogleFontLinkTags(
         extractInUseFontFamilies(overrides, defaultFonts)
--- a/packages/visual-editor/src/utils/applyTheme.ts
+++ b/packages/visual-editor/src/utils/applyTheme.ts
@@ -5,10 +5,10 @@ import {
   constructGoogleFontLinkTags,
   defaultFonts,
   extractInUseFontFamilies,
-  googleFontLinkTags,
 } from "./visualEditorFonts.ts";
 import { ThemeConfig } from "./themeResolver.ts";
 import { hexToHSL } from "./colors.ts";
+import { googleFontLinkTags } from "./visualEditorFonts";
 
 export type Document = {
   [key: string]: any;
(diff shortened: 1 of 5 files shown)


packages/visual-editor/src/utils/visualEditorFonts.ts
```diff
@@ -217,3 +218,37 @@ const filterFontWeights = (
...
+  // Iterate over all the keys in the theme data to find font names.
+  for (const key in data) {
+    // searches for keys with "fontFamily" like "--fontFamily-h1-fontFamily"
+    if (typeof key === "string" && key.includes("fontFamily")) {
+      const value = data[key];
+      // key / value looks like "--fontFamily-h1-fontFamily": "'Open Sans', sans-serif"
+      // parses fontName from the value
+      if (typeof value === "string" && value.length > 0) {
+        const firstFont = value.split(",")[0];
+        const cleanedFontName = firstFont.trim().replace(/^['"]|['"]$/g, "");
```
mkilpatrick, line 238: Do we see any risk with this string parsing?

packages/visual-editor/src/utils/visualEditorFonts.ts
```diff
@@ -217,3 +218,37 @@ const filterFontWeights = (
...
+        fontFamilies.add(cleanedFontName);
+      }
+    }
+  }
+
+  const inUseFonts: FontRegistry = {};
+
+  // For each unique font family found, look it up in the avaliableFonts map.
+  for (const fontName of fontFamilies) {
+    if (avaliableFonts[fontName]) {
```
mkilpatrick, line 248: Maybe as a catch to the parsing comment above, if we find a font not in the registry then we console warn

packages/visual-editor/src/utils/visualEditorFonts.ts
benlife5: Yeah I think also long as we check that the parsed names actually exist, we're ok. Will probably need some updates when we allow customer-uploaded fonts but I don't think there's anything we can do for that now

packages/visual-editor/src/utils/applyTheme.ts
```diff
@@ -29,8 +34,12 @@ export const applyTheme = (
   const publishedTheme = document?.__?.theme;
   const overrides = publishedTheme ? JSON.parse(publishedTheme) : undefined;
 
+  const fontLinkTags = constructGoogleFontLinkTags(
+    extractInUseFontFamilies(overrides, defaultFonts)
+  );
+
   if (Object.keys(themeConfig).length > 0) {
-    return `${base ?? ""}${googleFontLinkTags}<style id="${THEME_STYLE_TAG_ID}" type="text/css">${internalApplyTheme(overrides, themeConfig)}</style>`;
+    return `${base ?? ""}${fontLinkTags}<style id="${THEME_STYLE_TAG_ID}" type="text/css">${internalApplyTheme(overrides, themeConfig)}</style>`;
```
benlife5, line 42: I might be missing it but I think we need something (maybe in `updateThemeInEditor`) to add them all for the edit page. I think that's the only way we can have the font selector dropdown render the options in their own font

asanehisa: Oh oddly the Theme Editor page seems to have them all in the dropdown w/ their fonts? 

Screenshot 2025-07-01 at 3 43 01 PM

packages/visual-editor/src/utils/visualEditorFonts.ts
asanehisa: I agree that it seems to be okay / safe for now until we add customer-uploaded fonts (then we'll wanna update tests too)

packages/visual-editor/src/utils/applyTheme.ts
asanehisa: But maybe some fonts r handled differently than others? Will take a look

asanehisa: Yup! Fixed!

packages/visual-editor/src/utils/applyTheme.ts
```diff
@@ -27,16 +27,20 @@ const devLogger = new DevLogger();
 export const applyTheme = (
   document: Document,
   themeConfig: ThemeConfig,
-  base?: string
+  base?: string,
+  inEditor?: boolean
```
mkilpatrick, line 31: This is going to be breaking (or at least no honored) for existing non-artifact sites, such as the hybrid RBC site. We could tell them about needing to update this, but is there a way we could automatically determine it?

asanehisa: Hmm yeah let me try something else

asanehisa: Confirmed it works! Hmm let me know if we'd prefer I use document._pageset.codeTemplate, I'm not super familiar with hybrid developer works flow. Maybe @benlife5 ?

mkilpatrick (approved):

benlife5 (approved):

PR #574, opened 2025-07-02
feat: add non-link analytics to FAQ section and mobile header

benlife5:
Confirmed the analytics events made it to snowflake, will check in report builder later. 

I'll also add to the Reviews component once it is ready, in the meantime I'll keep the JIRA item open

Screenshot 2025-07-02 at 3 28 16 PM
Screenshot 2025-07-02 at 3 26 55 PM

### DIFF
--- a/packages/visual-editor/src/components/header/Header.tsx
+++ b/packages/visual-editor/src/components/header/Header.tsx
@@ -4,6 +4,7 @@ import {
   AnalyticsScopeProvider,
   CTA as CTAType,
   ComplexImageType,
+  useAnalytics,
 } from "@yext/pages-components";
 import { ComponentConfig, Fields } from "@measured/puck";
 import {
@@ -108,6 +109,7 @@ interface HeaderLayoutProps {
 const HeaderLayout = (props: HeaderLayoutProps) => {
   const { t } = useTranslation();
   const [menuOpen, setMenuOpen] = React.useState(false);
+  const analytics = useAnalytics();
   const {
     logo,
     logoWidth,
@@ -147,7 +149,19 @@ const HeaderLayout = (props: HeaderLayoutProps) => {
             </EntityField>
             <button
               className="flex md:hidden ml-auto my-auto"
-              onClick={() => setMenuOpen(!menuOpen)}
+              onClick={() => {
+                analytics?.track({
+                  action: menuOpen ? "COLLAPSE" : "EXPAND",
+                  eventName: "toggleMobileMenu",
+                });
+                setMenuOpen(!menuOpen);
+              }}
+              data-ya-action={
+                analytics?.getDebugEnabled() ? "COLLAPSE/EXPAND" : undefined
+              }
+              data-ya-eventname={
+                analytics?.getDebugEnabled() ? `toggleMobileMenu` : undefined
+              }
               aria-label={
                 menuOpen
                   ? t("closeHeaderMenu", "Close header menu")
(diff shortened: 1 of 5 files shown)


packages/visual-editor/src/components/header/Header.tsx
```diff
@@ -147,7 +149,19 @@ const HeaderLayout = (props: HeaderLayoutProps) => {
             </EntityField>
             <button
               className="flex md:hidden ml-auto my-auto"
-              onClick={() => setMenuOpen(!menuOpen)}
+              onClick={() => {
+                analytics?.track({
```
mkilpatrick, line 153: So scope is already accounted for automatically?

packages/visual-editor/src/components/pageSections/FAQsSection.tsx
```diff
@@ -147,7 +153,26 @@ const FAQsSectionComponent: React.FC<FAQSectionProps> = ({ data, styles }) => {
...
+              <AccordionItem
+                key={index}
+                data-ya-action={
+                  analytics?.getDebugEnabled() ? "COLLAPSE/EXPAND" : undefined
+                }
+                data-ya-eventname={
+                  analytics?.getDebugEnabled() ? `toggleFAQ${index}` : undefined
+                }
+                onToggle={(e) =>
+                  e.currentTarget.open
```
mkilpatrick, line 165: Just verifying that open is the current state and toggling is the state it's about to be in, so that's why if open then we say action collapse?

packages/visual-editor/src/components/header/Header.tsx
```diff
@@ -147,7 +149,19 @@ const HeaderLayout = (props: HeaderLayoutProps) => {
...
+                  action: menuOpen ? "COLLAPSE" : "EXPAND",
+                  eventName: "toggleMobileMenu",
+                });
+                setMenuOpen(!menuOpen);
+              }}
+              data-ya-action={
+                analytics?.getDebugEnabled() ? "COLLAPSE/EXPAND" : undefined
```
mkilpatrick, line 160: EXPAND/COLLAPSE is usually the phrase so I'd swap them, same for FAQs

packages/visual-editor/src/components/header/Header.tsx
```diff
@@ -147,7 +149,19 @@ const HeaderLayout = (props: HeaderLayoutProps) => {
...
               className="flex md:hidden ml-auto my-auto"
-              onClick={() => setMenuOpen(!menuOpen)}
+              onClick={() => {
+                analytics?.track({
+                  action: menuOpen ? "COLLAPSE" : "EXPAND",
+                  eventName: "toggleMobileMenu",
+                });
+                setMenuOpen(!menuOpen);
+              }}
+              data-ya-action={
```
mkilpatrick, line 159: Too bad we don't do this automatically so someone would need to know to do this. Nbd though.

packages/visual-editor/src/components/pageSections/FAQsSection.tsx
benlife5: ah good catch, updated

packages/visual-editor/src/components/header/Header.tsx
benlife5: Yep, `analytics.track` from `pages-components` handles scope and conversion value and passes everything else along

jwartofsky-yext (approved):

PR #579, opened 2025-07-07
fix: fix default_directory in /starter

jwartofsky-yext:
The field on the account is "defaultdirectory", not "default_directory"

Strangely the DM ID is "defaultDirectory"


starter/localData/dev-location-stream__en__3bafed68a7dd4323de9e996a92b0206c.json
```diff
@@ -0,0 +1,371 @@
+{
+  "__": {
+    "entityPageSet": {},
+    "name": "dev-location"
+  },
+  "_site": null,
```
benlife5, line 6: These new ones remove a lot of the fields we have manually added to the test data

PR #583, opened 2025-07-08
feat(components): add card styles

asanehisa:
This added card heading level to relevant components. 

Checked with aaron and we wanted it look like: 
```
styles: {
  cards: {
     backgroundColor: ...
     headingLevel: ....
  }
}
```

And we want the card heading level to default to h3. 

Added migration and tests.

### DIFF
--- a/packages/visual-editor/src/components/migrations/0007_add_card_styles.ts
+++ b/packages/visual-editor/src/components/migrations/0007_add_card_styles.ts
@@ -1,0 +1,106 @@
+import { Migration } from "../../utils/migrate.ts";
+
+export const addCardStylesMigration: Migration = {
+  EventSection: {
+    action: "updated",
+    propTransformation: ({ styles, ...props }) => {
+      const { cardBackgroundColor, heading, ...otherStyles } = styles || {};
+      return {
+        styles: {
+          heading: heading,
+          cards: {
+            headingLevel: heading.level < 6 ? heading.level + 1 : 3,
+            backgroundColor: cardBackgroundColor,
+          },
+          ...otherStyles,
+        },
+        ...props,
+      };
+    },
+  },
+  InsightSection: {
+    action: "updated",
+    propTransformation: ({ styles, ...props }) => {
+      const { cardBackgroundColor, heading, ...otherStyles } = styles || {};
+      return {
+        styles: {
+          heading: heading,
+          cards: {
+            headingLevel: heading.level < 6 ? heading.level + 1 : 3,
+            backgroundColor: cardBackgroundColor,
+          },
+          ...otherStyles,
+        },
+        ...props,
+      };
+    },
+  },
+  NearbyLocationsSection: {
+    action: "updated",
+    propTransformation: ({ styles, ...props }) => {
+      const { cardBackgroundColor, cardHeadingLevel, ...otherStyles } =
+        styles || {};
+      return {
+        styles: {
+          cards: {
+            headingLevel: cardHeadingLevel,
+            backgroundColor: cardBackgroundColor,
+          },
+          ...otherStyles,
+        },
+        ...props,
+      };
+    },
+  },
+  ProductSection: {
+    action: "updated",
+    propTransformation: ({ styles, ...props }) => {
+      const { cardBackgroundColor, heading, ...otherStyles } = styles || {};
+      return {
+        styles: {
+          heading: heading,
+          cards: {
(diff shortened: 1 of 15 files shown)


packages/visual-editor/src/components/pageSections/EventSection.tsx
```diff
@@ -148,14 +159,7 @@ const EventCard = ({
...
         {event.title && (
-          <Heading
-            level={6}
-            semanticLevelOverride={
-              sectionHeadingLevel < 6
-                ? ((sectionHeadingLevel + 1) as HeadingLevel)
-                : "span"
-            }
-          >
+          <Heading level={cardStyles.headingLevel}>
```
benlife5, line 162: Check with Aaron I guess but I think we still need to do the semanticLevelOverride stuff. I think we should style based on what the user inputs but we need the subheadings to be one level less important than the section headings for WCAG to pass

packages/visual-editor/src/components/pageSections/EventSection.test.tsx
```diff
@@ -157,7 +157,7 @@ const tests: ComponentTest[] = [
           bgColor: "bg-palette-primary-light",
           textColor: "text-black",
         },
-        headingLevel: 5,
+        headingLevel: 2,
```
benlife5, line 160: Why were the old tests updated?

packages/visual-editor/src/components/pageSections/TestimonialSection.test.tsx
```diff
@@ -200,6 +200,57 @@ const tests: ComponentTest[] = [
       expect(page.getByText("Jan 1, 2025")).toBeVisible();
     },
   },
+  {
+    name: "version 7 props with constant value",
```
benlife5, line 204: should probably have an entity value test too

packages/visual-editor/src/components/migrations/0007_add_card_styles.ts
```diff
@@ -0,0 +1,106 @@
+import { Migration } from "../../utils/migrate.ts";
+
+export const addCardStylesMigration: Migration = {
```
benlife5, line 3: Directory migration too

packages/visual-editor/src/components/pageSections/EventSection.test.tsx
asanehisa: oops was an accident for this one. For the ones with headingLevel 6, I did it so the WCAG tests would pass. But yeah I'll ask aaron about the semantic level override.

packages/visual-editor/src/components/migrations/0007_add_card_styles.ts
asanehisa: Good catch! Thanks!

packages/visual-editor/src/components/pageSections/EventSection.tsx
asanehisa: Aaron said that for now, we won't do anything special + just use whatever values the user selects. If their choice fails WCAG, its their fault. In the future, maybe we'll add a warning system / other handling.

packages/visual-editor/src/components/pageSections/TestimonialSection.tsx
```diff
@@ -34,11 +34,14 @@ export interface TestimonialSectionProps {
   };
   styles: {
     backgroundColor?: BackgroundStyle;
-    cardBackgroundColor?: BackgroundStyle;
     heading: {
       level: HeadingLevel;
```
jwartofsky-yext, line 38: Is this headingLevel used anywhere? There's now 2

benlife5 (dismissed):

jwartofsky-yext (dismissed):

packages/visual-editor/src/components/pageSections/EventSection.tsx
asanehisa: Hmm so you think we should use semanticLevelOverride ? aka when a user makes card heading level 5, the heading level isn't actually changed but the styles are?

mkilpatrick: Yea. I feel like the heading levels are more styles that get applied over the semantic html tags that are pre-defined. I would say that should be true globally for heading levels used anywhere. If you start a thread I'm happy to chime in.

jwartofsky-yext (approved):

PR #585, opened 2025-07-08
feat: add backgroundColor to Breadcrumbs and Directory

jwartofsky-yext:
Adds backgroundColor prop to breadcrumbs and directory components

Directory also has "breadcrumbsBackgroundColor", which gets passed to the internal Breadcrumbs component.

image


benlife5 (commented):
LGTM. Do we want/need a migration? Might not be necessary if the background falls back to white but we might want to backfill the default value

packages/visual-editor/src/components/migrations/0007_add_breadcrumbs_directory_background.ts
```diff
@@ -0,0 +1,27 @@
+import { Migration } from "../../utils/migrate.ts";
```
benlife5, line 1: Add version 7 props to the Breadcrumbs/Directory tests too

asanehisa: Oops I think its version 8 now

jwartofsky-yext: Done

mkilpatrick (approved):

PR #587, opened 2025-07-09
fix: re-add directory card border

benlife5:
Got lost in a recent PR; noticed while reviewing the new screenshots


asanehisa (approved):
Oops sorry that was probably me. Thanks for catching!

PR #598, opened 2025-07-09
chore: clean up header/footer image styling

colton-demetriou:
Screenshot 2025-07-09 at 3 14 17 PM
I can revert the default change, but I thought this looks better than it going all the way to the right side of the page. (Centering the items also fixes this, but then if you increase image size it grows towards the left side of the page instead of to the right).

### DIFF
--- a/packages/visual-editor/src/components/header/ExpandedHeader.tsx
+++ b/packages/visual-editor/src/components/header/ExpandedHeader.tsx
@@ -314,8 +314,8 @@ const ExpandedHeaderWrapper: React.FC<ExpandedHeaderProps> = ({
             displayName={pt("fields.logoUrl", "Logo")}
           >
             <HeaderLogo
-              logo={buildComplexImage(logo, logoStyle.width || 100)}
-              logoWidth={logoStyle.width || 100}
+              logo={buildComplexImage(logo, logoStyle.width ?? 200)}
+              logoWidth={logoStyle.width ?? 200}
               aspectRatio={logoStyle.aspectRatio}
             />
           </EntityField>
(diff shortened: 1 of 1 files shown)


packages/visual-editor/src/components/migrations/0009_update_expanded_components_image_styling.ts
```diff
@@ -0,0 +1,55 @@
+import { Migration } from "../../utils/migrate.ts";
```
mkilpatrick, line 1: Don't need a migration because no one has used the component yet

packages/visual-editor/src/components/header/ExpandedHeader.tsx
```diff
@@ -619,8 +612,10 @@ export const ExpandedHeader: ComponentConfig<ExpandedHeaderProps> = {
     },
     styles: {
       primaryHeader: {
-        logoWidth: 100,
-        aspectRatio: 2,
+        logo: {
+          width: undefined,
```
benlife5, line 616: This can stay at width: 100 and aspect 2:1 per the original [spec](https://docs.google.com/document/d/1--1VVtcxCYbZE-FsAM99xH-2jOAPe9L8rjaqpfqLKho/edit?tab=t.0#heading=h.2alb408nn0jg)
(same for the footer logo and width:60 aspect 1:1 for the footer utility images)

benlife5 (approved):
LGTM if we're good with the default values

colton-demetriou:
> LGTM if we're good with the default values

Aaron said it looks good 👍

briantstephan (approved):

PR #599, opened 2025-07-09
feat: add theme Image section with BorderRadius

jwartofsky-yext:
Adds Image category in Theme manager with an option for Border Radius

Applies to Hero, Promo, and Photo Gallery components only

image


packages/visual-editor/src/editor/BorderRadiusSelector.tsx
```diff
@@ -90,7 +108,10 @@ export const convertDefaultBorderRadiusToOptions = (
   });
 };
 
-export const BorderRadiusSelector = (label?: string): Field => {
+export const BorderRadiusSelector = (
+  label?: string,
+  borderRadiusOptions?: BorderRadiusOption[]
```
mkilpatrick, line 113: Is this passed in anywhere? Seems weird to have the fallback on line 138

benlife5: `BorderRadiusSelector` isn't actually used anywhere, it's leftover from when we had more styling in the Layout editor. I guess this is because there's now two sets of border radius options. 

Is `imageBorderRadiusOptions` used anywhere except being passed to `getBorderRadiusOptions`? If not, seems like we could just have `imageBorderRadiusOptions` already transformed into the output of `getBorderRadiusOptions(imageBorderRadiusOptions)`

mkilpatrick (approved):

benlife5 (approved):

PR #606, opened 2025-07-10
fix: text wrapping for footer links

benlife5:
Screenshot 2025-07-10 at 1 07 37 PM
Screenshot 2025-07-10 at 1 07 33 PM
Screenshot 2025-07-10 at 1 01 37 PM
Screenshot 2025-07-10 at 1 01 32 PM

The entity field tooltip was also causing some boxing issues so I moved it outside all the links, which I think is fine because it just says "static content"


jwartofsky-yext (approved):
lgtm

mkilpatrick (approved):

PR #612, opened 2025-07-10
chore: more Directory and PageSection tests

benlife5:
Created new item for the Directory overflow issue: https://yexttest.atlassian.net/browse/SUMO-7543
I think the FAQSection was updated as part of one of our breaking changes, so I've removed the version 0 prop test


benlife5:
Will look into why the FAQ section header is dark tomorrow, may be because we had a few breaking changes even after introducing migrations. Also will look to surface the wcag failure better in the action

packages/visual-editor/src/components/pageSections/TestimonialSection.test.tsx
```diff
@@ -292,6 +292,8 @@ describe("TestimonialSection", async () => {
         puckConfig
       );
 
+      console.log(data);
```
asanehisa, line 295: rm

packages/visual-editor/src/components/pageSections/FAQsSection.test.tsx
```diff
@@ -47,33 +58,35 @@ const tests: ComponentTest[] = [
     version: migrationRegistry.length,
   },
   {
-    name: "version 0 props with entity values",
+    name: "version 10 props with entity values",
```
asanehisa, line 61: Oh is there a reason we're replacing v0 with v10 instead of making another test?

packages/visual-editor/src/components/pageSections/CoreInfoSection.test.tsx
```diff
@@ -398,6 +398,271 @@ const tests: ComponentTest[] = [
...
+          },
+          hours: {
+            constantValue: {},
+            field: "hours",
+          },
+        },
+        info: {
+          address: {
+            constantValue: {
+              city: "",
```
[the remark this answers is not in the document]
benlife5: You can't enter a constant value for address in the editor so the empty address is just to satisfy TS

packages/visual-editor/src/components/Directory.test.tsx
```diff
@@ -17,121 +17,157 @@ import {
 import { Render, Config } from "@measured/puck";
 import { page } from "@vitest/browser/context";
```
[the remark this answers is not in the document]
benlife5: The rendering is the same for root/country/region so I think it's ok to just test Cards (city) vs. List (region/country/root) but I can add tests for those if you think that's important

mkilpatrick (approved):

packages/visual-editor/src/components/testing/screenshots/Directory/[mobile] default props - city list - document data.png
Comment (mkilpatrick) [inline packages/visual-editor/src/components/testing/screenshots/Directory/[mobile] default props - city list - document data.png:1]: Interesting that the bottom gets cut off

mkilpatrick (approved):

packages/visual-editor/src/components/testing/screenshots/Directory/[mobile] default props - city list - document data.png
Comment (benlife5) [reply packages/visual-editor/src/components/testing/screenshots/Directory/[mobile] default props - city list - document data.png:1]: Yeah I'm a little confused by that too. It might be that the screenshot is using the full content height but the render is using the specified viewport height of 667. Maybe worth a follow up item later

packages/visual-editor/src/components/testing/screenshots/Directory/[desktop] default props - city list - document data.png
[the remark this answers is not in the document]
Comment (benlife5) [reply packages/visual-editor/src/components/testing/screenshots/Directory/[desktop] default props - city list - document data.png:1]: I think there's no wrapping on the link and so it overflows the grid. I tried to fix it briefly but wasn't successful, created an item here: https://yexttest.atlassian.net/browse/SUMO-7543

asanehisa (approved):

PR #615, opened 2025-07-11
fix: use shared getPath for Directory/Locator/NearbyLocations links

benlife5:
test site: https://www.yext.com/s/4259018/yextsites/159383/pagesets?focus=locations


packages/visual-editor/src/utils/getLocationPath.ts
```diff
@@ -0,0 +1,31 @@
...
+): string | undefined => {
+  if (!location || (!location.slug && !location.address && !location.id)) {
+    return;
+  }
+
+  if (location.slug) {
+    return `${relativePrefixToRoot ?? ""}${location.slug}`;
+  }
+
+  const locale = document.locale;
```
mkilpatrick, line 23: Should this use the useTranslation like a recent comment you had on another PR?

packages/visual-editor/src/utils/getLocationPath.ts
```diff
@@ -0,0 +1,31 @@
+import { normalizeSlug } from "./slugifier.ts";
+import { AddressType } from "@yext/pages-components";
+
+interface LocationDocument extends Record<string, any> {
+  slug?: string;
```
mkilpatrick, line 5: Don't we need the locale of the thing we're directing to, not the locale of the current document? Theoretically they should be the same, but for correctness.

benlife5: https://yext.slack.com/archives/C02985ZF96Y/p1752247373835579
I think they should always be the same, but I made it a bit more generic so we can adjust as necessary

benlife5:
Still working on testing these changes

@colton-demetriou the create-dev-release flow didn't seem to work on the initial commit. And is there a way to run it again on subsequent commits?

mkilpatrick (dismissed):

mkilpatrick (dismissed):

jwartofsky-yext (approved):
lgtm

PR #618, opened 2025-07-11
feat: copy/paste component data

mkilpatrick:
Allows users to copy the prop data of one component into another component of the same type. This makes it easier to copy the header/footer data across page groups.

https://github.com/user-attachments/assets/b7638e5b-1b0a-4e0f-8fb1-3bac137f7700


packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
```diff
@@ -55,6 +61,7 @@ export const InternalLayoutEditor = ({
     useState<boolean>(false);
   const historyIndex = useRef<number>(0);
   const { i18n } = usePlatformTranslation();
+  const { t } = useTranslation();
```
benlife5, line 64: This should be `pt` from `usePlatformTranslation()` since it's editor-facing

benlife5:
^ image loading is still a little flaky

asanehisa (approved):

PR #621, opened 2025-07-14
chore: expanded footer tests

benlife5:
- moves the footer files to a footer folder (like header)
- fixes an ARIA landmark issue flagged by the WCAG tests
- adds screenshot tests for expanded footer


packages/visual-editor/src/components/testing/screenshots/ExpandedFooter/[desktop] version 10 props - expanded.png
Comment (asanehisa) [inline packages/visual-editor/src/components/testing/screenshots/ExpandedFooter/[desktop] version 10 props - expanded.png:1]: Should footer in the screenshots be at the bottom of the page? Unsure how it works with our current testing lib

Comment (benlife5) [reply packages/visual-editor/src/components/testing/screenshots/ExpandedFooter/[desktop] version 10 props - expanded.png:1]: oh I think the test container doesn't have the height property we added to the root zone, but that's going to change in the puck upgrade so I think I'll fix it then

asanehisa (approved):

mkilpatrick (approved):

PR #630, opened 2025-07-15
feat: normalize locales

jwartofsky-yext:
Normalizes locales coming from the document and from the content endpoints.


packages/visual-editor/src/components/header/languageDropdown.tsx
```diff
@@ -231,13 +239,13 @@ export function parseDocumentForLanguageDropdown(
   }
 
   let contentEndpointId: string = "";
-  let locales: string[] = [];
+  let locales: string[] | undefined = undefined;
```
mkilpatrick, line 242: Why undefined vs empty now?

packages/visual-editor/src/utils/normalizeLocale.ts
```diff
@@ -0,0 +1,24 @@
+export function normalizeLocale(locale: string): string {
```
mkilpatrick, line 1: Third place we have this now 😢. Can everything be passed down pre-normalized?

packages/visual-editor/src/components/header/languageDropdown.tsx
jwartofsky-yext: Because that is what it actually was before

There's a check a few lines down if (!locales) then error

That check won't trigger if locales is []

But it will if it is undefined, which it would be if pageset.scope.locales was undefined

packages/visual-editor/src/utils/normalizeLocale.ts
jwartofsky-yext: Would have to ask Spruce

the locales on the document come from _pageset, which gets injected by cog somewhere

From the other thread, it seems the locales come from profilequeryserver somewhere, so we may have to get KG to fix it at the source

For now I'm trying to just normalize everything that comes into VE, which unfortunately means a lot of normalizing

Ideally all of Yext would just use the correct format :)

packages/visual-editor/src/editor/Editor.tsx
```diff
@@ -106,7 +107,9 @@ export const Editor = ({
 
   useEffect(() => {
     if (templateMetadata?.platformLocale) {
-      i18nPlatformInstance.changeLanguage(templateMetadata?.platformLocale);
+      i18nPlatformInstance.changeLanguage(
+        normalizeLocale(templateMetadata?.platformLocale)
```
mkilpatrick, line 111: Normalize at the outermost level - so do this directly on the templateMetadata when we receive the message.

packages/visual-editor/src/hooks/useDocument.tsx
```diff
@@ -24,7 +30,12 @@ const useDocument = <T,>(): T => {
     throw new Error("document does not exist on TemplateProps");
   }
 
-  return context.document as T;
+  const normalized = React.useMemo(
+    () => normalizeLocalesInObject(context.document),
```
mkilpatrick, line 34: Same

packages/visual-editor/src/utils/VisualEditorProvider.tsx
```diff
@@ -34,7 +35,9 @@ const VisualEditorProvider = <T extends Record<string, any>>({
   const queryClient = new QueryClient();
 
   if (templateProps?.document?.locale) {
-    i18nComponentsInstance.changeLanguage(templateProps.document.locale);
+    i18nComponentsInstance.changeLanguage(
+      normalizeLocale(templateProps.document.locale)
```
mkilpatrick, line 39: Won't need this

packages/visual-editor/src/utils/normalizeLocale.ts
```diff
@@ -0,0 +1,64 @@
...
+  try {
+    return JSON.parse(str);
+  } catch {
+    return undefined;
+  }
+}
+
+function tryNormalizeLocalesInString(str: string): string {
+  const parsed = tryParseJSON(str);
+  if (!parsed) return str;
```
mkilpatrick, line 60: Always use braces

packages/visual-editor/src/components/testing/screenshots/NearbyLocationsSection/[desktop] default props with multiple nearby locations.png
Comment (mkilpatrick) [inline packages/visual-editor/src/components/testing/screenshots/NearbyLocationsSection/[desktop] default props with multiple nearby locations.png:1]: Looks wrong

packages/visual-editor/src/utils/VisualEditorProvider.tsx
```diff
@@ -32,6 +33,7 @@ const VisualEditorProvider = <T extends Record<string, any>>({
   children,
 }: VisualEditorProviderProps<T>) => {
   const queryClient = new QueryClient();
+  templateProps = normalizeLocalesInObject(templateProps);
```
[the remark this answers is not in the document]
mkilpatrick: Your comment is that we don't need this but it's still here. Separately (but maybe related?), where do the locales that support the header language toggle come from? We want to make sure that works properly.

mkilpatrick (approved):

mkilpatrick (approved):

packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[desktop] version 1 props with partial entity values 3.png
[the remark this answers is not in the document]
Comment (jwartofsky-yext) [reply packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[desktop] version 1 props with partial entity values 3.png:1]: I think it's from https://github.com/yext/visual-editor/commit/1a6ecd8a35d69248a55d9f1e2282036f33cc6cc2#diff-85a487e868bb1d05dc8a589d4590fde85ad26d697a429a97a532f7476d2a4510R114

mkilpatrick (dismissed):

asanehisa (approved):

PR #632, opened 2025-07-16
docs: add autogenerated component docs 

asanehisa:
Now component docs can be autogenerated on changes to the tsdocs.

The image it uses is the first png in the test screenshots (if it finds an image for that component)

https://github.com/user-attachments/assets/eb4a0d83-ccf9-4bc0-866f-5db0c800fc14


packages/visual-editor/src/docs/components.md
```diff
@@ -0,0 +1,623 @@
+---
+title: Pre-Built Components
+outline: 2
+---
+
+## BannerSection
+
+The Banner Section component displays a single, translatable line of rich text. It's designed to be used as a simple, full-width banner on a page. Avaliable on Location templates.
+
+![Preview of the BannerSection component](../components/testing/screenshots/BannerSection/%5Bdesktop%5D%20default%20props%20with%20empty%20document.png)
```
mkilpatrick, line 10: Idk if there's anything you can do, but the screenshot is a full page so there's a bunch of whitespace, rather than just showing the component. Maybe there's a way our screenshot stuff can be adjusted later. Nothing to do here, just pointing out.

packages/visual-editor/src/docs/components.md
```diff
@@ -0,0 +1,623 @@
...
+
+![Preview of the CoreInfoSection component](../components/testing/screenshots/CoreInfoSection/%5Bdesktop%5D%20default%20props%20with%20document%20data.png)
+
+### Props
+
+#### Data Props
+
+This object contains all the content to be displayed within the three columns.
+
+| Prop            | Type                                                                                                                                                                                                                    | Description                           | Default |
```
mkilpatrick, line 87: Default not showing when it's an entity field

.github/workflows/autogenerate-docs.yml
```diff
@@ -0,0 +1,54 @@
...
+    defaults:
+      run:
+        working-directory: packages/visual-editor
+
+    permissions:
+      contents: write
+
+    strategy:
+      matrix:
+        node-version: [18.x, 20.x]
```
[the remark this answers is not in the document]
asanehisa: Oh yeah is a draft rn (figuring out workflow). Will mark when ready!

packages/visual-editor/tsdoc.json
```diff
@@ -0,0 +1,19 @@
...
+  "$schema": "https://developer.microsoft.com/json-schemas/tsdoc/v0/tsdoc.schema.json",
+  "extends": ["@microsoft/api-extractor/extends/tsdoc-base.json"],
+  "tagDefinitions": [
+    {
+      "tagName": "@propCategory",
+      "syntaxKind": "block",
+      "allowMultiple": true
+    },
+    {
+      "tagName": "@fakeTag",
```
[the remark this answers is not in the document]
asanehisa: Oh was for testing, am removing

packages/visual-editor/src/docs/components.md
```diff
@@ -0,0 +1,623 @@
+---
+title: Pre-Built Components
```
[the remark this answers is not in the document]
asanehisa: Hmmm I don't see it as a table in the rendered markdown?

packages/visual-editor/src/docs/components.md
asanehisa: Oh yeah in this case I'd have to further extract out each of these to give them defaultValues. Similar to this: https://github.com/yext/visual-editor/pull/632#discussion_r2211145640

I assumed we wouldn't wanna just keep extracting

packages/visual-editor/src/docs/components.md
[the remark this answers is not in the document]
mkilpatrick: Ah ok, it just showed up in the raw markdown view.

packages/visual-editor/src/components/Directory.tsx
```diff
@@ -25,18 +25,50 @@ import {
...
 
-export interface DirectoryProps {
-  data: {
-    directoryRoot: TranslatableString;
-  };
-  styles: {
+export interface DirectoryData {
+  /**
+   * The display label for the root link in the breadcrumbs navigation.
+   * @defaultValue "Directory Root" (constant)
```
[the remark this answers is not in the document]
asanehisa: Hmmm not with api-extractor since it is based off the dist/index.d.ts code.

packages/visual-editor/src/docs/components.md
```diff
@@ -0,0 +1,625 @@
...
```
mkilpatrick, line 2: Component Sections

mkilpatrick, line 2: In some future projects we're going to have exposed atoms specifically for the Core Info Section. In another for "global components" we may create some sort of component builder with atoms.

How hard would it be to put specific atoms we expose for the Core Info case under the Core Info section as subsections?  And then how hard would it be for atoms to show up in an atom.md doc? We could put the core info ones into atoms.md and somehow say that they're only allowed for core info if that ends up being easier, but at least reference that section from this file in the core info section section.

asanehisa: This would be do-able and would likely require we add a tsdoc tag like `@atoms `or something to be able to mark which components go in the atoms.md doc. Then I believe we could use the info built from `packages/visual-editor/src/components/registry/components.ts` (which is in `dist/registry/index.json`) to map which atoms should be linked to which components and build a atoms section for the CoreInfoSection (linked to the atoms.md). Would take a bit of adding code to the `generateMarkdown` func

pnpm-lock.yaml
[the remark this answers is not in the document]
asanehisa: Hmmm good question why things are downgraded. I ran `npx pnpm@9.12.3 i` to generate it and per the package.json `"packageManager": "pnpm@9.12.3",`

jwartofsky-yext (approved):
Nice!

packages/visual-editor/api-extractor.json
```diff
@@ -0,0 +1,36 @@
+{
+  "$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
+  "mainEntryPointFilePath": "dist/index.d.ts",
```
jwartofsky-yext, line 3: Would this work if you set it to index.ts instead of dist?

PR #634, opened 2025-07-16
chore: update registry/components.ts

asanehisa:


benlife5 (commented):
Good to merge this, although I think there's a number of things broken with the registry right now. TBH I haven't been spending time on it because we've been so focused on the no code flow, it seems a little superfluous

asanehisa:
> Good to merge this, although I think there's a number of things broken with the registry right now. TBH I haven't been spending time on it because we've been so focused on the no code flow, it seems a little superfluous

haha yaaa I only thought of it b/c of a comment duval mentioned in my other CR about us potentially using atoms in the future.

mkilpatrick (dismissed):

PR #645, opened 2025-07-17
fix: normalizeUndefined

jwartofsky-yext:


jwartofsky-yext:
I don't think this changes anything

It already handled undefined

PR #649, opened 2025-07-21
fix: name field on locator filters

k-gerner:
Now that we have updated the Search SDK to include an optional `name` property for the near-filter, we should include a name in the "hardcoded" filters that we're sending from the locator component.

J=[WAT-4942](https://yexttest.atlassian.net/browse/WAT-4942?atlOrigin=eyJpIjoiZGE3NTgyOWJiNWUyNGJkMGExM2JkOGM1ZTIyMDUwOWYiLCJwIjoiaiJ9)
TEST=manual
Used dev mode in Storm and saw the requests being fired with the `name` property present. Also saw expected `filterValue` value in the report builder


anguyen-yext2 (approved):

packages/visual-editor/src/components/Locator.tsx
```diff
@@ -246,6 +247,7 @@ const LocatorInternal: React.FC<LocatorProps> = (props) => {
                 lat: location.coords.latitude,
                 lng: location.coords.longitude,
                 radius: DEFAULT_RADIUS_METERS,
+                name: t("currentLocation", "Current Location"),
```
jwartofsky-yext, line 250: Do these filters work for different translations?

Fondryext (approved):

packages/visual-editor/src/components/Locator.tsx
```diff
@@ -267,8 +269,10 @@ const LocatorInternal: React.FC<LocatorProps> = (props) => {
                 lat: DEFAULT_MAP_CENTER[1],
                 lng: DEFAULT_MAP_CENTER[0],
                 radius: DEFAULT_RADIUS_METERS,
-                // TODO (kgerner): add name property once Search SDK updated
-                // name: "New York City, New York, NY",
+                name: t(
+                  "newYorkCity",
```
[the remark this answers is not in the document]
k-gerner: Not for testing. Mapbox would give us the translated version of NYC if we had searched in manually, so we want to be consistent

packages/visual-editor/src/components/Locator.tsx
k-gerner: Would you mind you clarifying your question?

asanehisa: I think the question is, if u translate "Current Location", does the filter still work?

jwartofsky-yext: Yes, that's the question ^

k-gerner: Yes it still works. The `name` field is only for internal display purposes in the Snowflake tables. The `lat`/`lng`/`radius` are the fields that actually affect the filter logic

PR #650, opened 2025-07-21
fix: add zh fallbacks for platform

jwartofsky-yext:
Moves the i18n fallback logic to its own file

Applies it to the platform i18n configuration aswell


asanehisa (commented):
ah not necessary but at this point the files could be in src/utils/i18n/....

asanehisa (dismissed):

PR #657, opened 2025-07-24
fix: use workspace for starter ve package

colton-demetriou:
confirmed staxkblitz and local dev both work (made local updates and made sure they were reflected when running locally)


jwartofsky-yext (approved):
starter still works I assume?

PR #658, opened 2025-07-24
chore: upgrade to Puck 0.19.3

benlife5:
Slots
- Updates the root render method to not use a DropZone and instead apply the styles directly to the content
- Updates Flex and SectionContainer to use slots instead of dropzones
  - Grid was more complex when I first looked at it a month ago, will work with @Fondryext 

Data Model
- Updates `migrate` function to use `walkTree`
- Updates `InternalLayoutEditor` to respect root fields from the `ve.config`

Types
- Puck seems a little stricter about ComponentConfig types. We had some that were extensions of React.HTMLAttributes that were causing TS errors and it's a little unclear to me if that was a bug or not, but regardless we don't need to expose those props to the Puck component
- Puck now requires a CustomField type argument

Optimization
- Uses optimized versions of `usePuck`
  - `createUsePuck` is used when we want to react to Puck data changes, and uses a selector so that we only re-render when necessary
  - `useGetPuck` is used when we want to get Puck data but do not need to re-render 
- Uses `setData` in `dispatch`
- Only pass `data` and `ui` to layoutSaveState because Puck has some other internal data on `histories` now

Test site: https://www.yext.com/s/4259018/yextsites/159481/pagesets


packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
```diff
@@ -169,11 +173,11 @@ export const InternalLayoutEditor = ({
         alert("Failed to paste: Invalid component data.");
         return;
       }
-      // Update the component data in the specified zone
+
```
mkilpatrick, line 176: I can't comment on unchanged lines. Per my below comment we can move the getPuck stuff into here so nothing needs to be passed/called.

packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
```diff
@@ -169,11 +173,11 @@ export const InternalLayoutEditor = ({
...
       dispatch({
```
mkilpatrick, line 177: We can get the data with updated ids with walkTree. Btw, I had a DM with someone who took this code and updated things to work with 0.19.x which is where I'm getting this.

```
// If the pasted data has children, we need to generate new ids for them
      const newData = walkTree(pastedData, baseConfig, (contents) => (
        contents.map((item: ComponentDataOptionalId) => {
          const id = `${item.type}-${uuidv4()}`;
          return {
            ...item,
            props: { ...item.props, id },
          };
        })
      ))
```

packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
```diff
@@ -169,11 +173,11 @@ export const InternalLayoutEditor = ({
...
-        type: "set",
-        state: (prevState) => {
-          const newContent = [...prevState.data.content];
+        type: "setData",
```
mkilpatrick, line 178: There is a "replace" type now. Does that make more sense? You'll still need to preserve the id of the target component.

packages/visual-editor/src/internal/puck/components/LayoutHeader.tsx
```diff
@@ -47,20 +49,12 @@ export const LayoutHeader = (props: LayoutHeaderProps) => {
...
-      forward,
-      histories,
-      index,
-      hasPast,
-      hasFuture,
-      setHistories,
-    },
-  } = usePuck();
   const { i18n } = usePlatformTranslation();
+  const getPuck = useGetPuck();
```
mkilpatrick, line 53: Is createUsePuck a superset of useGetPuck (used on line 23)? It sounded like it from your commits desc, except the former also rerenders.

asanehisa (approved):
Not u, but just realized I think this upgrade will be a potentially breaking change for hybrid sites (aka Lilian)

benlife5:
I've update the test site; still running into some copy/paste issues, will continue investigating tomorrow

asanehisa (dismissed):

asanehisa (approved):

asanehisa (approved):

PR #662, opened 2025-07-24
feat: custom code component

k-gerner:
Added a component that allows the user to add custom HTML/CSS/JavaScript onto the page. The component will only render if the HTML field has data. Otherwise, it will show a default empty component. The HTML is sanitized using DOMPurify. This component lives in the "Other" category.

J=[WAT-4957](https://yexttest.atlassian.net/browse/WAT-4957?atlOrigin=eyJpIjoiZDg4ZGRlOWJmZWY5NDk0OGEyNTRiMGUwYzA3MDYxODAiLCJwIjoiaiJ9)

See demo video [here](https://drive.google.com/file/d/11m2PIVX0Wn47QvcPBvJwr_ntZD_Kwu9i/view?usp=sharing).


packages/visual-editor/src/components/CustomCodeSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+  if (!html) {
+    return <EmptyCustomCodeSection />;
+  }
+
+  const containerRef = React.useRef<HTMLDivElement>(null);
+  const [sanitizedHtml, setSanitizedHtml] = React.useState("");
+
+  React.useEffect(() => {
+    if (typeof window !== "undefined") {
+      const DOMPurify = createDOMPurify(window);
```
k-gerner, line 60: Calling `DOMPurify.sanitize(str)` with a direct import was breaking in the `Preview Changes` view. This allows it to work.

packages/visual-editor/package.json
```diff
@@ -122,6 +121,7 @@
     "autoprefixer": "^10.4.2",
     "awesome-phonenumber": "^7.5.0",
     "axe-core": "^4.10.3",
+    "dompurify": "^3.2.3",
```
asanehisa, line 124: remove from devDependencies if its a peerDependency

packages/visual-editor/src/components/CustomCodeSection.tsx
```diff
@@ -0,0 +1,107 @@
+import React from "react";
+import createDOMPurify from "dompurify";
+import { useTranslation } from "react-i18next";
+import { CodeXml } from "lucide-react";
+import { AnalyticsScopeProvider } from "@yext/pages-components";
+import { YextField, msg } from "@yext/visual-editor";
+import { ComponentConfig, Fields, WithId, WithPuckProps } from "@measured/puck";
+
+export interface CustomCodeSectionProps {
```
asanehisa, line 9: If we want this component documented, you'll need to add` /** */` tsdocs to it.

packages/visual-editor/src/components/CustomCodeSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+        ref={containerRef}
+        dangerouslySetInnerHTML={{ __html: sanitizedHtml }}
+      />
+    </div>
+  );
+};
+
+export const CustomCodeSection: ComponentConfig<CustomCodeSectionProps> = {
+  label: msg("components.customCodeSection", "Custom Code Section"),
+  fields: customCodeSectionFields,
```
asanehisa, line 99: Do you want it to default to anything?

packages/visual-editor/src/components/CustomCodeSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+    type: "code",
+    codeLanguage: "css",
+  }),
+  javascript: YextField("JavaScript", {
+    type: "code",
+    codeLanguage: "javascript",
+  }),
+};
+
+const EmptyCustomCodeSection = () => {
```
asanehisa, line 35: demo looks awesome! small note, should the max-w-pageSection apply to this as well?

packages/visual-editor/package.json
```diff
@@ -155,6 +155,7 @@
     "@yext/search-ui-react": "^1.9.0",
     "mapbox-gl": "^2.9.2",
     "react": "^17.0.2 || ^18.2.0",
-    "react-dom": "^17.0.2 || ^18.2.0"
+    "react-dom": "^17.0.2 || ^18.2.0",
+    "dompurify": "^3.2.3"
```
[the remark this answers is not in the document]
benlife5: I think we can make just make it a regular dependency of VE

packages/visual-editor/src/components/CustomCodeSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const { t } = useTranslation();
+  return (
+    <div className="flex flex-col items-center justify-center gap-4 p-3 bg-gray-100 rounded-md">
+      <CodeXml className="w-12 h-12 text-gray-300" />
+      <span className="text-gray-500 text-lg font-medium font-body-fontFamily">
+        {t("missingHtmlWidget", "Add HTML to view component")}
```
benlife5, line 41: We probably only want to display this in the editor, right? There's a prop `puck.isEditing` you can use

packages/visual-editor/src/components/CustomCodeSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+      </span>
+    </div>
+  );
+};
+const CustomCodeSectionWrapper = ({
+  html,
+  css,
+  javascript,
+}: WithId<WithPuckProps<CustomCodeSectionProps>>) => {
+  if (!html) {
```
benlife5, line 51: Do we want `!html && !css && !javascript`?

packages/visual-editor/src/components/CustomCodeSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const prevScript = containerRef.current.querySelector(
+      "#custom-code-section-script"
+    );
+    if (prevScript) {
+      prevScript.remove();
+    }
+
+    if (javascript) {
+      const script = document.createElement("script");
+      script.id = "custom-code-section-script";
```
benlife5, line 79: Might need to use a numerical id if we want to allow >1 on the page?

packages/visual-editor/src/components/__screenshots__/Directory.test.tsx/Directory-renders--version-0-default-props-with-document----desktop--1.png
benlife5, line 1: The `__screenshots__` folders should be gitignored

packages/visual-editor/src/editor/CodeField.tsx
```diff
@@ -0,0 +1,83 @@
...
+          window.location.href.includes("http://localhost:5173/dev-location")
+        ) {
+          const userInput = prompt("Enter Code:");
+          onChange(userInput ?? "");
+        }
+      };
+
+      return (
+        <FieldLabel label={fieldLabel}>
+          <button className="RichTextField" onClick={handleClick}>
```
benlife5, line 76: Let's rename this class to something more generic if it's going to be used for more than RTF

packages/visual-editor/src/components/CustomCodeSection.tsx
benlife5: Yeah def want to allow just HTML, I wasn't sure if we want to allow just JS or just CSS

packages/visual-editor/src/components/__screenshots__/Directory.test.tsx/Directory-renders--version-0-default-props-with-document----desktop--1.png
benlife5: Ah shoot looks like it got lost in a merge. Yes if you could add `**/__screenshots__/**` to the gitignore and delete all these that were commited, that would be great, thanks!

packages/visual-editor/src/components/CustomCodeSection.tsx
k-gerner: I think we don't want to allow just CSS or JS based on [Aaron's comment](https://yext.slack.com/archives/C06A06BCUUF/p1753381395043379?thread_ts=1753300074.714109&cid=C06A06BCUUF)

packages/visual-editor/src/components/CustomCodeSection.tsx
k-gerner: Thanks! This component doesn't use the `PageSection` atom. Do we still want to add this? Sidenote, would it instead be `max-w-pagesection-contentWidth`?

packages/visual-editor/src/editor/CodeField.tsx
k-gerner: Got it. I went ahead and created a separate `CodeField` CSS class for now so that I could make the font `monospace` and avoid needed to change the `RichTextField` name

packages/visual-editor/src/components/CustomCodeSection.tsx
k-gerner: Ah I misunderstood your original comment. Thanks for the tip, updated!

packages/visual-editor/src/components/CustomCodeSection.tsx
k-gerner: Added some comments to the field, lmk if there were any other places I should add comments

packages/visual-editor/package.json
k-gerner: Weird, it wasn't working before, but maybe my change to use `createDOMPurify` made it work. Added to `dependencies` instead

packages/visual-editor/src/components/CustomCodeSection.tsx
asanehisa: Oh add a comment right above `CustomCodeSection ` on 110 that says what this component is / does

packages/visual-editor/src/components/CustomCodeSection.tsx
k-gerner: I added it, doesn't seem to have any visual change

packages/visual-editor/src/components/CustomCodeSection.tsx
benlife5: I'm not totally sure but I think this won't run in Page Generation so the html content will flash when you open the live page? The docs have this example, maybe we do this if window is undefined?
```
import { JSDOM } from 'jsdom';
import DOMPurify from 'dompurify';

const window = new JSDOM('').window;
const purify = DOMPurify(window);
const clean = purify.sanitize('<b>hello there</b>');
```

packages/visual-editor/vite.config.ts
```diff
@@ -28,6 +28,7 @@ export default defineConfig(() => ({
         "mapbox-gl",
         "@yext/search-headless-react",
         "@yext/search-ui-react",
+        "dompurify",
```
benlife5, line 31: I don't think we need this if it's a dependency

packages/visual-editor/src/components/CustomCodeSection.tsx
asanehisa: Oh you need "components" in the classnames somewhere for any of the theme based props to work.

k-gerner: Hm actually, comparing to the mocks, it was closer before this change. I think we should probably not have the empty white space on the sides. Going to remove it for now

packages/visual-editor/src/components/CustomCodeSection.tsx
k-gerner: Decided to move the sanitization logic into `CodeField` so that we don't have to worry about these issues

packages/visual-editor/src/components/CustomCodeSection.tsx
```diff
@@ -0,0 +1,116 @@
...
+    if (prevScript) {
+      prevScript.remove();
+    }
+
+    if (javascript) {
+      const script = document.createElement("script");
+      script.id = scriptTagId;
+      script.type = "text/javascript";
+      script.innerHTML = javascript;
+      containerRef.current.appendChild(script);
```
mkilpatrick, line 90: If someone inputs `<script>...</script>` does it double up on the script tag? Are we expecting people to use the script tag in their input?

packages/visual-editor/src/editor/CodeField.tsx
```diff
@@ -0,0 +1,89 @@
...
+            value: value,
+            id: messageId,
+            fieldName: fieldLabel,
+            codeLanguage: codeLanguage,
+          },
+        });
+
+        /** Handles local development testing outside of storm */
+        if (
+          window.location.href.includes("http://localhost:5173/dev-location")
```
mkilpatrick, line 73: Is there a better way we can check if running locally? Hardcoding against localhost or the port isn't going to work in envs like Stackblitz

packages/visual-editor/src/editor/index.css
```diff
@@ -81,3 +81,24 @@
 .RichTextField:hover {
   border-color: var(--puck-color-grey-05);
 }
+
+/* Temporary styling for CodeFields until we add CodeBlock component from CodeMirror */
```
mkilpatrick, line 85: Why aren't we using CodeMirror now if that's the goal?

k-gerner: Aaron / Elisabeth said not to worry about it for now

packages/visual-editor/src/editor/CodeField.tsx
k-gerner: I just copied the way it was done in the existing rich text field file

packages/visual-editor/src/components/CustomCodeSection.tsx
k-gerner: We are not expecting that, but I suppose it's possible. Maybe this could be communicated in the usage docs?

k-gerner: Actually the `sanitize` we have in `CodeField` should remove any `<script>` tags in the HTML

PR #667, opened 2025-07-28
feat: honor layoutTaskApprovals site setting

asanehisa:
```
If assignment = ENTITY:
	If LayoutTaskApproval:
		Open Modal + Send for Approval
	Else: 
		Publish Entity Layout Override
If assignment = ALL:
	Publish Layout
```

Tested in a packed version with storm changes.

### DIFF
--- a/packages/visual-editor/src/internal/puck/components/LayoutHeader.tsx
+++ b/packages/visual-editor/src/internal/puck/components/LayoutHeader.tsx
@@ -68,6 +68,27 @@ export const LayoutHeader = (props: LayoutHeaderProps) => {
 
   useEffect(translatePuckSidebars, [i18n.language]);
 
+  const buttonText =
+    templateMetadata.assignment === "ALL"
+      ? // TODO: translation concatenation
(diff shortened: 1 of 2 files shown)


packages/visual-editor/src/internal/puck/components/LayoutHeader.tsx
```diff
@@ -68,6 +68,27 @@ export const LayoutHeader = (props: LayoutHeaderProps) => {
 
   useEffect(translatePuckSidebars, [i18n.language]);
 
+  const buttonText =
```
mkilpatrick, line 71: This might be a little easier to ready by just using if/else. I prefer this because it clearly shows the ALL vs ENTITY logic/sublogic. While what you have is technically more terse, it's harder to understand IMO. Open to others' opinions though.

```
if (ALL) update x pages
else if (ENTITY)
     if (layoutTaskApprovals) send for approval
     return update page
```

asanehisa: I can do that! I agree that too many ternary operators can be harder to read

benlife5 (approved):

PR #671, opened 2025-07-28
feat: use favicon from document

asanehisa:
Works with new favicon (Spruce made) as long as you make a new deploy after adding the favicon.

### DIFF
--- a/packages/visual-editor/src/vite-plugin/templates/directory.tsx
+++ b/packages/visual-editor/src/vite-plugin/templates/directory.tsx
@@ -28,7 +28,7 @@ export const getHeadConfig: GetHeadConfig<TemplateRenderProps> = ({
   document,
 }): HeadConfig => {
   const { title, description } = getPageMetadata(document);
-  const faviconUrl = document?._site?.favicon?.url;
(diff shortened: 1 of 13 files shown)


benlife5 (approved):

packages/visual-editor/src/vite-plugin/templates/directory.tsx
```diff
@@ -28,7 +28,7 @@ export const getHeadConfig: GetHeadConfig<TemplateRenderProps> = ({
   document,
 }): HeadConfig => {
   const { title, description } = getPageMetadata(document);
-  const faviconUrl = document?._site?.favicon?.url;
+  const faviconUrl = document?._favicon;
```
mkilpatrick, line 31: These should fall back to the old site entity if _favicon is not available. This way existing sites don't end up breaking. Same throughout.

asanehisa: Oh ok, can do. I (assumed) we were planning to get rid of the site entity / delete it.

benlife5 (approved):

jwartofsky-yext (approved):

PR #673, opened 2025-07-28
feat: support for component filtering

anguyen-yext2:
- The gated components will only be visible and draggable if a bussiness has the appropriate product feature enabled.
- If a gated component is present on a page, but the feature is disabled later, it will not show on live page.
- In the editor, the component will not appear on the left hand side components list. If it was previously present on the page, a section with the text "No configuration for XyzSection" will be displayed on the canvas instead of the gated component. It is removable.

J=WAT-4956
TEST=manual

reference this packed version in a starter branch and tested in platform. verified that the editor and live page look as expected when with the product feature enabled/disabled.


starter/src/ve.config.tsx
```diff
@@ -51,3 +51,57 @@ export const mainConfig: Config<MainProps> = {
 export const componentRegistry = new Map<string, Config<any>>([
   ["dev", mainConfig],
 ]);
+
+const gatedLayoutComponents: string[] = ["CustomCodeSection", "GridSection"];
+
+export const filterComponentsFromConfig = <T extends DefaultComponentProps>(
```
benlife5, line 57: Looks good! I think we should move these functions to `src/utils` and then it could be imported in this ve.config but also the ve.config in pages-visual-editor-starter. Can you add some TSDoc for these functions too?

vijay267 (commented):
Generally looks fine to me outside of existing comments, so will defer to Sumo on the full approval.

benlife5 (dismissed):

packages/visual-editor/src/utils/filterComponents.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Filters gated layout components from a specific configuration in the registry.
+ *
+ * @param registry - The registry to be filtered.
+ * @param registryKey - The key in the registry whose associated configuration should be filtered.
+ * @param additionalLayoutComponents - An optional list of gated component names to retain.
+ * @returns The new registry with the filtered configuration under the specified key.
+ */
+export const filterComponentsFromRegistry = (
```
mkilpatrick, line 55: I feel like this isn't really necessary, at least to be exported. The user can use filterComponentsFromConfig directly since they're already working with the correct config for the template.

packages/visual-editor/src/utils/index.ts
```diff
@@ -41,3 +41,7 @@ export {
 } from "./migrate.ts";
 export { withPropOverrides } from "./withPropOverrides.ts";
 export { getLocationPath } from "./getLocationPath.ts";
+export {
+  filterComponentsFromConfig,
+  filterComponentsFromRegistry,
```
mkilpatrick, line 46: Same comment. Don't export this one.

packages/visual-editor/src/vite-plugin/templates/edit.tsx
```diff
@@ -40,6 +41,12 @@ const Edit: () => JSX.Element = () => {
   const entityDocument = usePlatformBridgeDocument();
   const entityFields = usePlatformBridgeEntityFields();
 
+  const filteredRegistry = filterComponentsFromRegistry(
+    componentRegistry,
+    "main",
```
mkilpatrick, line 46: This isn't going to work. This means that only the main config will ever work in the editor. Test going to a locator or directory pageset. I don't think it will work.

packages/visual-editor/src/utils/filterComponents.tsx
```diff
@@ -0,0 +1,69 @@
+import { DefaultComponentProps, type Config } from "@measured/puck";
+
+const gatedLayoutComponents: string[] = ["CustomCodeSection", "GridSection"];
```
jwartofsky-yext, line 3: These are components we remove by default?

Can you add a comment for this

packages/visual-editor/src/vite-plugin/templates/edit.tsx
benlife5: oh true. Instead of filtering here, we should filter when we setPuckConfig in `packages/visual-editor/src/internal/hooks/useMessageReceivers.ts`

packages/visual-editor/src/utils/filterComponents.tsx
anguyen-yext2: yep! added!

packages/visual-editor/src/vite-plugin/templates/edit.tsx
anguyen-yext2: Good call! 

- Updated `sendTemplateMetadata` in alpha to send `additionalLayoutComponents` (https://gerrit.yext.com/c/alpha/+/301424/1/gocode/src/yext/publish/yextsites/storm/js/visualeditor/UniversalMessenger.tsx). 
- Moved filtering logic in this PR to `useMessageReceivers.ts`

packages/visual-editor/src/utils/filterComponents.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+// The gated components that are disallowed by default. They are only visible and draggable if a business has the appropriate product features enabled.
+const gatedLayoutComponents: string[] = ["CustomCodeSection", "GridSection"];
+
+/**
+ * Filters out gated layout components from a configuration object unless explicitly allowed.
+ *
+ * @param config - The original configuration object.
+ * @param additionalLayoutComponents - An optional list of gated component names to retain.
+ * @returns The new configuration object with gated components that are not present in `additionalLayoutComponents` removed.
```
asanehisa, line 11: @returns A new configuration object with all gated components removed, except for those listed in `additionalLayoutComponents` ? Might be more clear? Took me a couple tries to understand the initial version

anguyen-yext2: Updated the comment per Alexis's recommendation. 
@jwartofsky-yext I didn't update the variable name because it matches with the entity document's field name. If you still think it make sense to update them, I'll be happy to do so, but I just wanted to point out that I intentionally kept the name consistent.

packages/visual-editor/src/vite-plugin/templates/main.tsx
```diff
@@ -90,6 +91,10 @@ export const getPath: GetPath<TemplateProps> = ({ document }) => {
 
 const Location: Template<TemplateRenderProps> = (props) => {
   const { document } = props;
+  const filteredConfig = filterComponentsFromConfig(
```
mkilpatrick, line 94: Noting that you'll need to add this to the starter too

mkilpatrick (commented):
lgtm other than merge conflicts with screenshots

packages/visual-editor/src/vite-plugin/templates/main.tsx
mkilpatrick: My mistake, we don't have the templates in the starter anymore, only the default data. Nvm then!

packages/visual-editor/src/utils/index.ts
```diff
@@ -41,3 +41,4 @@ export {
 } from "./migrate.ts";
 export { withPropOverrides } from "./withPropOverrides.ts";
 export { getLocationPath } from "./getLocationPath.ts";
+export { filterComponentsFromConfig } from "./filterComponents.tsx";
```
mkilpatrick, line 44: I don't believe we should be exporting this if it will never be used by anyone externally. Looks like we've now contained it all internally, right?

mkilpatrick: Nvm

PR #677, opened 2025-07-29
fix: coreInfo headingAlign prop applies to all columns

jwartofsky-yext:
HoursTable is now scrollable if it doesn't fit "x-overflow-auto"

The scrollbar vanishes a second after you stop scrolling, but it's there

image


packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[tablet] version 1 props with partial entity values 3 and align right.png
Comment (mkilpatrick) [inline packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[tablet] version 1 props with partial entity values 3 and align right.png:1]: Overlapping looks bad. Are we not just using a 3 col flex with a gap?

benlife5 (approved):

packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[tablet] version 10 props with partial entity values alight right.png
Comment (mkilpatrick) [inline packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[tablet] version 10 props with partial entity values alight right.png:1]: Still has overlap

mkilpatrick:
:( to the force push

packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[desktop] version 1 props with partial entity values 3.png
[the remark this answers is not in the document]
Comment (jwartofsky-yext) [reply packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[desktop] version 1 props with partial entity values 3.png:1]: I think it's because the hours are variable length now, and they weren't left aligned properly before. They were just identical before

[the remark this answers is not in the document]
Comment (jwartofsky-yext) [reply packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[desktop] version 1 props with partial entity values 3.png:1]: Just checked, it was right-aligned before. I think this is how we want it, but I sent a message to Aaron to confirm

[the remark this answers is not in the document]
Comment (jwartofsky-yext) [reply packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[desktop] version 1 props with partial entity values 3.png:1]: Aaron says to left align

However it is right-aligned in master (not affected by these changes)

It's noticeable now because I changed the hours testing data to have different timeframes during the week.

It seems that something in pages-components is forcing the right align and I haven't been able to get it to left align through VE yet

We may need a pages-components update to add a new prop

There's no regression, so this PR should be good to go

packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[mobile] version 1 props with constant value.png
Comment (benlife5) [inline packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[mobile] version 1 props with constant value.png:1]: Obviously the hours could be better here but that can be a separate ticket

packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[tablet] version 0 props with constant value.png
Comment (benlife5) [inline packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[tablet] version 0 props with constant value.png:1]: same

packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[tablet] version 0 props with entity values.png
Comment (benlife5) [inline packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[tablet] version 0 props with entity values.png:1]: Might need to fix this in this PR since otherwise we lose the close time data

Comment (jwartofsky-yext) [reply packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[tablet] version 0 props with entity values.png:1]: I think this is improved?

We can make the hours better in another CR, but this just makes the hours not spill into the other column. They're also scrollable on the live page so the user can see everything

benlife5 (approved):

packages/visual-editor/src/editor/index.css
```diff
@@ -9,13 +9,14 @@
 .HoursTable-row {
   display: flex;
   justify-content: space-between;
-  width: 18rem;
+  width: 100%;
```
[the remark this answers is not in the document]
jwartofsky-yext: https://github.com/yext/js/pull/102

mkilpatrick (approved):

benlife5 (approved):

PR #687, opened 2025-08-05
feat: open image upload drawer for static images

briantstephan:
This replaces the URL text input for static images with a thumbnail of the existing image and a button to open the image upload drawer and select a new image.

https://github.com/user-attachments/assets/b5bedb1f-b423-4618-b2df-f6a7018badc5

### DIFF
--- a/packages/visual-editor/CHANGELOG.md
+++ b/packages/visual-editor/CHANGELOG.md
@@ -1,3 +1,21 @@
+#### 0.0.52 (2025-08-11)
+
+##### Chores
+
+- add TranslatableString type to YextField (#706) (2fb957e8)
+
+##### New Features
+
+- update array field item labels (#708) (4b816dde)
+- support category filtering (#705) (d3663709)
+- warn when translation keys are deleted (#700) (f79ec675)
+- allow different urls per locale for CTA links (#698) (800601d8)
+- improve CoreInfoSection styling for tablet (#701) (1e03777a)
+
+##### Refactors
+
+- hybrid developer work flow (#699) (fdf820e1)
+
 #### 0.0.51 (2025-08-06)
 
 ##### Chores
--- a/packages/visual-editor/package.json
+++ b/packages/visual-editor/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@yext/visual-editor",
   "description": "Component library for Yext Pages Visual Editor",
-  "version": "0.0.51",
+  "version": "0.0.52",
   "type": "module",
   "author": "sumo@yext.com",
   "types": "dist/index.d.ts",
--- a/packages/visual-editor/src/internal/puck/ui/Combobox.tsx
+++ b/packages/visual-editor/src/internal/puck/ui/Combobox.tsx
@@ -12,6 +12,7 @@ import {
 } from "./Command.tsx";
 import { Popover, PopoverContent, PopoverTrigger } from "./Popover.tsx";
 import { pt } from "../../../utils/i18n/platform.ts";
+import { convertComputedStyleColorToHex } from "../../../utils/colors.ts";
 
 export type ComboboxOption = {
   label: string;
@@ -101,6 +102,10 @@ export const Combobox = ({
                     </p>
                   )}
                   {group.options.map((option) => {
+                    const [colorHexValue, setColorHexValue] = React.useState<
+                      string | undefined
+                    >(undefined);
+
                     return (
                       <CommandItem
                         className="ve-cursor-pointer"
@@ -119,8 +124,18 @@ export const Combobox = ({
                               : "ve-opacity-0"
                           )}
                         />
-                        <ColorIndicator color={option.color} />
-                        {option.label}
+                        <ColorIndicator
+                          color={option.color}
+                          setHexValue={setColorHexValue}
+                        />
+                        <div className="ve-flex ve-flex-col ve-gap-0.5">
+                          {option.label}
+                          {colorHexValue && (
+                            <span className="ve-text-sm ve-text-[#5B5D60]">
+                              {colorHexValue}
+                            </span>
+                          )}
+                        </div>
                       </CommandItem>
                     );
                   })}
@@ -134,14 +149,34 @@ export const Combobox = ({
   );
 };
 
-const ColorIndicator = ({ color }: { color?: string }) => {
+const ColorIndicator = ({
+  color,
+  setHexValue,
+}: {
+  color?: string;
+  setHexValue?: (s: string) => void;
+}) => {
+  const ref = React.useRef<HTMLDivElement>(null);
+  let colorHexValue = undefined;
+
+  React.useEffect(() => {
+    if (color && ref.current && setHexValue) {
+      colorHexValue = convertComputedStyleColorToHex(
+        window.getComputedStyle(ref.current).backgroundColor
+      );
+      setHexValue(colorHexValue);
+    }
+  }, [color, setHexValue]);
+
   if (!color) {
-    return;
+    return null;
   }
+
   return (
     <div
+      ref={ref}
       className={cn(
-        "ve-ring-1 ve-ring-inset ve-ring-ring ve-w-3 ve-h-3 ve-rounded-sm ve-mr-2 components",
+        "ve-ring-1 ve-ring-inset ve-ring-ring ve-w-5 ve-h-5 ve-rounded-sm ve-mr-2 components",
         color
       )}
     />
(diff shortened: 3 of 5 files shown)


packages/visual-editor/src/internal/puck/constant-value-fields/Image.tsx
```diff
@@ -1,21 +1,145 @@
 import { CustomField } from "@measured/puck";
 import { ImageType } from "@yext/pages-components";
-import { ConstantFields } from "./ConstantField.tsx";
-import { pt } from "../../../utils/i18n/platform.ts";
+
```
benlife5, line 3: empty line

packages/visual-editor/src/internal/puck/constant-value-fields/Image.tsx
```diff
@@ -1,21 +1,145 @@
...
+        });
+      }
+    };
+
+    return (
+      <>
+        {value?.url && (
+          <img
+            src={value.url}
+            alt="Selected Thumbnail"
```
benlife5, line 130: `pt`

packages/visual-editor/src/internal/puck/constant-value-fields/Image.tsx
```diff
@@ -1,21 +1,145 @@
...
+            className="mt-3 w-full min-h-24 max-h-48 object-cover rounded-md mb-3"
```
benlife5, line 131: These should be `ve-`

packages/visual-editor/src/internal/puck/constant-value-fields/Image.tsx
```diff
@@ -1,21 +1,217 @@
...
+          type: "ImageAsset",
+          value: value,
+          id: messageId,
         },
-      ],
-    });
+      });
+
+      /** Handles local development testing outside of storm */
+      if (window.location.href.includes("http://localhost:5173/dev-location")) {
```
jwartofsky-yext, line 122: might be better to just check up to localhost:5173

packages/visual-editor/src/internal/puck/constant-value-fields/Image.tsx
```diff
@@ -1,21 +1,217 @@
...
+          url: userInput ?? "",
+          height: 0,
+          width: 0,
+        });
+      }
+    };
+
+    const handleDeleteImage = (e: React.MouseEvent) => {
+      e.stopPropagation();
+      // Defer state update to prevent race condition
```
jwartofsky-yext, line 138: explain this more

this seems like a hack (but maybe it's needed)

Could there be a better way?

packages/visual-editor/src/internal/puck/constant-value-fields/Image.tsx
```diff
@@ -1,21 +1,217 @@
...
+      "constantValueEditorOpened",
+      TARGET_ORIGINS
+    );
+
+    useReceiveMessage(
+      "constantValueEditorClosed",
+      TARGET_ORIGINS,
+      (_, payload) => {
+        const imagePayload = payload as ImagePayload;
+        if (pendingMessageId && pendingMessageId === imagePayload.id) {
```
jwartofsky-yext, line 91: Is this how we determine that the message received is for the image selector? (implying we could receive messages here for other components/fields)

packages/visual-editor/src/internal/puck/constant-value-fields/Image.tsx
benlife5: in-platform dev mode runs at :5173 too

briantstephan: This is the same as the check we do in other places like RTF editing, probably best to keep it consistent.

benlife5 (approved):

packages/visual-editor/src/internal/puck/constant-value-fields/Image.tsx
```diff
@@ -1,21 +1,220 @@
...
+          width: 0,
+        });
+      }
+    };
+
+    const handleDeleteImage = (e: React.MouseEvent) => {
+      e.stopPropagation();
+      /* Deleting an image replaces the thumbnail with the "Choose Image" button.
+         The click event from this button is still "active", and gets passed off 
+         to the new Choose Image button, which triggers the image selector.
```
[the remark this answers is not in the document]
briantstephan: 1. No, because the click event propagates up to the FieldLabel, and we've already handled the clicking behavior for that with the hidden button. I've added a `setTimeout` here just to be safe though, in case we change the click behavior elsewhere later on.
2. My understanding is that, without the `setTimeout` the Delete button click event is still "active" after deleting the image (and removing the thumbnail). It then gets passed off to the Choose Image button that takes the place of the thumbnail and Delete button, triggering its onclick event. This causes the drawer to open when deleting the existing image. Even though this timeout is 0ms, the setTimeout causes onChange to run after handleDeleteImage finishes running, which completes the actual click event before the Choose Image button is added to the page. It effectively gets put in a queue of things to run after the click events completes.

packages/visual-editor/src/internal/puck/constant-value-fields/Image.tsx
```diff
@@ -1,21 +1,222 @@
...
+
+type TransformKind = "CROP" | "ROTATION";
+
+/** Outlines the possible image transformations. */
+type ImageTransformation = ImageRotation | ImageCrop;
+
+type ImageTransformations = Partial<Record<TransformKind, ImageTransformation>>;
+
+/** Describes the data corresponding to a piece of image content. */
+type ImageContentData = {
```
asanehisa, line 60: Hmmm is none of the required..?

asanehisa (approved):

packages/visual-editor/src/internal/puck/constant-value-fields/Image.tsx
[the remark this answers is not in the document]
mkilpatrick: > I've added a setTimeout here just to be safe though

Are you sure it's needed? `e.stopPropagation()`'s purpose is to stop the bubbling. Try adding `e.preventDefault()` as well

packages/visual-editor/src/internal/puck/constant-value-fields/Image.tsx
```diff
@@ -1,21 +1,222 @@
...
+        { types: ["type.string"] }
+      );
+    }, []);
+
+    return (
+      <>
+        {/* Thumbnail */}
+        <FieldLabel label={pt("Image")} className="ve-mt-3">
+          <div className="ve-relative ve-group ve-mb-3">
+            {/* FieldLabel grabs the onclick event for the first button in its children,
```
mkilpatrick, line 167: What was happening without this?

packages/visual-editor/src/internal/puck/constant-value-fields/Image.tsx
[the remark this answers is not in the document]
briantstephan: Ah `e.preventDefault()` was what I needed. Updated to use that instead of `setTimeout`.

packages/visual-editor/src/internal/puck/constant-value-fields/ProductSection.tsx
```diff
@@ -87,16 +89,7 @@ const ProductStructArrayField = (): ArrayField<ProductStruct[]> => {
...
-        type: "object",
-        label: t("fields.image", "Image"),
-        objectFields: {
-          url: {
-            label: t("fields.url", "URL"),
-            type: "text",
-          },
-        },
-      },
+      image: IMAGE_CONSTANT_CONFIG,
```
asanehisa, line 92: Oh! its possible eventSection, teamSection, and insight section need this same fix

briantstephan: Good call, updated

mkilpatrick (approved):

packages/visual-editor/src/components/pageSections/TeamSection.tsx
```diff
@@ -205,7 +205,7 @@ const PersonCard = ({
               aspectRatio={
                 person.headshot.width && person.headshot.height
                   ? person.headshot.width / person.headshot.height
-                  : 1.78
+                  : 1
```
jwartofsky-yext, line 208: this seems unrelated

jwartofsky-yext (approved):

benlife5 (approved):

benlife5 (dismissed):

benlife5 (approved):

PR #689, opened 2025-08-05
feat: use urlTemplate from _pageset.config

asanehisa:
Do not merge until supported and backfilled by Spruce. This has been tested with hybrid flow and confirmed it works. 

TODO: 
✅ Test again with actual data once Spruce is done with their item.
✅ Add fallback to old logic if the URL template is not found

Tested again with hybrid development mode

Verified that NearbyLocations, Locator, and Directory links work as expected.

Verified that entity/Location pages urls work as expected.


packages/visual-editor/src/components/testing/screenshots/Directory/[desktop] version 7 - directory cards - default props.png
Comment (mkilpatrick) [inline packages/visual-editor/src/components/testing/screenshots/Directory/[desktop] version 7 - directory cards - default props.png:1]: Why did the title overflow change? It's better now but this PR shouldn't affect that, right? Also the titles are no longer links it appears.

packages/visual-editor/src/utils/index.ts
```diff
@@ -1,44 +1,44 @@
...
-  type MigrationAction,
-  type MigrationRegistry,
-  migrate,
-} from "./migrate.ts";
+  constructFontSelectOptions,
+  defaultFonts,
+  getFontWeightOptions,
+  getFontWeightOverrideOptions,
+  type FontRegistry,
+} from "./visualEditorFonts.ts";
```
mkilpatrick, line 43: Seems like a lot of changes in this file

packages/visual-editor/src/utils/resolveUrlTemplate.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @param locale - The locale to use for resolving the URL.
+ * @param relativePrefixToRoot - Optional prefix to prepend to the resolved URL.
+ * @returns The resolved and normalized URL.
+ */
+export const resolveUrlTemplate = (
+  streamDocument: StreamDocument,
+  locale: string,
+  relativePrefixToRoot?: string
+): string => {
+  const isPrimaryLocale = streamDocument.__?.isPrimaryLocale ?? false;
```
mkilpatrick, line 19: Should we update our type so that __ and _pageset always exist?

packages/visual-editor/src/components/testing/screenshots/Directory/[desktop] version 7 - directory cards - default props.png
Comment (asanehisa) [reply packages/visual-editor/src/components/testing/screenshots/Directory/[desktop] version 7 - directory cards - default props.png:1]: Yeah it'll be fixed once I update the dev document data used for testing (rn it isn't showing up as links b/c it isn't a link)

packages/visual-editor/src/utils/index.ts
asanehisa: Ah just reordered it alphabetically + added my one new export

packages/visual-editor/src/utils/resolveUrlTemplate.ts
asanehisa: I don't think it exists in hybrid use case.

packages/visual-editor/src/components/pageSections/NearbyLocations.tsx
```diff
@@ -291,11 +289,10 @@ const LocationCard = ({
   address: AddressType;
   timezone: string;
   mainPhone: string;
-  id: string;
-  slug?: string;
 }) => {
-  const { relativePrefixToRoot } = useTemplateProps();
+  const { document: streamDocument, relativePrefixToRoot } = useTemplateProps();
```
mkilpatrick, line 293: This won't work since the NearbyLocation data comes from an api call, not the document.

packages/visual-editor/src/utils/resolveUrlTemplate.ts
mkilpatrick: Ah, true. So in that case for everything to work I think we need to default back to what getLocationPath had. But I think that actually doesn't fully solve the issue either. I'm thinking the true solve is for the different components that need the url format is for them to take in an optional function that returns the slug. If set, prefer that one, otherwise use this one.

packages/visual-editor/src/components/pageSections/NearbyLocations.tsx
asanehisa: Talked offline: Use urlTemplate from streamDocument, use data from api call

packages/visual-editor/src/utils/resolveUrlTemplate.ts
```diff
@@ -0,0 +1,36 @@
...
+    pagesetJson?.config?.urlTemplate?.[
+      isPrimaryLocale ? "primary" : "alternate"
+    ];
+  if (!urlTemplate) {
+    return "";
+  }
+
+  return (
+    (relativePrefixToRoot ?? "") +
+    normalizeSlug(
```
mkilpatrick, line 32: We need to check that the final slug (before relativePrefixToRoot is added) doesn't result in an empty string. If so, throw an error so that generation fails.

packages/visual-editor/src/components/Directory.tsx
```diff
@@ -175,15 +180,28 @@ const DirectoryCard = ({
...
+  // If a hybrid developer passes in a resolveUrlTemplate function, use that to resolve the URL.
+  // Otherwise, use the default resolveUrlTemplate function.
+  const resolvedUrl =
+    typeof puck.metadata?.resolveUrlTemplate === "function"
+      ? puck.metadata.resolveUrlTemplate(
+          streamDocument,
+          locale,
+          relativePrefixToRoot ?? ""
+        )
+      : resolveUrlTemplate(streamDocument, locale, relativePrefixToRoot ?? "");
```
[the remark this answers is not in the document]
asanehisa: Oh like have resolveUrlTemplate have an optional overrideFunction param?

packages/visual-editor/src/components/Locator.tsx
```diff
@@ -167,9 +171,9 @@ const LocatorWrapper: React.FC<LocatorProps> = (props) => {
 
 type SearchState = "not started" | "loading" | "complete";
 
-const LocatorInternal: React.FC<LocatorProps> = (props) => {
+const LocatorInternal: React.FC<WithPuckProps<LocatorProps>> = (props) => {
   const { t } = useTranslation();
-  const { mapStyle, openNowButton, entityTypeEnvVar } = props;
+  const { mapStyle, openNowButton, entityTypeEnvVar, puck } = props;
```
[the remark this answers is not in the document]
mkilpatrick: Yea, then you can just pass it down, especially since the props for both are the same.

packages/visual-editor/src/components/testing/screenshots/Directory/[desktop] version 8 - directory card - non-default props.png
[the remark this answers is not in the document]
Comment (asanehisa) [reply packages/visual-editor/src/components/testing/screenshots/Directory/[desktop] version 8 - directory card - non-default props.png:1]: Hmmm I can look at recent commits but since the before + after appear the same its possible fix wasn't merged yet / was reverted.

packages/visual-editor/src/components/Locator.tsx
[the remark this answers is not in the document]
asanehisa: Yeah I don't see a diff function to get this puck context with metadata.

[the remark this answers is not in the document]
mkilpatrick: Darn, ok

packages/visual-editor/src/utils/resolveUrlTemplate.test.ts
```diff
@@ -0,0 +1,158 @@
...
+        config: {
+          urlTemplate: {
+            primary:
+              "[[address.region]]/[[address.postalCode]]/[[address.city]]",
+          },
+        },
+      }),
+    };
+    const result = resolveUrlTemplate(docWithUnresolvableField, "en");
+    assert.equal(result, "ny//new-york");
```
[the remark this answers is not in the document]
asanehisa: Conclusion for now: handle gracefully if the field isn't filled. If the urlTemplate is "" for any reason, throw error

packages/visual-editor/src/components/testing/screenshots/Directory/[desktop] version 8 - directory card - non-default props.png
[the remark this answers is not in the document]
Comment (asanehisa) [reply packages/visual-editor/src/components/testing/screenshots/Directory/[desktop] version 8 - directory card - non-default props.png:1]: Not seeing it as a fix in recent commits. I see we fixed the ExpandedHeader / ExpandedFooter to work with max-width.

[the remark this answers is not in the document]
Comment (benlife5) [reply packages/visual-editor/src/components/testing/screenshots/Directory/[desktop] version 8 - directory card - non-default props.png:1]: The overflow is because it becomes styled as a link, which has different browser default wrapping. This item should fix but is still to do: https://yexttest.atlassian.net/browse/SUMO-7543?atlOrigin=eyJpIjoiZjUwYTFhYjYwMWZiNDhiMDlkYmE0NDdiMzczZDZkMDIiLCJwIjoiaiJ9

benlife5 (approved):

jwartofsky-yext (dismissed):

packages/visual-editor/src/docs/hybrid-development.md
```diff
@@ -149,14 +149,14 @@ Various existing components require additional data from the hybrid-developer. T
 This could look like:
 
 ```ts
-const resolveUrlTemplate = (
-  streamDocument: StreamDocument,
+function resolveUrlTemplate(
```
mkilpatrick, line 152: Can we change this back to the other version? Both here and the actual function.

mkilpatrick (dismissed):

packages/visual-editor/src/utils/getLocationPath.ts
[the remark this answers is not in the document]
asanehisa: Yup will add that (not intending to merge this until Spruce is done + I can test with their stuff)

briantstephan (dismissed):

packages/visual-editor/src/utils/resolveUrlTemplate.ts
```diff
@@ -0,0 +1,60 @@
...
+  const isPrimaryLocale = streamDocument.__?.isPrimaryLocale ?? false;
+
+  const pagesetJson = JSON.parse(streamDocument?._pageset || "{}");
+  const urlTemplate =
+    pagesetJson?.config?.urlTemplate?.[
+      isPrimaryLocale ? "primary" : "alternate"
+    ];
+
+  if (!urlTemplate) {
+    console.warn("No URL template found on document");
```
mkilpatrick, line 41: This will always log for in-repo pagesets (if they used this function) and, more importantly, any non-entity pageset (locator, directory).

packages/visual-editor/src/utils/resolveUrlTemplate.ts
```diff
@@ -0,0 +1,60 @@
...
+    return (
+      getLocationPath(
```
mkilpatrick, line 43: Maybe this function should take in a pageset type and switch on that? Actually, no, we should just have the two other templates pass in the alternateFunction

packages/visual-editor/src/utils/getLocationPath.ts
```diff
@@ -20,10 +22,12 @@ export const getLocationPath = (
     return `${relativePrefixToRoot ?? ""}${location.slug}`;
```
mkilpatrick, line 22: Since string defaults to empty now you don't need the ?? check

packages/visual-editor/src/utils/getLocationPath.ts
```diff
@@ -20,10 +22,12 @@ export const getLocationPath = (
...
   }
 
-  const localePath = locale !== "en" ? `${locale}/` : "";
+  const isPrimaryLocale = location.__?.isPrimaryLocale ?? false;
```
mkilpatrick, line 25: If this isn't always set I worry it can lead to something incorrect.

packages/visual-editor/src/utils/getLocationPath.ts
```diff
@@ -20,10 +22,12 @@ export const getLocationPath = (
...
+
+  const localePath = isPrimaryLocale ? "" : `${locale ?? location.locale}/`;
   const path = location.address
     ? `${localePath}${location.address.region}/${location.address.city}/${location.address.line1}-${location.id}`
```
mkilpatrick, line 29: If this is only used for nearby/last level DM page we'll need to align with what we backfill to, which won't have the suffix.

packages/visual-editor/src/utils/resolveUrlTemplate.ts
jwartofsky-yext: The two other templates don't use resolveUrlTemplate

Asked Spruce earlier and they said the templates are just for Location/Entity pagesets for now

packages/visual-editor/src/utils/resolveUrlTemplate.ts
```diff
@@ -25,35 +26,43 @@ export const resolveUrlTemplate = (
...
+    console.warn("No URL template found on the document");
+    return getLocationPath(
+      streamDocument as LocationDocument,
+      locale,
+      relativePrefixToRoot
     );
   }
 
+  // Replace [[locale]] in the template with the resolved locale
+  urlTemplate = urlTemplate.replace(/\[\[locale\]\]/g, locale);
```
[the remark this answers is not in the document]
jwartofsky-yext: re-added setting document.locale to document.meta.locale if it is missing

removed this replace

packages/visual-editor/src/utils/resolveUrlTemplate.ts
```diff
@@ -25,35 +26,43 @@ export const resolveUrlTemplate = (
...
+
   const normalizedSlug = normalizeSlug(
     resolveEmbeddedFieldsInString(urlTemplate, streamDocument, locale)
   ).replace(/\/+/g, "/"); // replace multiple slashes with a single slash
 
   if (!normalizedSlug) {
-    throw new Error(`Could not resolve URL template: ${urlTemplate}`);
+    throw new Error("Could not resolve URL template");
```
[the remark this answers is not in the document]
jwartofsky-yext: reverted

packages/visual-editor/src/utils/resolveUrlTemplate.ts
```diff
@@ -0,0 +1,65 @@
...
+    relativePrefixToRoot: string
+  ) => string
+): string => {
+  const locale = normalizeLocale(
+    streamDocument.locale || streamDocument?.meta?.locale || ""
+  );
+  if (!locale) {
+    throw new Error(`Could not determine locale from streamDocument`);
+  }
+  if (!streamDocument.locale) {
```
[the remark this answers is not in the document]
jwartofsky-yext: It's a little confusing because resolveEmbeddedFieldsInString takes in the document (which can have locale as a field) and also a locale param

It uses the locale param for TranslatableString type fields to choose which to display

However for our case, we're interested in the embedded field named "locale", which is usually going to be streamDocument.locale but we want it to fallback to streamDocument.meta.locale because sometimes streamDocument.locale is missing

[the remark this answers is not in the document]
mkilpatrick: But streamDocument.locale should ALWAYS be present because PagesJS (and platform) pagesets require locales. You can't set primary:true in the stream scope definition. If we want to be extra safe then we should always use document.meta.locale if that is always set no matter what. Then we don't need this fallback logic of guessing where the value is.

packages/visual-editor/src/utils/resolveUrlTemplate.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  const pagesetJson = JSON.parse(streamDocument?._pageset || "{}");
+  let urlTemplate =
+    pagesetJson?.config?.urlTemplate?.[
+      isPrimaryLocale ? "primary" : "alternate"
+    ];
+
+  if (!urlTemplate) {
+    return getLocationPath(
+      streamDocument as LocationDocument,
```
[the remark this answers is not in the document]
mkilpatrick: You only set streamDocument.locale if locale is not present. In practice I don't think it's a problem given what I said above, but if the locale is set on meta.locale only then you don't reset document.locale, meaning it's not normalized.

packages/visual-editor/src/components/Directory.tsx
```diff
@@ -429,6 +463,8 @@ const DirectoryComponent = ({ data, styles, analytics }: DirectoryProps) => {
     streamDocument
   );
 
+  const locale = streamDocument.locale || streamDocument?.meta?.locale;
```
mkilpatrick: I'm fine removing this and using i18n.language back in DirectoryCard. Confirmed that the dm_directoryChildren don't have locale so we need to pass it from the top level document.

packages/visual-editor/src/components/Directory.tsx
```diff
@@ -418,7 +447,12 @@ const DirectoryList = ({
   );
 };
 
-const DirectoryComponent = ({ data, styles, analytics }: DirectoryProps) => {
+const DirectoryComponent = ({
+  data,
+  styles,
+  analytics,
+  puck,
+}: WithPuckProps<DirectoryProps>) => {
```
mkilpatrick, line 444: Can you do this at the sub component that needs to use it instead of doing it at the top and passing it down?

packages/visual-editor/src/components/Locator.tsx
```diff
@@ -449,7 +453,9 @@ const LocatorInternal = ({
         <div id="innerDiv" className="overflow-y-auto" ref={resultsContainer}>
           {resultCount > 0 && (
             <VerticalResults
-              CardComponent={LocationCard}
+              CardComponent={(result: CardProps<Location>) => (
+                <LocationCard {...result} puck={puck} />
```
mkilpatrick, line 458: Same about passing puck down. Is there not a useHook or just use the WithPuckProps types and it's available at a lower level?

packages/visual-editor/src/components/Locator.tsx
```diff
@@ -610,6 +620,20 @@ const LocationCard: CardComponent<Location> = ({
     "TAP_TO_CALL"
   );
 
+  const documentForLocator = {
+    ...location,
```
mkilpatrick: I may have been wrong in my last comment. Is locale set in the results? It probably isn't so we need to use the locale of the current page, so you can use i18n.language.

Edit: I guess we have streamDocument already so you can use the locale from that

packages/visual-editor/src/components/pageSections/NearbyLocations.tsx
```diff
@@ -297,11 +302,24 @@ const LocationCard = ({
...
-  slug?: string;
+  locationData: any;
+  puck: PuckContext;
 }) => {
-  const { relativePrefixToRoot } = useTemplateProps();
-  const { i18n } = useTranslation();
+  const { document: streamDocument, relativePrefixToRoot } = useTemplateProps();
+
+  const documentForUrlTemplate = {
+    ...locationData,
```
mkilpatrick: This isn't using the right nearby location data now. It's using the current page document.

packages/visual-editor/src/utils/resolveUrlTemplate.ts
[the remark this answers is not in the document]
jwartofsky-yext: Did the "throw document" test again and the locale is present now..

odd

Removed the fallback

packages/visual-editor/src/components/Directory.tsx
jwartofsky-yext: Ack

packages/visual-editor/src/components/Locator.tsx
jwartofsky-yext: Ack

packages/visual-editor/src/utils/getLocationPath.ts
```diff
@@ -1,29 +1,38 @@
...
 }
 
 export const getLocationPath = (
   location: LocationDocument,
-  locale: string,
-  relativePrefixToRoot: string | undefined
-): string | undefined => {
+  relativePrefixToRoot: string = ""
+): string => {
   if (!location || (!location.slug && !location.address && !location.id)) {
```
mkilpatrick, line 15: This checks for address but if slug is being used slug is not necessary

jwartofsky-yext: if slug is being used slug is not necessary?

mkilpatrick: Typo. If slug is being used then address/id aren't necessary to check.

packages/visual-editor/src/utils/mergeMeta.ts
```diff
@@ -0,0 +1,22 @@
...
+  return {
+    locale: streamDocument?.locale,
+    ...profile,
+    meta: {
+      ...streamDocument?.meta,
+      ...profile?.meta,
+    },
+    __: {
+      ...streamDocument.__,
+      ...profile.__,
```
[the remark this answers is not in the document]
jwartofsky-yext: Confirmed with the robots that this will not override __ to be empty if profile.__ is empty

packages/visual-editor/src/utils/getLocationPath.ts
mkilpatrick: You right.

mkilpatrick (dismissed):

benlife5 (approved):

briantstephan (approved):

mkilpatrick (approved):

PR #695, opened 2025-08-06
feat: use entity type name in field selector

benlife5:
Screenshot 2025-08-06 at 2 12 58 PM
Screenshot 2025-08-06 at 2 15 52 PM
Screenshot 2025-08-06 at 2 14 47 PM

Screenshot 2025-08-06 at 2 13 14 PM

https://yext.slack.com/archives/C06A06BCUUF/p1754496210005529

### DIFF
--- a/packages/visual-editor/locales/zh/visual-editor.json
+++ b/packages/visual-editor/locales/zh/visual-editor.json
@@ -1,6 +1,7 @@
 {
   "Heading": "标题",
   "Link": "关联",
+  "a": "一个",
   "actions": {
     "copyToClipboard": "复制到剪贴板",
     "pasteFromClipboard": "从剪贴板粘贴"
@@ -17,6 +18,7 @@
     "region": "州/地区",
     "sublocality": "子域"
   },
+  "an": "一个",
   "approvals": {
     "addAComment": "添加评论以帮助您的批准者了解发生了什么变化",
     "send": "发送批准",
@@ -24,7 +26,6 @@
   },
   "arrayField": "数组字段",
   "base": "根据",
-  "basicSelectorContentLabel": "选择一个内容字段",
   "basicSelectorNoOptionsLabel": "没有可用的选项",
   "body": "身体",
   "borderRadiusDefaultLabel": "默认",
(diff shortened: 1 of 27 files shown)


packages/visual-editor/locales/zh/visual-editor.json
```diff
@@ -1,6 +1,7 @@
 {
   "Heading": "标题",
   "Link": "关联",
+  "a": "一个",
```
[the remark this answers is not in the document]
jwartofsky-yext: I talked to chatGPT and it said the approach would be something like this:

{
  "select": {
    "one": "Select an {{item}}",
    "other": "Select {{count}} {{item}}"
  },
  "item": {
    "image_one": "image",
    "image_other": "images",
    "potato_one": "potato",
    "potato_other": "potatoes"
  }
}

Where image/potato are the different possible entity types

What you have might be fine for now though. This might be handled better once we integrate Smartling

Also just checked and 一个 translates to "one", so that might mean "Select one field" which is close enough

Still not sure if the order makes sense, but I think we might need smartling for that

packages/visual-editor/locales/zh/visual-editor.json
```diff
@@ -94,11 +93,11 @@
     "warning": "此操作将消除您的本地更改。它不能撤消。"
   },
   "emailList": "电子邮件列表",
+  "entityTypeField": "{{entityType}}字段",
```
jwartofsky-yext, line 96: What does this look like after translation on the real page?

Check a few other languages too

jwartofsky-yext: If I translate "Financial Professional Field" to Spanish using Google translate I get: "Campo profesional financiero" 

In the screenshot its "Ubicación campo" which google translates to "Field Location"

For Chinese "Financial Professional Field" translates to 金融专业领域

jwartofsky-yext: That looks good enough to me 👍 

Approved

benlife5:
Updated to remove the label above the selector
Manually updated Spanish and French translations

PR #696, opened 2025-08-06
refactor: use slots for grid

benlife5:
https://github.com/user-attachments/assets/a7fa6de4-49af-42b6-a97f-b7bfb5b84d49

### DIFF
--- a/packages/visual-editor/src/components/layoutBlocks/Grid.tsx
+++ b/packages/visual-editor/src/components/layoutBlocks/Grid.tsx
@@ -118,25 +118,6 @@ export const Grid: ComponentConfig<GridProps> = {
       scope: "gridSection",
     },
   },
-  // resolveData: (data, params) => {
-  //   if (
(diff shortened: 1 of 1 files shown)


benlife5:
Yep! It helped that Watson removed a lot of the Grid options

packages/visual-editor/src/components/layoutBlocks/Grid.tsx
```diff
@@ -15,6 +15,7 @@ import { AnalyticsScopeProvider } from "@yext/pages-components";
 
 export interface GridProps extends layoutProps {
   columns: number;
+  slots: { Column: Slot }[];
```
jwartofsky-yext, line 18: Do we need a migration for this?

benlife5: Thankfully there's not any in-use grids

PR #698, opened 2025-08-07
feat: add translations for CTA link

jwartofsky-yext:
CTA links are now type translatable string and can be set for different locales

### DIFF
--- a/packages/visual-editor/src/internal/puck/constant-value-fields/CallToAction.tsx
+++ b/packages/visual-editor/src/internal/puck/constant-value-fields/CallToAction.tsx
@@ -39,7 +39,14 @@ export const CTA_CONSTANT_CONFIG: CustomField<TranslatableCTA> = {
   render: ({ onChange, value }) => {
     const labelField = useMemo(() => {
       return TranslatableStringField<TranslatableString | undefined>(
-        msg("label", "Label"),
+        msg("fields.label", "Label"),
+        { types: ["type.string"] }
+      );
+    }, []);
+
+    const linkField = useMemo(() => {
+      return TranslatableStringField<TranslatableString | undefined>(
+        msg("fields.link", "Link"),
         { types: ["type.string"] }
       );
     }, []);
@@ -48,11 +55,6 @@ export const CTA_CONSTANT_CONFIG: CustomField<TranslatableCTA> = {
       onChange: onChange,
       value: value,
       fields: [
-        {
-          label: pt("Link", "Link"),
-          field: "link",
-          fieldType: "text",
-        },
         {
           label: pt("linkType", "Link Type"),
           field: "linkType",
@@ -69,6 +71,11 @@ export const CTA_CONSTANT_CONFIG: CustomField<TranslatableCTA> = {
           value={value.label}
           onChange={(newValue) => onChange({ ...value, label: newValue })}
         />
+        <AutoField
+          field={linkField}
+          value={value.link}
+          onChange={(newValue) => onChange({ ...value, link: newValue })}
+        />
         {constantFields}
       </div>
     );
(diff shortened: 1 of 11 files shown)


packages/visual-editor/src/internal/puck/constant-value-fields/CallToAction.tsx
```diff
@@ -39,7 +39,14 @@ export const CTA_CONSTANT_CONFIG: CustomField<TranslatableCTA> = {
...
     const labelField = useMemo(() => {
       return TranslatableStringField<TranslatableString | undefined>(
-        msg("label", "Label"),
+        msg("fields.label", "Label"),
+        { types: ["type.string"] }
+      );
+    }, []);
+
+    const linkField = useMemo(() => {
+      return TranslatableStringField<TranslatableString | undefined>(
```
mkilpatrick, line 48: I wonder if we should have a different type. It will be similar but it's technically not translations. And then if the field gives us the apply to all UI functionality automatically (or w/e we come up with after discussing) that'd be nice.

jwartofsky-yext: added a param to TranslatableStringField "allowApplyAll" which enables the apply all functionality

I set it here on the linkField so that all CTA links have the button

packages/visual-editor/src/editor/TranslatableStringField.tsx
```diff
@@ -6,36 +6,61 @@ import React from "react";
 import { useTranslation } from "react-i18next";
 import { RenderEntityFieldFilter } from "../internal/utils/getFilteredEntityFields.ts";
 import { EmbeddedFieldStringInput } from "./EmbeddedFieldStringInput.tsx";
+import { Button } from "../internal/puck/ui/button.tsx";
 
 /**
  * Generates a translatable string config
  * @param label optional label. Takes in a value from msg.
  * @param filter optional filter for the entity fields that can be embedded.
+ * @param allowApplyAll optional enables the "Apply to All" button
```
asanehisa, line 15: showApplyAllOption ? am fine with allowApplyAll if thats what ppl prefer too. The "optional enables the "Apply to All" button" doesn't quite make sense with the current wording.

packages/visual-editor/src/internal/puck/constant-value-fields/CallToAction.tsx
jwartofsky-yext: I think we could have another type, but I'm not sure what it would be called

It would have the same functionality

Also having the ability to add this "Apply to All" for any TranslatableString field seems useful

So maybe we just roll with it as is for now?

mkilpatrick: sgtm

packages/visual-editor/src/editor/TranslatableStringField.tsx
```diff
@@ -6,36 +6,62 @@ import React from "react";
...
-                : {}),
-              [locale]: val,
-              hasLocalizedValue: "true",
-            } as Record<string, string> as T);
+      const applyAllButton = showApplyAllOption ? (
+        <Button
+          size="sm"
+          variant="small_link"
+          onClick={() => {
+            onChange(resolvedValue as T);
```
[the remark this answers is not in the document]
mkilpatrick: I think we need to, otherwise if you click the button how can you then go back and change it for another lang? Like say I want to apply to all 3 langs but then only es I want something different.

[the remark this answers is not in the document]
jwartofsky-yext: Done

Now sets all locales from the templateMetadata when you click "Apply to All"

mkilpatrick (approved):

benlife5 (approved):

PR #700, opened 2025-08-07
feat: warn when translation keys are deleted

jwartofsky-yext:
Translations are often overwritten when devs run
i18n:update

Example with keys deleted:
image


mkilpatrick (approved):

asanehisa (changes requested):
Actually sorry, this'll trigger a vulnerability like: https://yexttest.atlassian.net/browse/VULN-40060

PR #701, opened 2025-08-07
feat: improve CoreInfoSection styling for tablet

jwartofsky-yext:
In Desktop view, we put the sections horizontally

In Tablet view, put 2 sections per row

In Mobile view, we stack the sections vertically


benlife5 (approved):

packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[desktop] version 1 props with partial entity values 1.png
Comment (mkilpatrick) [inline packages/visual-editor/src/components/testing/screenshots/CoreInfoSection/[desktop] version 1 props with partial entity values 1.png:1]: Hmm, I think on desktop it should still use 2 col when there are 2 items, rather than having an empty col

PR #706, opened 2025-08-08
chore: add TranslatableString type to YextField

asanehisa:
This is helpful for developing with a unified YextField (especially for hybrid developers). Tested in /starter and confirmed translatableStrings worked for multiple locales.


packages/visual-editor/src/editor/CodeField.tsx
```diff
@@ -63,7 +64,7 @@ export const CodeField = ({
             type: "Code",
             value: value,
             id: messageId,
-            fieldName: fieldLabel,
+            fieldName: pt(fieldLabel),
```
mkilpatrick, line 67: This won't translate because there is no string. You can't translate variables, unless I'm wrong in how this is set up.

asanehisa: It is translated with pt in TranslatableRichTextField.tsx prior to being sent via payload as well!

packages/visual-editor/locales/cs/visual-editor.json
```diff
@@ -152,6 +156,7 @@
     "hours": "Hodiny",
     "hoursColumn": "Sloupec hodin",
     "hoursText": "Hodiny text",
+    "html": "HTML",
```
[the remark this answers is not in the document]
asanehisa: similar to CTA, I think HTML should remain HTML is all languages.

packages/visual-editor/src/components/CustomCodeSection.tsx
```diff
@@ -30,15 +30,15 @@ export interface CustomCodeSectionProps {
 }
 
 const customCodeSectionFields: Fields<CustomCodeSectionProps> = {
-  html: YextField("HTML", {
+  html: YextField(msg("fields.html", "HTML"), {
```
jwartofsky-yext, line 33: Since this doesn't actually get translated, I think we can remove this

packages/visual-editor/src/components/CustomCodeSection.tsx
```diff
@@ -30,15 +30,15 @@ export interface CustomCodeSectionProps {
...
     type: "code",
     codeLanguage: "html",
   }),
-  css: YextField("CSS", {
+  css: YextField(msg("fields.css", "CSS"), {
```
jwartofsky-yext, line 37: same

packages/visual-editor/src/components/CustomCodeSection.tsx
```diff
@@ -30,15 +30,15 @@ export interface CustomCodeSectionProps {
...
     type: "code",
     codeLanguage: "css",
   }),
-  javascript: YextField("JavaScript", {
+  javascript: YextField(msg("fields.javascript", "JavaScript"), {
```
jwartofsky-yext, line 41: same

packages/visual-editor/src/components/CustomCodeSection.tsx
asanehisa: Do you think that it'd be better to keep it so YextField enforces msgString type and developers to consciously make them translate the same? Wondering what our logic for translating CTA initially was.

benlife5: oh interesting, I hadn't thought about that there might be labels that wouldn't be translated. I'd lean towards enforcing it and have the translation of HTML be "HTML" in each language (since that's kinda what we're saying -- the word for HTML in each language is HTML), but I can see it both ways

asanehisa: Yeah I was leaning towards enforcing it since I assume thats why we chose to translate "CTA" to "CTA" in every language as well?

asanehisa: @jwartofsky-yext let me know if we want to continue this discussion offline, but if not I think I'll leave it such that YextField requires the label be msgString and if you have a value that should be the same in each language, you should ensure the translations occur properly.

jwartofsky-yext (approved):

PR #708, opened 2025-08-08
feat: update array field item labels

benlife5:
Array field items show the identifying value in the document locale if present. If not, they use the description of the field in the platform locale

Screenshot 2025-08-08 at 4 29 56 PM

### DIFF
--- a/packages/visual-editor/src/components/header/ExpandedHeader.tsx
+++ b/packages/visual-editor/src/components/header/ExpandedHeader.tsx
@@ -139,6 +139,13 @@ const expandedHeaderSectionFields: Fields<ExpandedHeaderProps> = {
               },
             },
             defaultItemProps: defaultMainLink,
+            getItemSummary: (item, i) => {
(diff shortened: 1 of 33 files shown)


packages/visual-editor/src/components/header/ExpandedHeader.tsx
```diff
@@ -139,6 +139,13 @@ const expandedHeaderSectionFields: Fields<ExpandedHeaderProps> = {
               },
             },
             defaultItemProps: defaultMainLink,
+            getItemSummary: (item, i) => {
+              const { i18n } = useTranslation();
+              return (
+                resolveComponentData(item.label, i18n.language) ||
+                pt("Link", "Link") + " " + ((i ?? 0) + 1)
```
mkilpatrick, line 146: Technically this won't translate correctly though it's better than what we had which is no translation at all

jwartofsky-yext (approved):

PR #717, opened 2025-08-12
feat: add Header Position prop to ExpandedHeader

jwartofsky-yext:
If set to "Scrolls with Page", the behavior is unchanged

If set to "Sticky", the header will stick as the user scrolls down

Note that the sticky header does not stick in the editor view, but it does for the live page


packages/visual-editor/src/components/header/ExpandedHeader.tsx
```diff
@@ -393,7 +408,11 @@ const ExpandedHeaderWrapper: React.FC<ExpandedHeaderProps> = ({
   );
 
   return (
-    <>
+    <div
+      className={`flex flex-col ${
+        styles.headerPosition === "sticky" ? "sticky top-0 z-50" : ""
```
benlife5, line 413: Can you confirm that Aaron wants Tailwind's `sticky`? The item says sticky but the description is ambiguous between sticky and fixed

benlife5 (dismissed):

packages/visual-editor/locales/da/visual-editor.json
```diff
@@ -153,6 +153,7 @@
     "footerSocialIcons": "Sidefod social ikoner",
     "format": "Format",
     "headerLinks": "Header Links",
+    "headerPosition": "HEADER POSITION",
```
asanehisa, line 156: uh i don't think this translated haha

asanehisa (approved):

asanehisa (approved):

benlife5 (approved):

PR #726, opened 2025-08-19
fix: adjust atoms to match new styling

asanehisa:
- `Address` has CTA variant as an option 
- `Body Text` supports RTF 
- `Image` uses newer image styling props (same as promo and other section components)
- `Phone` and `Email List` have correct icons 
- `Phone` has the label prop 
- `Get Directions`'s CTA is bolded 

Exported the fields from a lot of the content blocks so they can be used in CoreInfoSection to reduce duplicate code.


packages/visual-editor/src/components/pageSections/PromoSection.tsx
```diff
@@ -30,7 +30,7 @@ import { AnalyticsScopeProvider } from "@yext/pages-components";
 import {
   ImageStylingFields,
   ImageStylingProps,
-} from "../contentBlocks/ImageStyling.js";
+} from "../contentBlocks/image/styling.ts";
```
mkilpatrick, line 33: Remind me why this is imported directly rather than from the package like all the other components above.

packages/visual-editor/src/components/contentBlocks/Phone.tsx
```diff
@@ -10,46 +9,63 @@ import {
...
+      }),
     },
   }),
-  format: YextField(msg("fields.format", "Format"), {
-    type: "radio",
-    options: "PHONE_OPTIONS",
+  styles: YextField(msg("fields.styles", "Styles"), {
+    type: "object",
+    objectFields: {
+      format: YextField(msg("fields.format", "Format"), {
```
mkilpatrick, line 46: Is there a way to better share these? This is almost identical to what's in CoreInfoSection, though the text/translation is slightly different. Same would apply to everything else. Like both this Phone contentBlock and CoreInfoSection make use of the PhoneAtom, but each duplicates the data/styles props to send to the underlying atom. Is there a better way to avoid duplication?

packages/visual-editor/src/components/pageSections/PromoSection.tsx
asanehisa: Oh I am also unsure. I assume it is b/c we didn't want this ImageStylingFields and ImageStylingProps to be exported in the package, but I feel like they could be?

packages/visual-editor/src/components/testing/screenshots/Grid/[desktop] version 18 - atoms used to make a HeroSection.png
Comment (asanehisa) [inline packages/visual-editor/src/components/testing/screenshots/Grid/[desktop] version 18 - atoms used to make a HeroSection.png:1]: looking into this test and why image disappeared (but not in tablet or mobile view)

packages/visual-editor/src/components/contentBlocks/Phone.tsx
asanehisa: Added some consts so not all of the Fields are repeated everytime! They are defined in the contentBlocks and used by CoreInfoSection

packages/visual-editor/src/components/pageSections/CoreInfoSection.tsx
```diff
@@ -32,19 +30,31 @@ import {
...
+import {
+  AddressDataField,
+  AddressProps,
+  AddressStyleFields,
+} from "../contentBlocks/Address.tsx";
+import {
+  PhoneDataFields,
+  PhoneProps,
+  PhoneStyleFields,
+} from "../contentBlocks/Phone.tsx";
```
mkilpatrick, line 44: What do you think about these living in the atoms instead? Does that make more or less sense?

asanehisa: I think less sense as the /atoms/ don't use Puck fields at all or YextEntityField<...> types

mkilpatrick (approved):

briantstephan (approved):

PR #729, opened 2025-08-20
fix: heroSection variants properly resolve struct fields

asanehisa:
HeroSection and PromoSection both need to be resolved with `resolveYextStructField` for the static value overrides to work. Confirmed each Hero variant allowed toggling (from entity field to static value) of the various Hero values without problems.


mkilpatrick (approved):

benlife5 (commented):
LGTM, can you double check that embedded fields work, I might be remembering something else but I feel like there's been some back and forth between `resolveComponentData` and `resolveYextStructField`

asanehisa:
> LGTM, can you double check that embedded fields work, I might be remembering something else but I feel like there's been some back and forth between `resolveComponentData` and `resolveYextStructField`

Yup it works under the hood of resolveYextStructField!

benlife5 (approved):

PR #730, opened 2025-08-20
fix: remove "flex-reverse" usage

asanehisa:
Removed "flex-reverse" usages to be more accessible. 

Also fixes the screenshot tests for ExpandedFooter's mobile tests as they were cutting off a lot of the visual data.


mkilpatrick (dismissed):

packages/visual-editor/src/components/footer/ExpandedFooter.tsx
```diff
@@ -479,10 +479,10 @@ const ExpandedFooterWrapper = ({
         verticalPadding={"footer"}
         background={backgroundColor}
         maxWidth={maxWidth}
-        className={`flex flex-col ${primaryLinksAlignment === "right" ? `md:flex-row` : `md:flex-row-reverse`} md:justify-start w-full md:items-start gap-8 md:gap-10`}
+        className={`flex flex-col md:flex-row md:justify-start w-full md:items-start gap-8 md:gap-10`}
       >
         <div
-          className={`flex flex-col gap-10 md:gap-8 ${primaryLinksAlignment === "left" ? `items-end` : `items-start`}`}
+          className={`flex flex-col gap-10 md:gap-8 ${primaryLinksAlignment === "left" ? `items-end md:order-2` : `items-start md:order-1`}`}
```
benlife5, line 485: Unfortunately I think this creates the same accessibility problem
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flexible_box_layout/Ordering_flex_items#the_order_property_and_accessibility

asanehisa: Oops mb for not reading those docs closely enough. Fixed similar to how you did for Hero with two of the html elements and hiding the unused one.

packages/visual-editor/src/components/footer/ExpandedFooter.tsx
```diff
@@ -473,61 +474,71 @@ const ExpandedFooterWrapper = ({
...
-                logoWidth={logoWidth || 100}
-              />
+              <div className="grid grid-cols-3 gap-8">
+                {utilityImages.map((item, index) => (
+                  <FooterLogo
+                    aspectRatio={aspectRatioForUtilityImages}
+                    key={index}
+                    logo={buildComplexUtilityImage(
+                      item.url,
+                      utilityImagesWidth || 60
```
jwartofsky-yext, line 516: Did we want to change the fallback width to 60? When does 60 actually get used?

asanehisa: Oh yeah the fallback width is 60 for desktop / tablet utility images so i made it the same for the mobile utility images. It's just if the image given has no width set (which shouldn't happen once we switch to new image assets i think)

benlife5: This width is set in the styles props section so it can be 0, which is effectively an "auto" mode that uses the fallbacks set for each usage like this 60

PR #742, opened 2025-08-28
chore: update fonts to match platform

briantstephan:
This updates the `--puck-font-family` variable to use the font stack used in platform. Verified that fonts are updated in both `pages-visual-editor-starter` and `/starter`.

Screenshot 2025-08-28 at 1 00 23 PM


briantstephan:
> Can you add a screenshot?

Added. Puck actually already uses Inter so it should look pretty much the same as before, but the full font stack now matches our platform so it should be more consistent in some edge cases.

benlife5 (approved):

jwartofsky-yext (approved):
lgtm

Does it work without !important?

PR #746, opened 2025-08-28
fix: prevent HeroSection image flickering

briantstephan:
Defining the Image components inside the Hero components was causing the image on the HeroSection to flicker as it got re-defined every re-render. Verified manually that the image no longer flickers when hovering over it.

https://github.com/user-attachments/assets/c395d076-9f03-472c-b774-b54cd4e6770b


benlife5 (approved):
Ah yep, thanks

jwartofsky-yext (approved):

mkilpatrick (approved):

PR #748, opened 2025-09-02
fix: handle event card description overflow

briantstephan:
If the description text for an event overflows the event card in `EventSection`, this makes the card expand to show the full description and CTA. In this case, whitespace is added below the image.

There is also a new "Truncate Description" prop, enabled by default, which line-clamps the description text to prevent the card from extending.

Screenshot 2025-09-18 at 3 01 24 PM

Screenshot 2025-09-18 at 3 01 38 PM


packages/visual-editor/src/components/testing/screenshots/EventSection/[desktop] version 7 props with entity values.png
Comment (mkilpatrick) [inline packages/visual-editor/src/components/testing/screenshots/EventSection/[desktop] version 7 props with entity values.png:1]: Idk if the aspect ratio should have changed on the image here. Also replied in slack in general about the image padding.

Comment (briantstephan) [reply packages/visual-editor/src/components/testing/screenshots/EventSection/[desktop] version 7 props with entity values.png:1]: Updated based on feedback. The aspect ratio never changes now.

mkilpatrick (approved):

PR #753, opened 2025-09-04
fix: category filtering

benlife5:
In the latest artifact, the Custom Code Section appears in the sidebar even with the feature flag off however you cannot drag it onto the page. Tested this change in platform dev mode and confirmed that the feature flags worked as expected.

I thought this used to work but here's an artifact 0.0.17 site, which is when we introduced the Custom Code Component, and it shows up even with the feature flag off: https://dev.yext.com/s/1000152098/yextsites/65518/editor#pageSetId=locations&locale=en&themeId=artifact-0-0-20&entityId=8725521

We might want to release a new artifact since this is affecting GA.


benlife5:
Added tests, then had gemini condense some of the steps. Test in dev mode again

packages/visual-editor/src/utils/filterComponents.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+} from "../components/_componentCategories.ts";
+
+// TODO(SUMO-7666): Replace with importing the actual main config from our exported ve.config.tsx
+interface MainProps
+  extends PageSectionCategoryProps,
+    DeprecatedCategoryProps,
+    OtherCategoryProps,
+    AdvancedCoreInfoCategoryProps {}
+
+const components: Config<MainProps>["components"] = {
```
mkilpatrick, line 26: Does this file need to be .tsx because of the components? It's weird that the test is tsx but the file is ts.

benlife5: It needed it for the root render function, but we don't need a render function for this test. Removed.

jwartofsky-yext (approved):

PR #758, opened 2025-09-10
feat: handle embedded fields in html component

asanehisa:
Allows for handlebars usage in the HTML such as
```
<ul class="product-list">
  {{#each c_exampleProducts.products}}
    <li class="product-item">
      <strong>{{name}}</strong>
      {{#if image.url}}
        <br />
        <img src="{{image.url}}" class="product-img" alt="{{name}} alt"/>
      {{/if}}
      {{#if description.html}}
        <div class="product-desc">{{{description.html}}}</div>
      {{/if}}
    </li>
  {{/each}}
</ul>
```
Or just 
`<p> {{name}} </p>`

And normal embedded field use in the javascript allowing for data to be used in `[[]]` notation. 
```
function showName(el) {
  console.log([[name]]);
}
```

DOMPurify was removed from the HTML code cleanup to prevent erasing purposeful code written by the user such as 
```
<button id="color-btn" onclick="changeColor(this)">
  Change Color
</button>
```
which with DOMPurify became
```
<button id="color-btn">
  Change Color
</button>
```

Production Site: https://www.yext.com/s/4412098/yextsites/160429/pagesets

### DIFF
--- a/packages/visual-editor/src/components/CustomCodeSection.tsx
+++ b/packages/visual-editor/src/components/CustomCodeSection.tsx
@@ -132,14 +132,10 @@ const CustomCodeSectionWrapper = ({
   const scriptIdRef = React.useRef<number>(Math.floor(Math.random() * 1e9));
   const scriptTagId = `custom-code-section-script-${scriptIdRef.current}`;
 
-  const processedHtml = React.useMemo(
-    () => processHandlebarsTemplate(html, streamDocument),
-    [html, streamDocument.id]
-  );
-
-  const processedJavascript = React.useMemo(
-    () => resolveEmbeddedFieldsInString(javascript, streamDocument),
-    [javascript, streamDocument.id]
+  const processedHtml = processHandlebarsTemplate(html, streamDocument);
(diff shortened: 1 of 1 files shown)


packages/visual-editor/src/components/CustomCodeSection.tsx
```diff
@@ -90,14 +125,28 @@ const CustomCodeSectionWrapper = ({
...
   const scriptTagId = `custom-code-section-script-${scriptIdRef.current}`;
 
+  const processedHtml = React.useMemo(
+    () => processHandlebarsTemplate(html, stableStreamDoc),
+    [html, stableStreamDoc]
+  );
+
+  // Process Handlebars in JavaScript string
+  const processedJavascript = React.useMemo(
+    () => processHandlebarsTemplate(javascript, stableStreamDoc),
```
mkilpatrick, line 146: Handlebars isn't a js syntax so I don't think we want to use it here. But we do need to allow users to add embedded fields to the js. Ideally it's the same bracket notation as in our RTF fields.

PR #762, opened 2025-09-11
feat: component types documentation for AI

benlife5:
Filters the package's `.d.ts` file to the PageSection/ExpandedHeader/ExpandedFooter props and the types used within those props. Writes those types to a `docs/ai/components.d.ts` during the generate-docs Github Action. Hopefully should be a good source for the auto-layout LLM to understand the available options

Also adds additional documentation where missing -- I imagine this will need additional tweaking.

Tentatively introduced an `@ai` jsdoc tag which is meant to be a specific instruction for the LLM.


packages/visual-editor/src/components/contentBlocks/Address.tsx
```diff
@@ -25,7 +25,7 @@ export interface AddressProps {
     address: YextEntityField<AddressType>;
   };
   styles: {
-    /** Includes a "Get Directions" CTA to Google Maps */
+    /** Whether to included a "Get Directions" CTA to Google Maps */
```
asanehisa, line 28: very nit: "Whether to include"

PR #770, opened 2025-09-18
fix: resolve base entity url templates in directory pages

benlife5:
For directory pages, Spruce has added the base entity's url template under `__.entityPageSetUrlTemplates`

### DIFF
--- a/packages/visual-editor/src/utils/applyTheme.ts
+++ b/packages/visual-editor/src/utils/applyTheme.ts
@@ -23,6 +23,7 @@ export type StreamDocument = {
     name?: string;
     visualEditorConfig?: string;
     isPrimaryLocale?: boolean;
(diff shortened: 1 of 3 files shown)


packages/visual-editor/src/utils/applyTheme.ts
```diff
@@ -23,6 +23,7 @@ export type StreamDocument = {
     name?: string;
     visualEditorConfig?: string;
     isPrimaryLocale?: boolean;
+    entityPageSetUrlTemplates?: string;
```
mkilpatrick, line 26: Not you, but this type feels weird to be is this theme file

mkilpatrick: ack

PR #771, opened 2025-09-18
feat: send schema markup message to YSS

colton-demetriou:
Screenshot 2025-09-18 at 3 07 55 PM

Looks slightly different than the mock, but uses the same components as the custom code section.

### DIFF
--- a/packages/visual-editor/src/internal/components/AdvancedSettings.tsx
+++ b/packages/visual-editor/src/internal/components/AdvancedSettings.tsx
@@ -1,5 +1,5 @@
 import { ComponentConfig, Fields, CustomField } from "@measured/puck";
-import { msg, YextField, useDocument } from "@yext/visual-editor";
+import { msg, YextField } from "@yext/visual-editor";
 import React from "react";
 import {
   useSendMessageToParent,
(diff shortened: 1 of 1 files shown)


packages/visual-editor/src/internal/components/AdvancedSettings.tsx
```diff
@@ -14,14 +19,161 @@ export interface AdvancedSettingsProps {
...
+    }
+  }
+}`;
+
+const FALLBACK_SCHEMA = `{
+  "@context": "https://schema.org",
+  "@type": "Thing",
+  "name": "[[name]]",
+  "description": "[[description]]",
+  "url": "[[urlWriteback]]"
```
benlife5, line 64: Same with this, the field isn't `urlWriteback` but instead a specific field for the page set. Might be able to find on the document

packages/visual-editor/src/internal/components/AdvancedSettings.tsx
```diff
@@ -14,14 +19,161 @@ export interface AdvancedSettingsProps {
...
+  "image": "[[photoGallery]]",
+  "description": "[[description]]",
+  "telephone": "[[mainPhone]]",
+  "paymentAccepted": "[[paymentOptions]]",
+  "hasOfferCatalog": "[[services]]"
+}`;
+
+const DIRECTORY_LIST_ITEM_SCHEMA = `{
+  "@type": "ListItem",
+  "position": "[[index + 1]]",
```
benlife5, line 44: index isn't going to be a field on the document, it's from spruce's preprocessing. I think for directory we can just manually list out the 4 levels of our DMs

benlife5 (commented):
I think there's still some stuff to work out with the default data but the drawer trigger LGTM so I'd be fine with merging this to unblock the YSS items

packages/visual-editor/src/internal/components/AdvancedSettings.tsx
colton-demetriou: resolving because doing different approach

PR #772, opened 2025-09-18
fix: don't treat advanced settings as a component

colton-demetriou:
Screenshot 2025-09-18 at 6 30 42 PM
treating advanced settings as a component was breaking drag and drop functionality for other components, it also appeared in the page outline which was not ideal. We will need to use a back button here given we cannot have breadcrumbs without treating it as a component.

### DIFF
--- a/packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
+++ b/packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
@@ -39,54 +39,19 @@ const createAdvancedSettingsLink = () => ({
       <div className="ve-p-4 ve-flex ve-justify-center ve-items-center">
         <button
           onClick={() => {
-            const { appState, dispatch } = getPuck();
-            const advancedSettingsId = "AdvancedSettings";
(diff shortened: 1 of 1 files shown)


packages/visual-editor/src/internal/components/InternalLayoutEditor.tsx
```diff
@@ -316,6 +279,14 @@ export const InternalLayoutEditor = ({
                 if (schemaField.type === "custom" && schemaField.render) {
                   return (
                     <div className="ve-p-4 ve-mb-4">
+                      {/* Title */}
+                      <div className="ve-mb-4">
+                        <h3 className="ve-text-md ve-font-semibold ve-text-gray-900 ve-mb-2">
```
asanehisa, line 284: is this h3 for a reason?

colton-demetriou: you're right not needed

colton-demetriou:
> Test failed. Unclear if related.

it seems like an issue with the screenshots, which should be unrelated

PR #790, opened 2025-09-29
feat: add heading text slot

benlife5:
Updates EventSection to use a slot component for the section heading text. Refactors HeadingText to include an align prop and organize props into data/styles

Will update the rest of EventSection (+tests) in separate PR but heading will be a shared component

Used a SlotsCategory because it seems like `resolvePermissions` doesn't support checking if the component is in a slot or not. Will start a thread with the Puck team because that would be easier.

https://github.com/user-attachments/assets/b9758171-3f8e-4ba8-8f31-1ee1d18ceb26


packages/visual-editor/src/components/pageSections/EventSection.tsx
```diff
@@ -87,6 +71,10 @@ export interface EventSectionProps {
    */
   styles: EventStyles;
 
+  slots: {
```
benlife5, line 74: @asanehisa not sure if you have opinions on naming conventions/prop organization for slots. Essentially we need an array for each slot somewhere in the data object

benlife5:
At the risk of merge conflicts, I'm thinking we may want to have a separate branch for these changes in case we want/need to push out other fixes before the whole set of slot changes is ready

packages/visual-editor/src/components/testing/screenshots/EventSection/[desktop] default props with document data.png
Comment (benlife5) [inline packages/visual-editor/src/components/testing/screenshots/EventSection/[desktop] default props with document data.png:1]: lol migration clearly did not work, taking a look

packages/visual-editor/src/components/pageSections/EventSection.tsx
asanehisa: But I do appreciate having the Heading Slot split into data + styles! I think that'd be nice to keep for the slot atoms

packages/visual-editor/src/components/configs/mainConfig.tsx
```diff
@@ -49,6 +56,11 @@ export const mainConfig: Config<MainConfigProps> = {
       title: "Other",
       components: OtherCategory,
     },
+    slots: {
+      title: "Slots",
```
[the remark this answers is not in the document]
benlife5: oop, updated

asanehisa (approved):

colton-demetriou (approved):

PR #791, opened 2025-09-29
feat: provide default schema to parent window

briantstephan:
When sending the user's schema to the parent to allow for editing, we now also send the default schema. A new option in the UI will allow the user to restore the default schema and clear any edits they've made.

Ran everything locally and logged the payload that was sent to the front-end, confirming that it included the new defaultValue field.

### DIFF
--- a/packages/visual-editor/src/internal/components/AdvancedSettings.tsx
+++ b/packages/visual-editor/src/internal/components/AdvancedSettings.tsx
@@ -45,9 +45,11 @@ const SCHEMA_MARKUP_FIELD: CustomField<string> = {
       }
     );
 
+    const defaultSchema = getSchemaTemplate(entityTypeId);
+    const cleanDefaultSchema = defaultSchema.replace(/\n\s*/g, " ").trim();
+
     // Use the schema value from root, or default schema if not set
-    const displayValue =
-      value || (entityTypeId ? getSchemaTemplate(entityTypeId) : "");
+    const displayValue = value || (entityTypeId ? defaultSchema : "");
 
     const codeField = YextField(msg("schemaMarkup", "Schema Markup"), {
       type: "code",
@@ -75,6 +77,7 @@ const SCHEMA_MARKUP_FIELD: CustomField<string> = {
         const payload = {
           type: "SchemaMarkup",
           value: cleanSchemaValue,
+          defaultValue: cleanDefaultSchema,
           id: messageId,
         };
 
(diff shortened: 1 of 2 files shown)


packages/visual-editor/src/internal/components/AdvancedSettings.tsx
```diff
@@ -45,9 +45,11 @@ const SCHEMA_MARKUP_FIELD: CustomField<string> = {
       }
     );
 
+    const defaultSchema = getSchemaTemplate(entityTypeId);
+    const cleanDefaultSchema = defaultSchema.replace(/\n\s*/g, " ").trim();
```
asanehisa, line 49: Hmm clean is an interesting word, maybe change var name or add comment to clarify intent?

colton-demetriou: you could rename the var on line 48 to "formattedDefaultSchema" or something then rename the other to just "defaultSchema", since the first one is the one we use for the display

asanehisa (approved):

benlife5 (dismissed):

packages/visual-editor/src/internal/components/AdvancedSettings.tsx
```diff
@@ -45,9 +45,14 @@ const SCHEMA_MARKUP_FIELD: CustomField<string> = {
...
+    const defaultSchema = getSchemaTemplate(entityTypeId);
+    // Compact the default schema to remove newlines and extra whitespace
+    const compactDefaultSchema = defaultSchema.replace(/\n\s*/g, " ").trim();
+
     // Use the schema value from root, or default schema if not set
-    const displayValue =
-      value || (entityTypeId ? getSchemaTemplate(entityTypeId) : "");
+    const schema = value || (entityTypeId ? defaultSchema : "");
+    // Compact the schema to remove newlines and extra whitespace
+    const compactSchema = schema.replace(/\n\s*/g, " ").trim();
```
mkilpatrick, line 55: Is value (schema) already compacted if it was previously saved?

briantstephan: Yeah it should be, this extra check is just to be safe.

briantstephan: Updated to make `getSchemaTemplate` to the replacements and use the non-compact versions everywhere.

packages/visual-editor/src/utils/getSchema.ts
```diff
@@ -79,15 +79,15 @@ const FALLBACK_SCHEMA = `{
 // Function to get the appropriate schema template based on entity type
 export const getSchemaTemplate = (entityTypeId?: string): string => {
   if (!entityTypeId) {
-    return FALLBACK_SCHEMA;
+    return FALLBACK_SCHEMA.replace(/\n\s*/g, " ").trim();
```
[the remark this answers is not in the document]
mkilpatrick: Or just set to a var instead of returning and return the final this with this final replacement.

mkilpatrick (approved):

PR #793, opened 2025-10-01
fix: fonts not loading in editor

colton-demetriou:
Screenshot 2025-10-01 at 1 59 29 PM
Confirmed that fonts now load correctly in both the Theme and Layout Editor and match the live page. 

The problem was that the live page used applyTheme but the editor was only using updateThemeInEditor, so when the font was changed in the editor, the new styles were applied but the actual font files weren't loaded, leading to the editor falling back to defaults. The fix was essentially to just use the applyTheme logic in the editor as well.

Test Site: https://www.yext.com/s/4259018/yextsites/160744/theme#themeId=font-test-2&pageSetId=font-test-2&locale=en&entityId=1095527849

### DIFF
--- a/packages/visual-editor/src/utils/applyTheme.ts
+++ b/packages/visual-editor/src/utils/applyTheme.ts
@@ -135,41 +135,22 @@ const generateContrastingColors = (themeData: ThemeData) => {
   return contrastingColors;
 };
 
-// Helper function to update font links in the editor document
-const updateFontLinksInEditor = (fontLinkTags: string) => {
-  const existingLinks = window.document.querySelectorAll(
-    'link[href*="fonts.googleapis.com"]'
-  );
-  existingLinks.forEach((link) => link.remove());
-
-  if (fontLinkTags) {
-    const tempDiv = document.createElement("div");
-    tempDiv.innerHTML = fontLinkTags;
-    const links = tempDiv.querySelectorAll("link");
-    links.forEach((link) => {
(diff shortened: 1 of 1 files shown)


packages/visual-editor/src/utils/applyTheme.ts
```diff
@@ -135,18 +135,72 @@ const generateContrastingColors = (themeData: ThemeData) => {
...
+
+const updateFontLinksInIframe = (
+  iframe: HTMLIFrameElement,
+  fontLinkTags: string
+) => {
+  if (!iframe.contentDocument) {
+    return;
+  }
+
+  const existingLinks = iframe.contentDocument.querySelectorAll(
```
benlife5, line 163: Can you combine this with 140-152 and accept the relevant document as a param?

packages/visual-editor/src/utils/applyTheme.ts
```diff
@@ -135,18 +135,72 @@ const generateContrastingColors = (themeData: ThemeData) => {
   return contrastingColors;
 };
 
+// Helper function to update font links in the editor document
+const updateFontLinksInEditor = (fontLinkTags: string) => {
```
benlife5, line 139: Does the font selector dropdown still render all the font names in their own font?

colton-demetriou: i'll make a separate follow up for that

briantstephan (approved):

PR #796, opened 2025-10-01
fix: load fonts in font selector

colton-demetriou:
Screenshot 2025-10-01 at 4 45 11 PM

The problem was when the editor first loaded no google fonts were loaded (only theme-specific fonts), fixed by loading the fonts on initial load. Verified in local dev that the fonts loaded in the editor on the page and in the selector.

### DIFF
--- a/packages/visual-editor/src/internal/hooks/useMessageReceivers.ts
+++ b/packages/visual-editor/src/internal/hooks/useMessageReceivers.ts
@@ -12,6 +12,7 @@ import { ThemeData } from "../types/themeData.ts";
 import { migrate } from "../../utils/migrate.ts";
 import { migrationRegistry } from "../../components/migrations/migrationRegistry.ts";
 import { filterComponentsFromConfig } from "../../utils/filterComponents.ts";
+import { googleFontLinkTags } from "../../utils/visualEditorFonts.ts";
 
 const devLogger = new DevLogger();
 
@@ -21,9 +22,20 @@ export const useCommonMessageReceivers = (
 ) => {
   const { iFrameLoaded } = useCommonMessageSenders();
 
-  // Trigger data flow from parent
+  // Trigger data flow from parent and load initial fonts
   useEffect(() => {
     iFrameLoaded({ payload: { message: "iFrame is loaded" } });
+
+    // Load default Google Fonts for the font selector dropdown
+    if (!document.getElementById("visual-editor-default-fonts")) {
+      const tempDiv = document.createElement("div");
+      tempDiv.innerHTML = googleFontLinkTags;
+      const links = tempDiv.querySelectorAll("link");
+      links.forEach((link) => {
+        link.id = "visual-editor-default-fonts";
+        document.head.appendChild(link);
+      });
+    }
   }, []);
 
   // Base Template Info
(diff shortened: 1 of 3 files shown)


packages/visual-editor/src/internal/hooks/useMessageReceivers.ts
```diff
@@ -21,9 +22,20 @@ export const useCommonMessageReceivers = (
...
+  // Trigger data flow from parent and load initial fonts
   useEffect(() => {
     iFrameLoaded({ payload: { message: "iFrame is loaded" } });
+
+    // Load default Google Fonts for the font selector dropdown
+    if (!document.getElementById("visual-editor-default-fonts")) {
+      const tempDiv = document.createElement("div");
+      tempDiv.innerHTML = googleFontLinkTags;
+      const links = tempDiv.querySelectorAll("link");
+      links.forEach((link) => {
```
mkilpatrick, line 34: ids need to be unique

packages/visual-editor/src/internal/hooks/useMessageReceivers.ts
```diff
@@ -21,9 +22,20 @@ export const useCommonMessageReceivers = (
 ) => {
   const { iFrameLoaded } = useCommonMessageSenders();
 
-  // Trigger data flow from parent
+  // Trigger data flow from parent and load initial fonts
   useEffect(() => {
     iFrameLoaded({ payload: { message: "iFrame is loaded" } });
+
+    // Load default Google Fonts for the font selector dropdown
+    if (!document.getElementById("visual-editor-default-fonts")) {
```
mkilpatrick, line 30: Seems weird that this is in the message receiver code

packages/visual-editor/src/internal/utils/loadMapboxIntoIframe.tsx
```diff
@@ -32,6 +33,17 @@ export const loadMapboxIntoIframe = ({
       link.rel = "stylesheet";
       document.body.appendChild(link);
     }
+
+    // Load default Google Fonts for the font selector dropdown
+    if (!document.getElementById("visual-editor-default-fonts")) {
```
mkilpatrick, line 38: We need this here too? If so share the func.

packages/visual-editor/src/internal/hooks/useMessageReceivers.ts
colton-demetriou: moved to Editor.tsx

packages/visual-editor/src/internal/utils/loadMapboxIntoIframe.tsx
colton-demetriou: done

packages/visual-editor/src/utils/applyTheme.ts
```diff
@@ -141,7 +141,7 @@ const updateFontLinksInDocument = (
   fontLinkTags: string
 ) => {
   const existingLinks = document.querySelectorAll(
-    'link[href*="fonts.googleapis.com"]'
+    'link[href*="fonts.googleapis.com"]:not([id^="visual-editor-default-fonts"]):not([id^="visual-editor-iframe-fonts"])'
```
mkilpatrick, line 144: This only works because the passed idPrefix matches. If it's a variable and you used a different string then this would fail.

PR #1022, opened 2026-02-03
feat: add puck ai plugin

benlife5:
This PR adds the Puck AI plugin and successfully forwards requests to the Puck cloud server. It is not connected to any components so it cannot build any layouts, but the chat works.

Adds a Cloudflare worker using a Hono backend to proxy the Puck AI requests. The worker can be run locally or deployed.

To run: 
1. packages/puck-ai-backend:  `pnpm run dev` with `PUCK_API_KEY` in `.dev.vars`
2. packages/starter: `pnpm run dev:ai` to run 

Codex made some package changes to separate the tailwind imports from the rest of the package so that the postcss build issues pass. They seem reasonable and the starter works but we should verify in-platform


pnpm-lock.yaml
```diff
@@ -8103,6 +8302,12 @@ packages:
         integrity: sha512-NM8/P9n3XjXhIZn1lLhkFaACTOURQXjWhV4BA/RnOv8xvgqtqpAX9IO4mRQxSx1Rlo4tqzeqb0sOlruaOy3dug==,
       }
 
+  json-schema@0.4.0:
```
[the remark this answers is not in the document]
benlife5: https://yext.atlassian.net/servicedesk/customer/portal/56/PSR-173

packages/visual-editor/src/docs/ai/components.d.ts
benlife5: We should probably just remove this file since this is not the AI direction we are going in

packages/ai-backend/src/server.ts
```diff
@@ -0,0 +1,220 @@
...
+      res.setHeader(key, value);
+    });
+  };
+
+  try {
+    for await (const chunk of reader) {
+      const chunkText = chunk.toString("utf8");
+      buffer += chunkText;
+
+      const errorText = getSseErrorText(buffer);
```
mkilpatrick: What's Sse?

mkilpatrick (commented):
- If you can't remove wrangler we can see if Vaccine will allow it since it's well-known.
- Robot nits

PR #1146, opened 2026-04-08
feat: use paths for custom fonts

jwartofsky-yext:
This updates the custom font flow to use the expanded `customFonts` metadata directly instead of reconstructing asset paths in VLE.

### Changes

- Treat `templateMetadata.customFonts[fontFamily].facePath` as the source of truth for custom font stylesheet loading.
- Treat `templateMetadata.customFonts[fontFamily].variants[].filePath` as the source of truth for custom font preloads.
- Remove the old CSS fetch/parsing flow that indexed `@font-face` rules to rediscover `.woff2` files.
- Use the unified `FontRegistry` shape for both built-in and custom fonts, keyed by CSS font family name.
- Save resolved custom font assets into one theme field:
  - `__customFontAssets.stylesheetPaths`
  - `__customFontAssets.preloads`
- On the live page, load custom font stylesheets from `__customFontAssets.stylesheetPaths` and preload font files from `__customFontAssets.preloads`.
- Simplify the theme editor flow so custom font asset resolution happens in one pass and saves/removes one object instead of separate keys.
- Consolidate shared font helpers so asset path normalization, theme font-family resolution, and font loading decisions are handled in one place.

### Net Result

- less path reconstruction
- no CSS parsing in VLE for custom font preloads
- simpler and more standardized custom font logic
- one saved theme asset object instead of split custom font keys
- saved theme metadata is now the source of truth for live-page custom font assets


packages/visual-editor/src/utils/fonts/visualEditorFonts.ts
```diff
@@ -22,17 +24,45 @@ type FontSpecification = {
...
   const fontOptions: StyleSelectOption[] = [];
   for (const fontName in fonts) {
     const fontDetails = fonts[fontName];
+    const displayName = getFontDisplayName(fontName, fontDetails);
     fontOptions.push({
-      label: fontName,
-      value: `'${fontName}', '${fontName} Fallback', ${fontDetails.fallback}`,
+      label: displayName,
+      value: `'${displayName}', '${displayName} Fallback', ${fontDetails.fallback}`,
     });
```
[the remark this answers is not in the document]
jwartofsky-yext: The custom font will have the same key and displayName

The displayName and name fields are used to be more accurate

Eventually we would like to remove all usages of the key

packages/visual-editor/src/utils/fonts/visualEditorFonts.ts
```diff
@@ -382,3 +434,25 @@ export const extractInUseFontFamilies = (
...
+  displayNames.forEach((displayName) => {
+    for (const [fontKey, fontDetails] of Object.entries(customFonts)) {
+      if (getFontDisplayName(fontKey, fontDetails) === displayName) {
+        matchedCustomFonts[fontKey] = fontDetails;
+        break;
+      }
+    }
+  });
+
+  return matchedCustomFonts;
```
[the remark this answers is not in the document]
jwartofsky-yext: Storm has already been updated to pass name and displayName for all custom fonts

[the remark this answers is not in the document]
jwartofsky-yext: The key and displayName are the same

We update to use the displayName now, but the older versions used the key. The key and displayName have always matched though, so this shouldn't be a concern

packages/visual-editor/scripts/generateFontRegistry.js
```diff
@@ -44,9 +44,14 @@ const getFontData = async () => {
 };
 
 const parseFontDetails = (font) => {
+  const baseDetails = {
```
asanehisa: this works but any reason baseDetails is an object?

packages/visual-editor/src/utils/applyTheme.ts
```diff
@@ -185,16 +190,19 @@ let pendingObserver: MutationObserver | null = null;
 export const updateThemeInEditor = async (
```
asanehisa, line 202: Hmm do we need to pass customFonts in ? We can't just extract it in `extractInUseFontFamilies`

packages/visual-editor/src/utils/index.ts
```diff
@@ -43,6 +43,7 @@ export {
   defaultFonts,
   getFontWeightOptions,
   getFontWeightOverrideOptions,
+  type FontSpecification,
```
asanehisa: should this be exported?

packages/visual-editor/src/utils/fonts/visualEditorFonts.ts
```diff
@@ -25,18 +27,31 @@ type FontSpecification = {
...
+export const defaultFonts = fontsJs as FontRegistry;
 
+/**
+ * Finds the font specification whose display name matches the font-family name
+ * stored in theme CSS values.
+ */
+export const findFontByDisplayName = (
+  fonts: FontRegistry,
+  displayName: string
+) => {
```
briantstephan: Add return type

packages/visual-editor/src/utils/fonts/visualEditorFonts.ts
```diff
@@ -25,18 +27,31 @@ type FontSpecification = {
...
+    (fontDetails) => fontDetails.displayName === displayName
+  );
+};
+
+/**
+ * Builds font-family select options from a font registry. The label and CSS
+ * font-family value are both derived from the font's display name so theme data
+ * remains human-readable.
+ */
 export const constructFontSelectOptions = (fonts: FontRegistry) => {
```
briantstephan: Add return type

packages/visual-editor/src/utils/fonts/visualEditorFonts.ts
```diff
@@ -110,13 +125,46 @@ export const generateGoogleFontLinkData = (
...
+/**
+ * Derives the family-level CSS id for a custom font by removing the final
+ * hyphen-delimited subfamily segment from `FontSpecification.name`.
+ *
+ * `getCustomFontCssId` expects names shaped like
+ * `<font-family-id>-<subfamily>`, where the trailing segment represents a
+ * subfamily or weight such as `regular` or `bold`. For example,
+ * `ebbmelvynregular-regular` becomes `ebbmelvynregular`.
+ */
+const getCustomFontCssId = (fontName: string) => {
```
briantstephan: nit, `fontName` is a little confusing when we're also dealing with display names here - maybe `fontId` or `fontAssetId` or something would be better? feel free to ack

packages/visual-editor/scripts/generateFontRegistry.js
jwartofsky-yext: originally I was going to put "name" in here as-well, so this would keep things easier to read as you would have something like:
```
  const baseDetails = {
    name: font.family
    .toLowerCase()
    .replace(/[^a-z0-9]/g, ''),
    displayName: font.family,
  };
```
Then we can make name required in visualEditorFont.ts line 12

but the name is not actually used for the built-in fonts, so that just seemed confusing

I think the baseDetails makes it a little more clear that we are adding displayName: font.family to all of the built-ins though, so I think it's ok to keep it

packages/visual-editor/src/utils/fonts/visualEditorFonts.ts
jwartofsky-yext: the name that is being passed in here derives from the resource name from FontApiServer, just trimmed to the last section

then we take that name and chop the -part off

so I think this is OK

packages/visual-editor/src/utils/index.ts
jwartofsky-yext: maybe?

We already export FontRegistry

export type FontRegistry = Record<string, FontSpecification>;

packages/visual-editor/scripts/generateFontRegistry.js
jwartofsky-yext: Removed baseDetails

packages/visual-editor/src/utils/index.ts
jwartofsky-yext: Removed

packages/visual-editor/src/utils/applyTheme.ts
jwartofsky-yext: there's a bunch of helpers all over that I think we can cleanup and condense

I'll review this again tomorrow

but functionally, yes we do need to pass this in

I think we would want to update extractInUseFontFamilies to take in the whole customFonts registry (like we do for the built-in registry), then it can return inUseCustomFonts as fontRegistry type instead of string array

packages/visual-editor/src/internal/puck/components/theme-editor-sidebars/ThemeFieldsSidebar.tsx
```diff
@@ -79,23 +49,27 @@ export const ThemeFieldsSidebar = ({
...
+      return;
+    }
+
+    const customFontAssets = buildCustomFontAssets({
+      themeConfig,
+      themeValues: newThemeValues,
+      customFonts,
+    });
+    if (
+      customFontAssets.stylesheetPaths.length > 0 ||
```
asanehisa: is it possible customFontAssets.stylesheetPaths.length will be greater than 0 but customFontAssets.preloads.length will be just zero?

packages/visual-editor/src/internal/utils/customFontPreloads.ts
```diff
@@ -234,88 +81,104 @@ export const buildCustomFontPreloads = ({
...
-  return preloads;
+  return {
+    stylesheetPaths: [...stylesheetPaths],
+    preloads,
+  };
 };
 
 /**
- * Removes the custom font preloads key from theme data if present.
+ * Removes saved custom font assets from theme data before re-saving.
```
asanehisa: not certain this wording makes sense to me

packages/visual-editor/src/utils/fonts/visualEditorFonts.ts
```diff
@@ -7,36 +7,41 @@ import { fontStyleOptions } from "../themeConfigOptions.ts";
 
 const variableFontRegex = /var\((--[^)]+)\)/;
 
-export type FontRegistry = Record<string, FontSpecification>;
-type FontSpecification = {
-  italics: boolean; // whether the font supports italics
+export type WeightSupport =
+  | { weights: number[] }
```
asanehisa, line 11: can we re-add comments on which is variable and which is static?

packages/visual-editor/src/internal/utils/customFontPreloads.ts
jwartofsky-yext: basically just makes it atomic

so we don't just append to what's there. We regenerate and replace.

This function just removes customFontAssets from the provided theme

I'll change the wording

packages/visual-editor/src/internal/puck/components/theme-editor-sidebars/ThemeFieldsSidebar.tsx
jwartofsky-yext: Unlikely now that they are both passed in together in template metadata

However we can simplify this to just always generate and upsert/remove customFontAssets in the theme with what we want from templateMetadata.customFonts

packages/visual-editor/src/internal/components/InternalThemeEditor.tsx
```diff
@@ -84,17 +85,25 @@ export const InternalThemeEditor = ({
     }
 
     const currentThemeHistory = themeHistories.histories[themeHistories.index];
+    // generate customFontAssets before publishing
+    const publishThemeData = themeConfig
+      ? updateThemeWithCustomFontAssets({
```
mkilpatrick: Is this effectively a migrate function? For layout migrations we do that in Editor (vs InternalLayoutEditor). Maybe we should also do it it in ThemeEditor instead of here to be consistent?

packages/visual-editor/src/internal/components/ThemeEditor.tsx
```diff
@@ -144,59 +148,54 @@ export const ThemeEditor = (props: ThemeEditorProps) => {
...
+        : [{ id: themeSaveState.hash, data: themeSaveState.history.data }];
+      index = themeData ? 1 : 0;
+    }
+
+    if (histories) {
+      setThemeHistories({
+        histories: histories.map((themeHistory) => ({
+          ...themeHistory,
+          data: themeConfig
+            ? updateThemeWithCustomFontAssets({
```
mkilpatrick, line 187: Is a lot of this handling an edge case of the data not being migrated? If so, can't we call a top level thing that does the migration like we do for layouts?

packages/visual-editor/src/internal/puck/components/theme-editor-sidebars/ThemeFieldsSidebar.tsx
```diff
@@ -37,38 +31,10 @@ export const ThemeFieldsSidebar = ({
...
-    return () => {
-      isActive = false;
-    };
-  }, [customFonts]);
-
   const handleThemeChange = (
     themeSectionKey: string,
     themeSection: ThemeConfigSection,
-    newValue: Record<string, any>
+    newValue: Record<string, unknown>
```
mkilpatrick, line 37: Why unknown over any?

packages/visual-editor/src/internal/utils/customFontAssets.ts
```diff
@@ -0,0 +1,207 @@
...
+      mergedThemeValues[`--fontStyle-${sectionKey}-fontStyle`];
+    const style = fontStyleValue === "italic" ? "italic" : "normal";
+
+    const variant = findMatchingVariant(customFont, style, weight);
+    if (!variant || seen.has(variant.filePath)) {
+      return;
+    }
+
+    seen.add(variant.filePath);
+    preloads.push(variant.filePath);
```
mkilpatrick: Can't you just make preloads a set so you don't need `seen` as well?

packages/visual-editor/src/internal/utils/customFontAssets.ts
```diff
@@ -0,0 +1,207 @@
...
+  themeValues: ThemeData;
+  customFonts?: FontRegistry;
+}): ThemeData => {
+  if (!customFonts || Object.keys(customFonts).length === 0) {
+    if (!(CUSTOM_FONT_ASSETS_KEY in themeValues)) {
+      return themeValues;
+    }
+
+    const rest = { ...themeValues };
+    delete rest[CUSTOM_FONT_ASSETS_KEY];
```
mkilpatrick, line 145: When would this key be there if there are no customFonts?

packages/visual-editor/src/internal/utils/customFontAssets.ts
```diff
@@ -0,0 +1,207 @@
...
+    };
+  }
+
+  return {
+    stylesheetPaths: Array.isArray((assets as CustomFontAssets).stylesheetPaths)
+      ? (assets as CustomFontAssets).stylesheetPaths.filter(Boolean)
+      : [],
+    preloads: Array.isArray((assets as CustomFontAssets).preloads)
+      ? (assets as CustomFontAssets).preloads.filter(Boolean)
+      : [],
```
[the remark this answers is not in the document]
mkilpatrick: Seems legit

packages/visual-editor/src/internal/utils/customFontAssets.ts
```diff
@@ -0,0 +1,207 @@
...
+  relativePrefixToRoot: string
+): string => {
+  if (preloads.length === 0) {
+    return "";
+  }
+
+  return (
+    preloads
+      .map((href) => {
+        return `<link rel="preload" href="${normalizeAssetPath(href, relativePrefixToRoot)}" as="font" type="font/woff2" crossorigin="anonymous">`;
```
mkilpatrick: We're guaranteed that preloads only ever comes in with woff2 urls?

packages/visual-editor/src/utils/fonts/generateFontFallbacks.ts
```diff
@@ -1,6 +1,6 @@
 import { createFontStack } from "@capsizecss/core";
 import { fontFamilyToCamelCase } from "@capsizecss/metrics";
-import { defaultFonts } from "./font_registry.js";
+import { defaultFonts } from "./visualEditorFonts.ts";
 import { writeFileSync } from "node:fs";
 import { resolve } from "node:path";
```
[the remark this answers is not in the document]
mkilpatrick: Probably want to do this

packages/visual-editor/src/internal/puck/components/theme-editor-sidebars/ThemeFieldsSidebar.tsx
jwartofsky-yext: According to robots, unknown is preferred as it is typesafe

🔹 any = “turn off type checking”

When you use any, TypeScript basically gives up:

```
let value: any = "hello";
value.toUpperCase();   // ✅ allowed
value.notARealMethod(); // ✅ also allowed (no error!)
```

You can do anything—even things that will crash at runtime. This defeats the purpose of TypeScript.

🔹 unknown = “I don’t know yet—prove it first”

unknown is the type-safe alternative:

```
let value: unknown = "hello";

value.toUpperCase(); 
```
// ❌ Error: Object is of type 'unknown'

packages/visual-editor/src/internal/utils/customFontAssets.ts
jwartofsky-yext: This case would trigger if the account had custom fonts and then they were removed

So this would remove __customFontAssets from the theme so we don't add unnecessary preloads and links

mkilpatrick (dismissed):

packages/visual-editor/src/internal/components/ThemeEditor.tsx
jwartofsky-yext: yes and no

for layouts, the migrations are only one time

for this case, we recompute what  __customFontAssets should be using a combination of the theme, themeConfig, and customFonts

__customFontAssets can be different if any of those change, and aren't tied to a specific version, so I think we should compute it dynamically like we do here

alternatively we can just modify before we save/publish, but I think this is better

packages/visual-editor/src/internal/utils/customFontAssets.ts
jwartofsky-yext: Currently they are always woff2

The path we are given has the fileType extension, so we can determine that from the path

If unknown, we can omit the type and it should still work

Alternatively we can just update YSS to pass that field through explicitly in customFonts since we do have it from fontApiServer

mkilpatrick (dismissed):

packages/visual-editor/src/internal/utils/customFontAssets.ts
jwartofsky-yext: Interpreting is better

Otherwise we need to store an object like this in the theme
```
		"__customFontAssets": {
				"preloads": [
						{
								"href": "/y-fonts/jackskull-regular.woff2",
								"type": "woff2"
						},
						{
								"href": "/y-fonts/resoterose-regular.woff2",
								"type": "woff2"
						}
				],
				"stylesheetPaths": [
						"y-fonts/jackskull.css",
						"y-fonts/resoterose.css"
				]
		},
```

This seems silly because the type is just there in the url already. Also if we can't interpret the type, we can just omit it and it will still work as expected.

mkilpatrick (dismissed):

packages/visual-editor/src/utils/index.ts
```diff
@@ -43,6 +43,9 @@ export {
   defaultFonts,
   getFontWeightOptions,
   getFontWeightOverrideOptions,
+  type Font,
+  type Variant,
```
benlife5: Do these need to be package exports? If so, probably `FontVariant` and `FontWeightSupport`

benlife5 (dismissed):

asanehisa (dismissed):

asanehisa (approved):

benlife5 (approved):

mkilpatrick (approved):

briantstephan (approved):

PR #1156, opened 2026-04-14
feat: get locator working in local dev fake starter

mkilpatrick:


packages/visual-editor/src/internal/hooks/useMessage.ts
```diff
@@ -194,10 +195,14 @@ const initializeGlobalOriginTracker = () => {
  * Includes TARGET_ORIGINS plus any tracked parent origin.
  * Also tries to infer the parent origin from document.referrer if not yet tracked.
  */
-const getOriginsForSending = (targetOrigins: string[]): string[] => {
+export const getOriginsForSending = (targetOrigins: string[]): string[] => {
```
jwartofsky-yext, line 198: Do we need to export this?

starter/src/mocks/locatorLocalDev.ts
```diff
@@ -0,0 +1,302 @@
...
+const BASE_LONGITUDE = -77.069915;
+
+type LocatorVerticalResult = {
+  data: Record<string, unknown>;
+  distance: number;
+  distanceFromFilter: number;
+  highlightedFields: Record<string, unknown>;
+};
+
+const createLocatorApiResult = (index: number): LocatorVerticalResult => {
```
jwartofsky-yext, line 33: Can you add an overview comment to this file?

Usually I think of "mocks" as being for automated tests, but in this case its for the manual localDev Locator

Would it be possible for it to generate results based on the localData, or have the data stored in a separate JSON file? Might be beyond the scope of what's needed

packages/visual-editor/src/internal/hooks/useMessage.ts
mkilpatrick: It's export for the test.

starter/src/mocks/locatorLocalDev.ts
mkilpatrick: This is mocking out all of the search api calls that occur within the locator components. I'll add some comments.

packages/visual-editor/src/internal/types/templateMetadata.ts
```diff
@@ -32,8 +33,92 @@ export type FieldTypeData = {
...
+    field_type_id: "type.string",
+  },
+  services: {
+    field_id: "services",
+    field_name: "Services",
+    field_type: {
+      variant_type: 1,
+    },
+    field_type_id: "type.string",
+  },
```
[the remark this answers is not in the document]
mkilpatrick: Verified against the current code, and this one does not need a fix.

Why I did not change emails / services in [templateMetadata.ts](visual-editor/packages/visual-editor/src/internal/types/templateMetadata.ts:79):

The locator result-card selectors currently filter these fields by field_type_id === "type.string" in [LocatorResultCard.tsx](visual-editor/packages/visual-editor/src/components/LocatorResultCard.tsx:332) and specifically use getDisplayFieldOptions("type.string") for both email and services in [LocatorResultCard.tsx](visual-editor/packages/visual-editor/src/components/LocatorResultCard.tsx:626) and [LocatorResultCard.tsx](visual-editor/packages/visual-editor/src/components/LocatorResultCard.tsx:647).
The project’s existing entity-field tests already model emails as a list while still using typeRegistryId: "type.string" plus isList: true, see [getFilteredEntityFields.test.ts](visual-editor/packages/visual-editor/src/internal/utils/getFilteredEntityFields.test.ts:1489).
variant_type is not used anywhere in the source logic beyond these hardcoded metadata objects, so changing it would not make the editor “more correct” today.
If I changed emails / services to a list-specific field_type_id, they would disappear from the current locator selector UI and regress behavior.
So the current metadata is intentionally matching the current selector implementation, even though the underlying values are arrays. No code changes were made for this finding.

asanehisa (dismissed):
will conflict with work brian / i did in other branch but thats ok haha

packages/visual-editor/src/internal/hooks/useMessageReceivers.ts
```diff
@@ -15,6 +15,43 @@ import { StreamDocument } from "../../utils/types/StreamDocument.ts";
 
 const devLogger = new DevLogger();
 
+const createEmptyLocalDevLayout = (): Data => ({
```
benlife5: Doesn't need to be a function?

starter/localData/dev-locator-stream__en__8967748.json
```diff
@@ -2,21 +2,23 @@
   "__": {
     "codeTemplate": "locator",
     "entityPageSet": {},
-    "name": "locations-locator",
-    "isPrimaryLocale": true
+    "name": "dev-locator",
+    "isPrimaryLocale": true,
+    "layout": "{\"root\":{\"props\":{\"version\":73,\"title\":{\"field\":\"\",\"constantValue\":{\"defaultValue\":\"Find Locations\"},\"constantValueEnabled\":true},\"description\":{\"field\":\"\",\"constantValue\":{\"defaultValue\":\"Find the right location for you.\"},\"constantValueEnabled\":true}}},\"content\":[{\"type\":\"MainContent\",\"props\":{\"id\":\"MainContent-b6c58f1c-102d-46af-a58c-65751efa30da\",\"content\":[{\"type\":\"Locator\",\"props\":{\"pageHeading\":{\"title\":{\"defaultValue\":\"Find a Location\"}},\"id\":\"Locator-2ae506f4-a3ee-46ea-b5f9-e4c3236243a7\",\"mapStyle\":\"mapbox://styles/mapbox/streets-v12\",\"locationStyles\":[],\"filters\":{\"openNowButton\":false,\"showDistanceOptions\":false},\"resultCard\":[],\"distanceDisplay\":\"distanceFromUser\"}}]}}],\"zones\":{}}"
```
benlife5: I'd lean away from putting layout data in these, I think it is likely to be a bother during development bc of type mismatches

starter/src/ve.config.tsx
```diff
@@ -28,6 +29,10 @@ export const devConfig: Config<DevProps> = {
   root: mainConfig.root,
 };
 
-export const componentRegistry: Record<string, Config<DevProps>> = {
+export const mainComponentRegistry: Record<string, Config<DevProps>> = {
```
benlife5: Could put the two configs in the same componentRegistry with different key names that correspond to their templates, but this is fine too

jwartofsky-yext (dismissed):
lgtm!

starter/src/ve.config.tsx
mkilpatrick: The bots wanted to keep "dev" as the key due to how it was hardcoded in the templateMetadata. I've updated it to use the streamDoc.__.name. Note that I did not update this value on the localData of the directory docs. That can happen when a directory template is added.

benlife5 (approved):

starter/src/ve.config.tsx
benlife5: Ah right, forgot template metadata was hardcoded. This is better though, thanks

PR #1157, opened 2026-04-14
chore: export breadcrumb and nearby location utils

benlife5:
To handle breadcrumbs and nearby locations in the codex generated components, we'll need access to these utils. 

Refactored NearbyLocations to expose a single hook that handles all the details. Cleaned up the types a bit.


packages/visual-editor/src/components/pageSections/NearbyLocations/utils.ts
jwartofsky-yext: not you, but I think the core function of NearbyLocations should get its own file and not just be in "utils"

packages/visual-editor/src/components/pageSections/index.ts
```diff
@@ -44,6 +44,12 @@ export {
   NearbyLocationsSection,
   type NearbyLocationsSectionProps,
 } from "./NearbyLocations/NearbyLocations.tsx";
+export {
+  parseDocumentForNearbyLocations,
```
jwartofsky-yext: this is kind of a weirdly named function to have publicly exported

packages/visual-editor/src/components/pageSections/NearbyLocations/utils.ts
```diff
@@ -10,27 +11,49 @@ type Coordinate = {
...
+    /** The entity data for the nearby locations. */
+    docs: NearbyLocationDoc[];
+    /** The number of nearby locations returned. */
+    count: number;
+  };
+};
+
 /** parseDocument parses the streamDocument to get the businessId, apiKey, contentEndpointId, and contentDeliveryAPIDomain */
-export const parseDocument = (
+export const parseDocumentForNearbyLocations = (
```
jwartofsky-yext: I wonder if this should become a more generic parseDocumentVars or something like that

Most of the fields returned are fairly generic. The apiKey comes the YEXT_PUBLIC_VISUAL_EDITOR_APP which also gives access to other things

As a more generic function, this could also be used for Locators and possibly other things if we get to that point

jwartofsky-yext (commented):
lgtm, just minor nitpicks

Once we export something we can't really "unexport" it in future versions without some headache

packages/visual-editor/src/components/pageSections/NearbyLocations/utils.ts
benlife5: hmm I'm not sure a generic function would be very useful since most things you can access directly off the document. This function is mostly just to gather the specific properties you need for nearby locations.

I guess it probably could be refactored such that the useQuery takes the document directly and then you wouldn't need to use both parseDocumentForNearbyLocations and fetchNearbyLocations, I'll take a look at that

jwartofsky-yext: Sounds good!

I'm just wary of exporting hyper-specific functions that don't have a clear use case to outside users

packages/visual-editor/src/docs/ai/components.d.ts
```diff
@@ -24,7 +24,7 @@ import {
   HoursType,
   AddressType,
   DayOfWeekNames,
-  Coordinate,
+  Coordinate as Coordinate$1,
```
mkilpatrick: This one is weird

benlife5: This file was started for PaigeBuilder and I've been meaning to clean it up. I've gone ahead and removed it.

mkilpatrick (approved):

asanehisa (approved):

jwartofsky-yext (approved):
LGTM!

PR #1158, opened 2026-04-14
refactor: merge dynamicSingleSelect into basicSelector

mkilpatrick:
`dynamicSingleSelect`'s only real differences from `basicSelector` were:
* handles dynamic options in the way of a function call
* used Puck's `select` field under the hood (while `basicSelector` uses Combobox)

This merges the dynamic functionality of `dynamicSingleSelect` into `basicSelector`. It also doesn't make sense for us to have different types of dropdown fields. They should all look and function the same way.


packages/visual-editor/src/components/Locator.test.tsx
```diff
@@ -533,6 +533,116 @@ const tests: ComponentTest[] = [
     },
     version: migrationRegistry.length,
   },
+  {
+    name: "version 72 wrapped result card selector values",
```
asanehisa: placement of this test is odd? it's above the version 24 test

packages/visual-editor/src/components/LocatorResultCard.tsx
```diff
@@ -1464,16 +1413,19 @@ const parseHoursFromLocation = (
   return resolveProjectedField(location, hoursFieldId) as HoursType;
 };
 
+type LegacyLocatorResultCardFieldValue = {
```
benlife5: If the migration works then you don't need this fallback I think

benlife5 (dismissed):

jwartofsky-yext (dismissed):
lgtm!

packages/visual-editor/src/components/Locator.test.tsx
```diff
@@ -1138,6 +1138,136 @@ const tests: ComponentTest[] = [
   },
 ];
 
+const wrappedSelectorValuesMigrationFixture = {
```
asanehisa, line 1141: looks like u removed it from normal test table we have? if so we don't need the screenshots added anymore?

benlife5 (approved):

asanehisa (approved):

PR #1161, opened 2026-04-16
feat: add migrations for themeData

jwartofsky-yext:
*Not required for the VE customFont changes

Adds a migration to migrate theme data

Adds versioning to themes

The first migration simply removes __customFontPreloads as these have been replaced with __customFontAssets which are dynamically generated on theme load and change.

Refactors the code for loading themes to match the pattern used for Layouts

migrate is renamed to migrateLayout. migrate still works, but is marked deprecated


packages/visual-editor/src/utils/migrateTheme.ts
```diff
@@ -0,0 +1,43 @@
...
+    );
+
+  if (migrationRegistry.length === 0) {
+    return migratedThemeValues;
+  }
+
+  return {
+    ...migratedThemeValues,
+    [THEME_VERSION_KEY]: migrationRegistry.length,
+  };
```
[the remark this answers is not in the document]
jwartofsky-yext: This should be fixed now

packages/visual-editor/src/internal/hooks/theme/useMessageReceivers.ts
```diff
@@ -26,7 +33,9 @@ export const useThemeMessageReceivers = (localDev: boolean) => {
     if (payload?.history) {
       receivedThemeSaveState = {
         hash: payload.hash,
-        history: { data: JSON.parse(payload.history) },
+        history: {
+          data: migrateTheme(JSON.parse(payload.history), migrationContext),
+        },
```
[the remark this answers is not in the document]
jwartofsky-yext: this should be fixed

packages/visual-editor/src/utils/index.ts
```diff
@@ -10,11 +10,18 @@ export { themeManagerCn, themeManagerTwMergeConfiguration } from "./cn.ts";
...
 export {
+  migrateLayout,
   migrate,
   type Migration,
   type MigrationAction,
   type MigrationRegistry,
 } from "./migrate.ts";
+export {
+  migrateTheme,
+  THEME_VERSION_KEY,
```
benlife5: Does this need to be a package export?

packages/visual-editor/src/utils/index.ts
```diff
@@ -10,11 +10,18 @@ export { themeManagerCn, themeManagerTwMergeConfiguration } from "./cn.ts";
 export { getPageMetadata } from "./getPageMetadata.ts";
 export { normalizeLink } from "./normalizeLink.ts";
 export {
+  migrateLayout,
   migrate,
   type Migration,
```
benlife5: Might as well rename these types too

packages/visual-editor/src/internal/themeMigrations/migrationRegistry.ts
```diff
@@ -0,0 +1,6 @@
+import type { ThemeMigrationRegistry } from "../../utils/migrateTheme.ts";
+import { removeCustomFontPreloadsMigration } from "./0001_remove_custom_font_preloads.ts";
+
+export const migrationRegistry: ThemeMigrationRegistry = [
```
benlife5: probably want this to be called themeMigrationRegistry

packages/visual-editor/src/utils/migrateLayout.ts
benlife5, line 1: rename this file too

packages/visual-editor/src/utils/migrate.ts
```diff
@@ -153,3 +153,8 @@ export const migrate = (
   data.root.props.version = migrationRegistry.length;
   return data;
 };
+
+/**
+ * @deprecated Use migrateLayout instead.
```
benlife5: I know it is technically correct to not make breaking changes to the package exports, but I feel like keeping our code clean is more important right now since the package is only used internally

jwartofsky-yext: The templates use the migrations

Ik the templates also live in this repo for platform pagesets, but it's possible they could have a in-repo template using migrations

packages/visual-editor/src/components/migrations/0018_add_directory_site_name.ts
```diff
@@ -1,6 +1,6 @@
-import { Migration } from "../../utils/migrate.ts";
+import { LayoutMigration } from "../../utils/migrate.ts";
 
-export const addDirectorySiteNameMigration: Migration = {
+export const addDirectorySiteNameMigration: LayoutMigration = {
```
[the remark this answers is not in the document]
jwartofsky-yext: This just changes the imports because we renamed "migration" to "layoutMigration" because we've also added themeMigrations

This doesn't change the actual functionality of the migrations

briantstephan (dismissed):

briantstephan (dismissed):

asanehisa (commented):
> migrate is renamed to migrateLayout. migrate still works, but is marked deprecated

This is false now right? 
Also just to check, the backfill in VES didn't do this work as well during that one time backfill?

jwartofsky-yext:
> > migrate is renamed to migrateLayout. migrate still works, but is marked deprecated
> 
> This is false now right? Also just to check, the backfill in VES didn't do this work as well during that one time backfill?

Migrate is still exposed. It's just marked as deprecated. That way any old sites that are using the exported migrate won't break.

The backfill populated __customFontAssets, but did not cleanup __customFontPreloads

The migrate should cleanup __customFontPreloads because we only want to remove those when upgrading. We also generate the __customFontAssets here, but those were also added by the backfill.

packages/visual-editor/src/utils/migrate.ts
asanehisa: Who is they? If it's rbc they do not have migrations in their in-repo code and they are the only hybrid site we have

jwartofsky-yext:
not needed

PR #1167, opened 2026-04-20
feat: handle single linked entity field

jwartofsky-yext:
Also handles embedded fields

Screenshot 2026-04-21 at 4 12 58 PM

If there are multiple linked entities, we use the first one and show a warning toast:

image

https://jam.dev/c/2f23c855-1bf6-4186-b02d-ec98358d9e8a


packages/visual-editor/src/utils/linkedEntityFieldUtils.ts
asanehisa, line 1: should we move this into utils/ ?

packages/visual-editor/src/utils/linkedEntityFieldUtils.ts
```diff
@@ -0,0 +1,103 @@
...
+  YextSchemaField,
+  type YextFieldDefinition,
+} from "../types/entityFields.ts";
+
+export type LinkedEntitySchema = {
+  displayName: string;
+  fields: YextSchemaField[];
+};
+
+export type LinkedEntitySchemas = Record<string, LinkedEntitySchema>;
```
benlife5, line 16: ts doc, particularly for what the key represents in this type

packages/visual-editor/src/utils/resolveYextEntityField.ts
```diff
@@ -189,27 +194,44 @@ export const findField = <T>(
   document: any,
   fieldName: string
 ): T | undefined => {
+  return resolveField<T>(document, fieldName).value;
+};
+
+export const resolveField = <T>(
+  document: any,
```
asanehisa: should this be streamDocument: StreamDocument?

packages/visual-editor/src/editor/linkedEntityFieldUtils.ts
```diff
@@ -0,0 +1,103 @@
...
+  const [referenceFieldName] = fieldPath.split(".");
+  return Object.hasOwn(linkedEntitySchemas, referenceFieldName);
+};
+
+/**
+ * Creates a minimal synthetic schema field for a linked entity reference root.
+ * The root intentionally does not advertise a scalar type so only its children
+ * are presented for normal single-value field selection.
+ */
+export const createLinkedEntityRootDefinition = (
```
benlife5: Doesn't need to be exported?

asanehisa (commented):
where in the editor is that toast?

packages/visual-editor/src/editor/YextEntityFieldSelector.tsx
```diff
@@ -508,6 +526,48 @@ export const EntityFieldInput = <T extends Record<string, any>>({
     templateMetadata.entityTypeDisplayName,
     typeSelectorConfig,
     value?.selectedType,
+    linkedEntitySchemas,
+  ]);
+
+  React.useEffect(() => {
```
benlife5, line 466: Add a brief comment re: the goal of this useEffect

packages/visual-editor/src/components/testing/screenshots/Locator/[tablet] latest version default props.png
Comment (benlife5) [inline packages/visual-editor/src/components/testing/screenshots/Locator/[tablet] latest version default props.png:1]: I'd add a screenshot test that references a linked entity field (can be whatever component makes sense)

Comment (benlife5) [inline packages/visual-editor/src/components/testing/screenshots/Locator/[tablet] latest version default props.png:1]: Also, can you add some linked entity data to the `localData` and `dev.config.ts`?

benlife5 (commented):
Nice! Just left a few housekeeping comments

packages/visual-editor/src/editor/YextEntityFieldSelector.tsx
```diff
@@ -247,8 +278,6 @@ export const ConstantValueModeToggler = ({
   label: string;
   showLocale?: boolean;
 }) => {
-  // If disableConstantValue is true, constantValueInputSupported is always false.
```
asanehisa, line 250: any reason to remove comments?

packages/visual-editor/src/editor/YextEntityFieldSelector.tsx
```diff
@@ -66,12 +77,31 @@ const isYextPuckFieldType = (
 
 export type { YextEntityField } from "./yextEntityFieldUtils.ts";
 
+/**
+ * Configuration for the type selector dropdown in the YextEntityFieldSelector.
+ * This allows for a two-tiered selection: first a "type", then a "field" of that type.
+ */
+export type TypeSelectorConfigProps = {
```
asanehisa: do u use this / need it?

briantstephan (dismissed):

packages/visual-editor/src/editor/YextEntityFieldSelector.tsx
jwartofsky-yext: was an artifact of a merge commit

fixed

packages/visual-editor/src/editor/YextEntityFieldSelector.tsx
jwartofsky-yext: it got re-added by a merge commit

cleaned up

briantstephan (dismissed):

packages/visual-editor/src/editor/YextEntityFieldSelector.tsx
```diff
@@ -50,8 +51,17 @@ import {
   getFieldsForSelector,
   type YextEntityField,
 } from "./yextEntityFieldUtils.ts";
+import { useDocument } from "../hooks/useDocument.tsx";
+import { resolveField } from "../utils/resolveYextEntityField.ts";
+import { toast } from "sonner";
+import { isLinkedEntityFieldPath } from "../utils/linkedEntityFieldUtils.ts";
 
 const devLogger = new DevLogger();
+let warnedLinkedEntityFieldPathsByDocument = new WeakMap<any, Set<string>>();
```
asanehisa: any reason for this to be a WeakMap? also would it not be <string, StreamDocument> ?

packages/visual-editor/src/hooks/useLinkedEntitySchemas.tsx
```diff
@@ -0,0 +1,60 @@
...
+  if (context === undefined) {
+    throw new Error(
+      "useLinkedEntitySchemas must be used within LinkedEntitySchemasContext.Provider"
+    );
+  }
+
+  return context;
+};
+
+export { useLinkedEntitySchemas, LinkedEntitySchemasContext };
```
asanehisa: we tend to avoid doing exports like this and inline them instead

packages/visual-editor/src/editor/YextEntityFieldSelector.tsx
jwartofsky-yext: so this is actually <StreamDocument, Set<string>>

The WeakMap allows the key (steamDocument) to be garbage collected.

It's a map of the document to the linked entity fields so we don't show the same warning multiple times.

I think we should change this to not store a set and only store the last field that was warned for. Will make that change.

packages/visual-editor/src/hooks/useLinkedEntitySchemas.tsx
jwartofsky-yext: Fixed and updated my robots

packages/visual-editor/src/editor/YextEntityFieldSelector.tsx
jwartofsky-yext: Updated to use a ref instead

packages/visual-editor/src/hooks/useLinkedEntitySchemas.tsx
```diff
@@ -0,0 +1,58 @@
...
+ */
+export const usePlatformBridgeLinkedEntitySchemas = () => {
+  const [linkedEntitySchemas, setLinkedEntitySchemas] =
+    React.useState<LinkedEntitySchemas | null>(null);
+
+  useReceiveMessage(
+    "getLinkedEntitySchemas",
+    TARGET_ORIGINS,
+    (send, payload: LinkedEntitySchemasPayload) => {
+      const receivedSchemas =
```
mkilpatrick, line 28: useEntityFields converts the raw payload into a nicer object in `assignDefinitions`. I feel like we want to do something similar.

Actually I see you are but when the selector is called. Should we do it once here instead?

packages/visual-editor/src/utils/linkedEntityFieldUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+ * entity that reference points to.
+ */
+export type LinkedEntitySchemas = Record<string, LinkedEntitySchema>;
+
+/**
+ * Builds a synthetic {@link StreamFields} object from linked entity schemas so
+ * the existing field filtering pipeline can treat linked entity fields the same
+ * way as primary entity schema fields.
+ */
+export const buildLinkedEntityStreamFields = (
```
mkilpatrick, line 23: Probably want tests for this file

packages/visual-editor/src/hooks/useLinkedEntitySchemas.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Receives linked entity schemas from the platform bridge and keeps them in a
+ * dedicated store separate from the base entity schema.
+ */
+export const usePlatformBridgeLinkedEntitySchemas = () => {
+  const [linkedEntitySchemas, setLinkedEntitySchemas] =
+    React.useState<LinkedEntitySchemas | null>(null);
+
+  useReceiveMessage(
```
mkilpatrick, line 24: Make sure if you have xYextDebug we can see the schema. I don't think we have it for the normal yext fields so we should add it to that too. It will help with debugging.

jwartofsky-yext: There is a slight race condition here and with that since xYextDebug comes through the templateMetadata so if we get this message before xYextDebug the log won't appear

I'll investigate solutions

jwartofsky-yext: Added a buffer to the devLogger

Now it will log properly if we get a message linked entity fields or linked entity fields before receiving templateMetadata

packages/visual-editor/src/hooks/useLinkedEntitySchemas.tsx
jwartofsky-yext: useEntityFields normalizes the shape from getTemplateStream and getDevEntity fields

For these linked fields there's only one shape received

The received linkedEntitySchemas shape is better for linked-path logic, while the derived schema is just used towards the end 

There is potential to refactor everything into fewer shapes, but I'm not sure if that's an improvement. Robots don't think it is. I will try in another CR to see how that looks

packages/visual-editor/src/hooks/useEntityFields.tsx
```diff
@@ -32,7 +37,9 @@ export const usePlatformBridgeEntityFields = () => {
   });
 
   useReceiveMessage("getDevEntityFields", TARGET_ORIGINS, (send, payload) => {
-    setEntityFields({ fields: assignDefinitions(payload) });
+    const receivedValues = { fields: assignDefinitions(payload) };
```
mkilpatrick, line 40: I forget what the difference is between getTemplateStream and getDevEntityFields. But any reason why we call assignDefinitions here but not above?

jwartofsky-yext: assignDefinitions converts the payload from getDevEntityFields to match what we get for entityFields

entityFields already comes in the correct format

packages/visual-editor/src/editor/YextEntityFieldSelector.tsx
```diff
@@ -410,12 +414,22 @@ export const EntityFieldInput = <T extends Record<string, any>>({
   label,
 }: InputProps<T>) => {
   const entityFields = useEntityFields();
+  const linkedEntitySchemas = useLinkedEntitySchemas();
   const templateMetadata = useTemplateMetadata();
-
+  const streamDocument = useDocument();
+  const lastWarnedLinkedEntityFieldRef = React.useRef<
+    | {
+        document: object;
```
asanehisa: nit use slightly diff variable name, document messes us up b/c of it existing as a window document.

mkilpatrick (approved):

asanehisa (approved):

PR #1169, opened 2026-04-21
fix: types for local starter

jwartofsky-yext:
Running the local visual-editor/starter would result in type errors

This change fixes them and allows the visual-editor to start

```
│ src/editor/DynamicOptionsSelector.test.tsx(52,15): error TS2322: Type '(newValue: any, uiState: any) => void' is not assignable to type '(value: DynamicOptionsSelectorType<string> | undefined) => void'.
│   Target signature provides too few arguments. Expected 2 or more, but got 1.
│ src/editor/DynamicOptionsSelector.test.tsx(52,26): error TS7006: Parameter 'newValue' implicitly has an 'any' type.
│ src/editor/DynamicOptionsSelector.test.tsx(52,36): error TS7006: Parameter 'uiState' implicitly has an 'any' type.
│ src/editor/DynamicOptionsSelector.test.tsx(54,42): error TS2554: Expected 1 arguments, but got 2.
```

Confirmed fix with node 20 and 22


briantstephan (approved):

benlife5 (approved):

packages/visual-editor/src/editor/DynamicOptionsSelector.test.tsx
```diff
@@ -49,9 +49,11 @@ const renderDynamicOptionsField = (
           render: (props: Parameters<typeof field.render>[0]) =>
             field.render({
               ...props,
-              onChange: (newValue, uiState) => {
+              onChange: (
+                newValue: Parameters<typeof props.onChange>[0]
```
asanehisa, line 53: i think ideally uiState would exist in this test as well as it should work for things like array fields to work properly.

jwartofsky-yext: I had old node modules,

pnpm I fixed the issue so we don't need this

jwartofsky-yext:
Not needed

Run pnpm i, ensure that node modules is up to date, particularly the puck version

Lockfiles and package.json are already set correctly

PR #1173, opened 2026-04-21
fix: localStorage history state not loading

mkilpatrick:
If an older localStorage entry had state.data but no state.ui, that turned into ui: undefined, and Puck then crashed while reading state.ui.itemSelector.


asanehisa (approved):
Do we need this or does clearing localStorage fix it?

mkilpatrick:
> Do we need this or does clearing localStorage fix it?

Clearing would fix it, but now you don't need to!

jwartofsky-yext (approved):

PR #1177, opened 2026-04-23
refactor: make static map component more flexible 

benlife5:
- Exports the react component and list of mapbox styles
- Increases layout flexibility so the map can adapt to the container created by the codex templates 
- preserves the current rendering (see below)

Screenshot 2026-04-23 at 3 38 01 PM

### DIFF
--- a/packages/visual-editor/src/components/contentBlocks/MapboxStaticMap.tsx
+++ b/packages/visual-editor/src/components/contentBlocks/MapboxStaticMap.tsx
@@ -193,7 +193,6 @@ export const MapboxStaticMap: ComponentConfig<{ props: MapboxStaticProps }> = {
         longitude: 0,
       },
     },
-    height: "300px",
     mapStyle: "streets-v12",
   },
   render: (props) => <MapboxStaticMapComponent {...props} />,
--- a/packages/visual-editor/src/components/contentBlocks/MapboxStaticMap.tsx
+++ b/packages/visual-editor/src/components/contentBlocks/MapboxStaticMap.tsx
@@ -124,7 +124,7 @@ export const MapboxStaticMapComponent: PuckComponent<MapboxStaticProps> = ({
   const staticImageSizes = {
     large: "1280x720",
     medium: "960x540",
-    small: "412x232",
+    small: "412x412",
   } as const;
 
   type StaticImageSize = keyof typeof staticImageSizes;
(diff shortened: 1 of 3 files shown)


packages/visual-editor/src/components/contentBlocks/MapboxStaticMap.tsx
```diff
@@ -196,6 +193,7 @@ export const MapboxStaticMap: ComponentConfig<{ props: MapboxStaticProps }> = {
         longitude: 0,
       },
     },
+    height: "300px",
```
jwartofsky-yext: Should this be the height prop?

packages/visual-editor/src/components/contentBlocks/MapboxStaticMap.tsx
```diff
@@ -75,6 +68,7 @@ const getPrimaryColor = (streamDocument: StreamDocument) => {
 export const MapboxStaticMapComponent: PuckComponent<MapboxStaticProps> = ({
   apiKey,
   coordinate: coordinateField,
+  height = "300px",
```
jwartofsky-yext, line 71: Do we need a migration for this?

packages/visual-editor/src/components/contentBlocks/MapboxStaticMap.tsx
benlife5: Ah yeah this can be removed

packages/visual-editor/src/components/contentBlocks/MapboxStaticMap.tsx
benlife5: The current value is 300px and this is a react-only prop not a puck field so I think we're fine

packages/visual-editor/src/components/contentBlocks/MapboxStaticMap.tsx
```diff
@@ -143,39 +138,41 @@ export const MapboxStaticMapComponent: PuckComponent<MapboxStaticProps> = ({
       displayName={pt("coordinate", "Coordinate")}
       fieldId={coordinateField.field}
       constantValueEnabled={coordinateField.constantValueEnabled}
-      className="w-full"
+      className="h-full w-full"
+      fullHeight
```
asanehisa, line 142: do we need this with h-full?

asanehisa (dismissed):

packages/visual-editor/src/components/contentBlocks/MapboxStaticMap.tsx
benlife5: Nope

asanehisa (approved):

jwartofsky-yext (approved):

PR #1183, opened 2026-04-28
feat: allow Other color option in basicSelector

asanehisa:
https://github.com/user-attachments/assets/a29d3c9b-bf14-401e-9f23-1505f1a42ee0


packages/visual-editor/src/fields/BasicSelectorField.tsx
```diff
@@ -147,17 +197,40 @@ export const BasicSelectorFieldOverride = ({
...
       : [{ options }]);
   const noOptionsMessage = providedNoOptionsMessage ?? undefined;
+  const isThemeColorSelector =
+    field.optionGroups === undefined &&
+    (field.options === "SITE_COLOR" || field.options === "BACKGROUND_COLOR");
+  const customColorHex =
+    getThemeColorHexValue(
+      isThemeColorValue(value) ? value.selectedColor : undefined,
+      templateProps?.document
+    ) ?? "#000000";
```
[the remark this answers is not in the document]
asanehisa: existing data won't provide string color tokens

benlife5 (dismissed):
Nice!

packages/visual-editor/src/components/MapPinIcon.tsx
```diff
@@ -9,26 +10,6 @@ const resolveCssVarColor = (
   return window.getComputedStyle(element).getPropertyValue(varName).trim();
 };
 
-const resolvePinFillColor = (
```
mkilpatrick, line 12: This is ok to remove?

packages/visual-editor/src/fields/ColorSelector.tsx
```diff
@@ -6,11 +6,35 @@ import { pt } from "../../../utils/i18n/platform.ts";
...
+      <ColorPickerInput
+        ariaLabel={field.label || pt("colorPicker.open", "Open color picker")}
+        value={value}
+        onChange={onChange}
+      />
+    </FieldLabel>
+  );
+};
+
+export const ColorPickerInput = ({
```
mkilpatrick, line 24: Should we also make this a fieldType?

asanehisa: Oh rather than part of the basicSelector?

mkilpatrick: Well, it could be used standalone but basicSelector would use that color fieldType. Idk if that's necessary though as it might never be used. 

Regardless, I feel like we should move this file. All of the fieldType rendering lives in src/fields now but this lives elsewhere which seems strange.

packages/visual-editor/src/components/MapPinIcon.tsx
asanehisa: Yup, confirmed with dev mode testing that icons colors still resolve correctly. 
Screenshot 2026-04-29 at 3 45 29 PM
Screenshot 2026-04-29 at 3 45 57 PM

packages/visual-editor/src/fields/ColorSelector.tsx
asanehisa: Moved file!

packages/visual-editor/src/fields/ColorSelector.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { pt } from "../utils/i18n/platform.ts";
+
+type RenderProps = Parameters<CustomField<any>["render"]>[0];
+
+export const ColorSelector = ({ field, value, onChange }: RenderProps) => {
+  const translatedLabel = field.label && pt(field.label);
+
+  return (
+    <FieldLabel
+      label={translatedLabel || "Label is undefined"}
```
benlife5: Should this have a different or no fallback?

asanehisa: Yeah good point, I think label is required but I'll adjust it.

packages/visual-editor/src/fields/ColorSelector.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { CustomField, FieldLabel } from "@puckeditor/core";
+import { Color, ColorResult, SketchPicker } from "react-color";
+import { pt } from "../utils/i18n/platform.ts";
+
+type RenderProps = Parameters<CustomField<any>["render"]>[0];
+
+export const ColorSelector = ({ field, value, onChange }: RenderProps) => {
+  return (
+    <FieldLabel
+      label={field.label ?? "Color Picker"}
```
benlife5, line 11: Might need `pt`?

asanehisa: Oh yeah I can add if we want? This shouldn't ever pop up as ColorSelector has field.label everywhere it is used, so its just a fallback

benlife5: I see, `field.label` is already translated in `packages/visual-editor/src/internal/utils/constructThemePuckFields.ts` by the time it gets here 👍

benlife5 (approved):

briantstephan (approved):
🔥

mkilpatrick (approved):

PR #1193, opened 2026-04-30
fix: footer logo migration

benlife5:
Prior to #1034, the Footer Logo's prop type was an image YextEntityFieldSelector but the field used was AssetImage (no entity field). #1034 fixed this, but should've included a migration to fix any mismatched data. 

Unfortunately, the resolver happens to pick the old data path if both exist ([thread](https://yext.slack.com/archives/C02UVSE7P6W/p1777559499279649))

This migration:
- moves data from the old path to the new path, if nothing in the new path exists
- removes data from the old path, if data exists in the new path

### DIFF
--- a/packages/visual-editor/src/components/migrations/0075_normalize_footer_logo_image.ts
+++ b/packages/visual-editor/src/components/migrations/0075_normalize_footer_logo_image.ts
@@ -15,11 +15,13 @@ const normalizeFooterLogoImage = (image: unknown) => {
 
   const { field, constantValue, constantValueEnabled, ...localizedImage } =
     image;
+  const hasActiveEntityField =
+    typeof field === "string" && field !== "" && constantValueEnabled === false;
 
   return {
-    field: typeof field === "string" ? field : "",
+    field: hasActiveEntityField ? field : "",
     constantValue: constantValue ?? localizedImage,
-    constantValueEnabled: constantValueEnabled ?? true,
+    constantValueEnabled: hasActiveEntityField ? false : true,
   };
 };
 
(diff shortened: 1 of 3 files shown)


packages/visual-editor/src/components/migrations/0075_normalize_footer_logo_image.ts
```diff
@@ -0,0 +1,91 @@
...
+      return {
+        ...props,
+        slots: {
+          ...props.slots,
+          LogoSlot: normalizedLogoSlot,
+        },
+      };
+    },
+  },
+  FooterLogoSlot: {
```
asanehisa, line 50: do u need to migrate FooterLogoSlot and ExpandedFooter?

jwartofsky-yext (dismissed):

packages/visual-editor/src/components/migrations/0075_normalize_footer_logo_image.ts
benlife5: migrate finds the components through Puck's walkTree, which recurses through all the slots, so this will update the logo inside of any ExpandedFooters

asanehisa: Yup! Don't you only need FooterLogoSlot then?

benlife5: 🤦‍♂️ oof my apologies. Yep, removed!

asanehisa: no worries!

asanehisa (approved):

mkilpatrick (approved):

PR #1196, opened 2026-04-30
refactor: add entityField type

briantstephan:
This converts the current `YextField(..., { type: "entityField", ... })` special-case into a first-class registered field type under `src/fields`, and updates all callers to use `entityField` directly. This will allow us to remove `YextField` entirely.

Tested on this site and confirmed that entity fields still seem to work like before.
Feel free to play around with it.
https://dev.yext.com/s/1000167375/yextsites/67128/pagesets


packages/visual-editor/src/editor/YextField.tsx
```diff
@@ -74,23 +69,17 @@ export type YextFieldDefinition<ValueType = any> =
       ? YextObjectField<ValueType>
```
mkilpatrick, line 69: YextField should be deleted now as it has no meaningful usages anymore. You _may_ need to keep around some of the types in here and move them to fields.ts. Make sure they're needed though.

packages/visual-editor/src/internal/puck/components/meta-title/MetaTitleField.tsx
```diff
@@ -25,19 +26,23 @@ export const MetaTitleField = (): Field<
...
       const metaTitleField = React.useMemo(
-        () =>
-          YextField(msg("fields.metaTitle", "Meta Title"), {
-            type: "entityField",
-            filter: {
-              types: ["type.string"],
-            },
-          }),
+        () => ({
+          type: "entityField" as const,
```
mkilpatrick: Do you need the `as`s here?

packages/visual-editor/src/fields/CTASelectorField.tsx
benlife5, line 1: Why is this file changed?

packages/visual-editor/src/internal/puck/components/meta-title/MetaTitleField.tsx
briantstephan: It supposedly tells TS that the value is specifically `entityField` rather than any string.

packages/visual-editor/src/editor/YextField.tsx
briantstephan: I think that would be better for a follow-up PR since there are still a bunch of usages throughout the code that would need to be updated.

packages/visual-editor/src/fields/CTASelectorField.tsx
briantstephan: It has the same constant value mode toggle flow as EntityField, and there was a bot comment that pointed out that the toggle could appear even when no constant value editor can be resolved, but I'm thinking that case shouldn't ever actually occur for CTASelectorField. I've reverted the changes in this file.

packages/visual-editor/src/internal/puck/components/meta-title/MetaTitleField.tsx
mkilpatrick: I think you can type the useMemo and then it should be fine

packages/visual-editor/src/fields/EntityFieldSelectorField.tsx
```diff
@@ -0,0 +1,596 @@
...
+  }, [
+    filter.includeListsOnly,
+    linkedEntitySchemas,
+    streamDocument,
+    value?.field,
+  ]);
+
+  return (
+    <div className={"ve-inline-block ve-w-full " + className}>
+      <YextAutoField
```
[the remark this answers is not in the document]
mkilpatrick: Looks legit

packages/visual-editor/src/editor/YextField.tsx
mkilpatrick: Ok that can be a followup then. It's literally just a passthrough now. Can you do that next?

briantstephan: Yep will do

packages/visual-editor/src/internal/puck/components/meta-title/MetaTitleField.tsx
briantstephan: Good call, updated

mkilpatrick (dismissed):

packages/visual-editor/src/fields/EntityFieldSelectorField.tsx
```diff
@@ -0,0 +1,596 @@
...
+              ...value,
+              constantValueEnabled,
+            })
+          }
+          disableConstantValue={field.disableConstantValueToggle}
+          label={translatedLabel}
+          showLocale={supportsLocalizedConstantValue(constantFieldConfig)}
+        />
+      ) : (
+        <p
```
asanehisa: I think we can use FieldLabel from Puck here instead of this?

packages/visual-editor/src/fields/EntityFieldSelectorField.tsx
```diff
@@ -0,0 +1,596 @@
...
+            uiState
+          );
+        }}
+        value={value?.field}
+      />
+    </div>
+  );
+};
+
+export type { YextEntityField } from "../editor/yextEntityFieldUtils.ts";
```
asanehisa: this seems odd?

briantstephan: Removed

packages/visual-editor/src/fields/EntityFieldSelectorField.tsx
briantstephan: Done

packages/visual-editor/src/fields/EntityFieldSelectorField.tsx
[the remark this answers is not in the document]
briantstephan: Updated

asanehisa (dismissed):

packages/visual-editor/src/editor/README.md
```diff
@@ -268,13 +272,12 @@ const myComponentFields: YextFields<MyComponentProps> = {
 
 ## YextField
```
jwartofsky-yext, line 273: Should we mark this as deprecated?

jwartofsky-yext (commented):
generally lgtm

Can you add some more detail to the description? A summary of what we are migrating over and why

packages/visual-editor/src/editor/README.md
briantstephan: I'm going to be removing it entirely in a follow-up PR

mkilpatrick (approved):

asanehisa (approved):

PR #1197, opened 2026-04-30
feat: add 'styledText' fieldType

asanehisa:
Unused in current code, but will be used in mission's code. 

Tested by temporarily changing HeadingText to use this type and confirmed options were as expected. Default was an option and weight / styling was dependent on font family. We might change Default to "Use Global Styles" or something else. 

Only look at the values beneath `Test Text Styles`

https://github.com/user-attachments/assets/b6353ac2-8a0f-435e-b7fe-25550baf9ff0


packages/visual-editor/src/fields/StyledTextField.tsx
```diff
@@ -0,0 +1,255 @@
...
+  defaultFonts,
+  getFontStyleOptionsForFontFamily,
+  getFontWeightOptionsForFontFamily,
+  type FontRegistry,
+} from "../utils/fonts/visualEditorFonts.ts";
+import { ThemeOptions } from "../utils/themeConfigOptions.ts";
+import { pt, type MsgString } from "../utils/i18n/platform.ts";
+import type { StyleSelectOption } from "../utils/themeResolver.ts";
+
+export type StyledTextValue = {
```
benlife5, line 19: Can these subfields be required?

packages/visual-editor/src/fields/StyledTextField.tsx
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  return correctedValue;
+};
+
+export const StyledTextFieldOverride = ({
+  field,
+  value,
+  onChange,
+}: StyledTextFieldProps) => {
```
benlife5, line 103: Are all the useMemo/useCallbacks necessary? I know sometimes field rendering needs stable objects but also a lot of these calculations are pretty basic

packages/visual-editor/src/fields/StyledTextField.tsx
```diff
@@ -0,0 +1,255 @@
...
+        [key]: nextValue,
+      })
+    );
+  };
+
+  return (
+    <div className="ve-flex ve-flex-col ve-gap-3">
+      {field.label && (
+        <div className="ve-text-sm ve-font-medium">{pt(field.label)}</div>
+      )}
```
benlife5, line 190: Don't know how easy it would be to replicate, but I think ideally these subfields would be in a box like objectFields to denote that they are subproperties of the main label

asanehisa: Oooh good idea!

packages/visual-editor/src/fields/StyledTextField.tsx
asanehisa: I think one or two might be needed but let me check which ones cuz ur right i don't think we need them all

asanehisa: Oops we didn't need any lol, thanks for callout!

packages/visual-editor/src/fields/StyledTextField.tsx
asanehisa: Done!

asanehisa: Screenshot 2026-05-01 at 11 31 37 AM

benlife5 (dismissed):

mkilpatrick (approved):

benlife5 (approved):

PR #1198, opened 2026-05-01
feat: add styledButton option for fields

asanehisa:
Refactored some of the code used in styledText as it is reused in styledButton. StyledButton is not used in any components in this branch but tested by temporarily hooking it up to the CTAWrapper component. 

Got rid of the i18n key for `fontSizeDefaultLabel` as it was identical to `default` key that also existed. 

Screenshot 2026-05-01 at 1 17 49 PM


mkilpatrick (dismissed):

packages/visual-editor/src/fields/styledFields/index.ts
benlife5, line 1: Should this file be exported from `packages/visual-editor/src/fields/index.ts`?

asanehisa: Yup good call!

mkilpatrick (dismissed):

benlife5 (approved):

briantstephan (approved):

PR #1199, opened 2026-05-04
feat: add styledLink option for fields

asanehisa:
Very similar to past two PRs for styledText and styledButton but it has includeCaret and letterSpacing.
Once again not used in any components in this branch (yet). 

Screenshot 2026-05-04 at 2 52 11 PM


packages/visual-editor/src/fields/styledFields/StyledLinkField.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+const defaultStyledLinkValue: StyledLinkValue = {
+  ...defaultBaseTextStyles,
+  letterSpacing: "default",
+  includeCaret: "default",
+};
+
+/**
+ * A custom field override for configuring link typography, letter spacing,
+ * and an optional caret indicator.
```
jwartofsky-yext, line 35: Do we want character spacing and text transform to also be optional?

jwartofsky-yext (approved):
LGTM!

I'm assuming these will be exclusive to the generated components?

benlife5 (approved):

mkilpatrick (approved):

packages/visual-editor/src/fields/styledFields/StyledLinkField.tsx
asanehisa: Possibly? I think not yet so the skill is "forced" to have these fields for now

PR #1200, opened 2026-05-04
feat: add styledPageSection option to field

asanehisa:
Tested with GridSection but is not being used in this branch. Will be used in generated components (in other branch). 

https://github.com/user-attachments/assets/65d10a3d-0f3b-48f5-b41a-235d1c3f5a1f


packages/visual-editor/src/fields/styledFields/StyledPageSection.tsx
```diff
@@ -0,0 +1,104 @@
+import { BaseField, FieldLabel, FieldProps } from "@puckeditor/core";
+import { MsgString, pt } from "../../utils/i18n/platform.ts";
+import { Combobox } from "../../internal/puck/ui/Combobox.tsx";
+import { ThemeOptions } from "../../utils/themeConfigOptions.ts";
+import { withDefaultOption } from "./baseText.tsx";
+
+export type StyledPageSectionValue = {
+  contentWidth: string;
```
benlife5: Same comment as on @briantstephan's PR -- can these be string union types because they are dropdown fields, or does that cause more problems than its worth?

asanehisa: Yup they can!

mkilpatrick (approved):

benlife5 (approved):

PR #1201, opened 2026-05-04
refactor: remove YextField

briantstephan:
This removes the YextField helper, which was no more than a basic wrapper after our recent field type refactors.

Tested on the site linked below and didn't notice any issues. Feel free to play around with it.
https://dev.yext.com/s/1000167375/yextsites/67153/pagesets


packages/visual-editor/src/components/pageSections/AboutSection/AboutSectionDetailsColumn.tsx
```diff
@@ -213,7 +213,7 @@ const aboutSectionDetailsColumnFields: YextFields<AboutSectionDetailsColumnProps
         content: {
           type: "custom",
           label: msg("fields.content", "Content"),
-          render: (props) => {
+          render: (props: any) => {
```
benlife5: Would prefer less any typing through this section, if possible

packages/visual-editor/src/components/LocatorResultCard.tsx
```diff
@@ -341,7 +340,7 @@ const LOCATOR_IMAGE_CONSTANT_CONFIG: ImageField = {
 const getDisplayFieldOptions = (
   fieldTypeId: string | string[]
 ): EmbeddedStringOption[] => {
-  // TODO: This breaks the rule of hooks, refactor after YextField is converted to a fieldType
+  // TODO: This breaks the rule of hooks, refactor the custom render path
```
benlife5, line 343: Not sure the context here, should this be handled in this PR?

mkilpatrick: The issue was that this function uses hooks and is dynamically called by the fieldType code, which is not within a React component, so it breaks the rule of hooks even though it works fine. I had left that comment to try and fix, but it probably needs looking into outside the scope of YextField now (since YextField wasn't even being used for this anymore). We can circle back to this one later.

packages/visual-editor/src/fields/fields.ts
```diff
@@ -62,28 +57,59 @@ export type YextPuckFields = {
...
 
-// TODO(SUMO-8378): Remove this and make YextFieldsInternal -> YextFields once Puck fixes their objectField typing
 export type YextFields<
   T extends DefaultComponentProps = DefaultComponentProps,
-> = YextFieldsInternal<T> & YextFieldMap<T>;
-
-type YextFieldsInternal<
-  T extends DefaultComponentProps = DefaultComponentProps,
-> = NonNullable<YextComponentConfig<T>["fields"]>;
+> = Fields<T, any>;
```
mkilpatrick, line 117: Wow you were able to remove YextFieldsInternal without the Puck change for objectFields?

briantstephan: Asked codex about it and it seems like the change it applied here just broadened the type enough to work around the issue. I've restored what we had before since that change doesn't actually fix the issue.

packages/visual-editor/src/components/pageSections/AboutSection/AboutSectionDetailsColumn.tsx
briantstephan: Done

benlife5 (dismissed):

packages/visual-editor/src/index.ts
```diff
@@ -4,3 +4,4 @@ export * from "./types/index.ts";
 export * from "./editor/index.ts";
 export * from "./components/index.ts";
 export * from "./fields/index.ts";
+export { type YextFieldDefinition } from "./fields/index.ts";
```
asanehisa: isn't this exported twice now from `export * from "./fields/index.ts";` and `export { type YextFieldDefinition } from "./fields/index.ts";`

briantstephan: It was like that because there were two types with the name `YextFieldDefinition`, so one needed to be explicitly exported. I've refactored slightly to rename one to `YextSchemaFieldDefinition`, since it's the type for the definition of `YextSchemaField`, and removed the extra export.

benlife5 (approved):

packages/visual-editor/src/index.ts
asanehisa: Ohh I see haha

asanehisa (approved):

mkilpatrick (approved):

PR #1205, opened 2026-05-05
chore: merge main branch into custom template branch

briantstephan:


mkilpatrick:
What does the `sync-main-to-skill` branch have that `main` doesn't? Do we need that to merge from that? Seems weird to have the interstitial branch.

briantstephan:
> What does the `sync-main-to-skill` branch have that `main` doesn't? Do we need that to merge from that? Seems weird to have the interstitial branch.

Nothing, it's just a branch off of `2026-custom-components-templates`. I just set it up this way so I could put up a PR (since it involved resolving conflicts) rather than merging `main` into `2026-custom-components-templates` directly.

benlife5 (commented):
Generally LGTM, thanks!
1. I think we don't need the main default layout data in this branch
2. We should probably have an item for teaching the skill about the new `Main` wrapper
3. Can you look into what's the best way to merge this? I feel like we often have to re-resolve merge conflicts all the time when we do these branches and I wonder if Squash & Merge isn't the best option for this

benlife5 (approved):

asanehisa (commented):
Curious if the pnpm run dev flow in /starter still works and if the npm run dev flow in ~/pages-visual-editor-starter works with the local editor shell

briantstephan:
> Generally LGTM, thanks!
> 
> 1. I think we don't need the main default layout data in this branch
> 2. We should probably have an item for teaching the skill about the new `Main` wrapper
> 3. Can you look into what's the best way to merge this? I feel like we often have to re-resolve merge conflicts all the time when we do these branches and I wonder if Squash & Merge isn't the best option for this

1. I've removed it. Maybe it would be helpful to add a skill that performs the merge, with guidance about what files should be included or excluded?
2. I updated the upcoming item for teaching the skill about the field types to also add guidance for `MainContent`.
3. I think temporarily enabling "Allow merge commits" and using that option would be the best here for preserving the commit history and making future merges go smoother.

asanehisa (approved):
didn't look closely but lgtm

briantstephan:
> Curious if the pnpm run dev flow in /starter still works and if the npm run dev flow in ~/pages-visual-editor-starter works with the local editor shell

I tested and the only issue is `MainContent` getting added to existing custom templates' layouts via the migration but not to the template config. I'll address this as part of the skill updates in a follow-up PR.

PR #1207, opened 2026-05-06
feat: allow boolean display fields in locator

anguyen-yext2:
Allow boolean display fields in locator result card headings (Primary Heading, Secondary Heading, and Tertiary Heading). When a boolean field is selected, allow the users to set what should display in the result card for true and false cases via 2 static text boxes.

J=WAT-5527
TEST=manual

tested in dev mode, Rob approved UI
https://yext.slack.com/archives/C0A1G447MPD/p1778015267221689


packages/visual-editor/src/components/Locator.tsx
```diff
@@ -243,26 +250,15 @@ const ResultCardPropsField = ({
...
-        enabled: value.tertiaryHeading?.constantValueEnabled ?? false,
-      },
-      {
-        key: "image",
-        enabled: value.image?.constantValueEnabled ?? false,
-      },
-    ];
+    BOOLEAN_SUPPORTED_FIELDS.forEach((key) => {
+      const headingConfig = value[key];
+      const enabled = headingConfig?.constantValueEnabled ?? false;
```
Fondryext: tiny nit: I know it's clear from this line what 'enabled' is referring to, but I feel like later down it becomes a little muddy, like

`const isBooleanFieldSelected = !enabled && fieldTypeId === "type.boolean";`

reads a little weird. Maybe we can just still call it 'constantValueEnabled'? I don't feel strongly, though.

Fondryext (dismissed):

packages/visual-editor/src/components/Locator.tsx
anguyen-yext2: that's very fair! updated

packages/visual-editor/src/components/Locator.tsx
```diff
@@ -243,40 +250,53 @@ const ResultCardPropsField = ({
...
-      {
-        key: "tertiaryHeading",
-        enabled: value.tertiaryHeading?.constantValueEnabled ?? false,
-      },
-      {
-        key: "image",
-        enabled: value.image?.constantValueEnabled ?? false,
-      },
-    ];
+    BOOLEAN_SUPPORTED_FIELDS.forEach((key) => {
```
jwartofsky-yext, line 253: This logic for a selector + alternate fields seems like it should be separated from Locator and made available to other components

packages/visual-editor/src/components/LocatorResultCard.tsx
```diff
@@ -402,9 +414,21 @@ export const LocatorResultCardFields: YextObjectField<LocatorResultCardProps> =
             type: "translatableString",
             showApplyAllOption: false,
             showFieldSelector: true,
-            getOptions: () => getDisplayFieldOptions("type.string"),
+            getOptions: () => getDisplayFieldOptions(["type.string"]),
+          },
+          field: DisplayFieldSelector(["type.string", "type.boolean"]),
+          trueDisplayText: {
+            type: "translatableString",
+            label: msg("fields.whenTrue", "When true"),
```
asanehisa, line 422: When True ? or make consistent with below

packages/visual-editor/src/components/Locator.tsx
anguyen-yext2: but aren't display fields specific to locator only? I wasn't aware that `locatorDisplayFields` are used outside of locator

oh or do you mean we should allow boolean fields to be selected in other field selector too (and display true/false text box logic if a boolean is selected)?

packages/visual-editor/src/components/LocatorResultCard.tsx
anguyen-yext2: oops. Fixed!

packages/visual-editor/src/components/Locator.tsx
jwartofsky-yext: yes, I was thinking we would extract out this logic from Locator.tsx since Locator.tsx is over 2000 lines long

The logic for the toggle-able field selector that is

You can just put it in another file

Feel free to ack to close out the item and Locator.tsx can be split up as another item. Codex can split it up into multiple files in one-shot if you use plan + the grill-me skill

anguyen-yext2: gotcha. I'll ack for now and keep in mind to refactor stuff to separate files in the future

Fondryext (approved):

asanehisa (approved):

PR #1208, opened 2026-05-06
feat: allow editing of the directory card's title

benlife5:
Uses the embedded field utilities to support modifying the directory card title

https://github.com/user-attachments/assets/d192a5a5-3a71-4dcc-b6cf-7af15ffa0815


benlife5:
We probably need conditional statements in embedded fields to make this more useful (so you could do `[[name]] - [[geomodifier]]` and get `Galaxy Grill - 2nd Floor`  but if geomodifier doesn't exist you wouldn't get `Galaxy Grill - `)

asanehisa (dismissed):
nice! for the labels like `label: pt("addressLine1", "Address > Line 1"),` I thought for other usages, we built it off the stream data we had?

packages/visual-editor/src/components/directory/DirectoryCard.tsx
```diff
@@ -321,16 +338,20 @@ const DirectoryCardComponent: PuckComponent<DirectoryCardProps> = (props) => {
 
     if (
       JSON.stringify(sharedCardProps?.cardStyles) === JSON.stringify(styles) &&
-      JSON.stringify(sharedCardProps?.slotStyles) === JSON.stringify(slotStyles)
+      JSON.stringify(sharedCardProps?.slotStyles) ===
```
jwartofsky-yext: Is there a library to make these nicer? Perhaps deepEquals?

packages/visual-editor/src/components/directory/DirectoryCard.tsx
```diff
@@ -321,16 +338,20 @@ const DirectoryCardComponent: PuckComponent<DirectoryCardProps> = (props) => {
...
       return;
     }
 
     setSharedCardProps({
       cardStyles: styles,
       slotStyles: slotStyles,
+      headingText,
     });
-  }, [styles, slotStyles]);
+  }, [headingText, styles, slotStyles]);
```
jwartofsky-yext: nit, but the order here is different

packages/visual-editor/src/components/directory/DirectoryCardTitleSlot.tsx
```diff
@@ -0,0 +1,162 @@
...
+            { label: pt("name", "Name"), value: "name" },
+            { label: pt("slug", "Slug"), value: "slug" },
+            { label: pt("geomodifier", "Geomodifier"), value: "geomodifier" },
+            { label: pt("id", "ID"), value: "id" },
+            {
+              label: pt("addressLine1", "Address > Line 1"),
+              value: "address.line1",
+            },
+            {
+              label: pt("addressLine2", "Address > Line 2"),
```
jwartofsky-yext: Are these hardcoded options to be set in the embedded field picker to select subfields from a specific directory entity?

My big PR may be helpful here, but this seems fine in the interim

benlife5: Yeah was realizing there's probably some useful stuff in your PR for this. But yeah basically we just want the entity fields that are available on dm_directoryChildren and apiNamesToDisplayNames doesn't include those fields so I was just hardcoding these translations. Will take more of a look tomorrow

packages/visual-editor/src/components/directory/DirectoryCard.tsx
benlife5: Yeah we do have an `isDeepEqual` function, but all the slots use string comparisons. Might want to switch, but I'm going to consider that a wider refactor

benlife5:
@asanehisa 

> for the labels like label: pt("addressLine1", "Address > Line 1"), I thought for other usages, we built it off the stream data we had?

Yeah we pass the display names but not for the dm subfields, will TAL if there's a better way to handle tomorrow

packages/visual-editor/src/components/directory/DirectoryCardTitleSlot.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const streamDocument = useDocument();
+  const { i18n } = useTranslation();
+
+  const justifyClass = styles?.align
+    ? {
+        left: "justify-start",
+        center: "justify-center",
+        right: "justify-end",
+      }[styles.align]
+    : "justify-start";
```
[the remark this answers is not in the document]
benlife5: Styles is a required prop

benlife5:
@asanehisa @jwartofsky-yext Normally we get the field display names via the templateStream, but the templateStream doesn't include the display names for the dm children's subfields (even with the linked entity changes because the dm children are automatically included, separate from the new linked entity process). I think hardcoded translations is the easiest way forward since it's just this limited set of subfields for the directory card titles but let me know if you disagree

jwartofsky-yext:
> @asanehisa @jwartofsky-yext Normally we get the field display names via the templateStream, but the templateStream doesn't include the display names for the dm children's subfields (even with the linked entity changes because the dm children are automatically included, separate from the new linked entity process). I think hardcoded translations is the easiest way forward since it's just this limited set of subfields for the directory card titles but let me know if you disagree

I think this is fine for now

Note that I removed "LINKED_ENTITY_SCHEMA" recently. On my PR I am seeing the displayNames coming through for the linked fields, but I haven't checked directory. Perhaps we could update something on the Octopus side to make these come through so we don't need translations? I think this workaround is OK for now and we can update it later.

packages/visual-editor/src/components/directory/DirectoryCard.tsx
asanehisa: I imagine string comparisons are quicker too than a full isDeepEqual

packages/visual-editor/src/components/directory/DirectoryCardTitleSlot.tsx
```diff
@@ -0,0 +1,162 @@
...
+};
+
+const DirectoryCardTitleSlotComponent: PuckComponent<
+  DirectoryCardTitleSlotProps
+> = (props) => {
+  const { data, styles, puck } = props;
+  const streamDocument = useDocument();
+  const { i18n } = useTranslation();
+
+  const justifyClass = styles?.align
```
asanehisa: not u but just noting we could probably have made this a helper that's exported since we do it a lot and the skill likely will too lol

asanehisa (dismissed):

packages/visual-editor/src/components/directory/DirectoryCardTitleSlot.tsx
```diff
@@ -0,0 +1,162 @@
...
+          color={styles.color}
+        >
+          {resolvedHeadingText}
+        </Heading>
+      </EntityField>
+    </div>
+  ) : puck.isEditing ? (
+    <div className="h-[30px]" />
+  ) : (
+    <></>
```
briantstephan: nit, I feel like it's generally cleaner to return `null` instead of `<></>`

packages/visual-editor/src/components/directory/DirectoryCardTitleSlot.tsx
```diff
@@ -0,0 +1,162 @@
...
+    objectFields: {
+      text: {
+        type: "translatableString",
+        label: msg("fields.text", "Text"),
+        filter: {
+          types: ["type.string"],
+        },
+        showApplyAllOption: true,
+        getOptions: () => {
+          return [
```
briantstephan: super nit, I think you can drop the `return` and just do `getOptions: () => ([...])`

benlife5: Done

packages/visual-editor/src/components/directory/DirectoryCardTitleSlot.tsx
benlife5: Unfortunately the `PuckComponent` requires a `React.JSX.Element` return value and null doesn't work

asanehisa (dismissed):

briantstephan (dismissed):

packages/visual-editor/locales/platform/en/visual-editor.json
```diff
@@ -6,6 +6,12 @@
   "addEntityField": "Add entity field",
   "addImage": "Add Image",
   "address": "Address",
+  "addressCity": "Address > City",
```
mkilpatrick: Do these need translation? Do we do that for the other entity fields?

packages/visual-editor/src/components/directory/DirectoryCardTitleSlot.tsx
```diff
@@ -0,0 +1,160 @@
...
+    objectFields: {
+      text: {
+        type: "translatableString",
+        label: msg("fields.text", "Text"),
+        filter: {
+          types: ["type.string"],
+        },
+        showApplyAllOption: true,
+        getOptions: () => [
+          { label: pt("name", "Name"), value: "name" },
```
mkilpatrick: Can we not get the options from the schema? It's not currently logged so I can't verify what's actually in there.

packages/visual-editor/locales/platform/en/visual-editor.json
benlife5: The tools we have are built for getting  main entity document fields and display names, not the subfields of a linked dm field. I've come around that we should do this properly, but I think we should wait for Jake's change first because it includes a decent bit of the work that would be needed here

mkilpatrick: Does that mean you want to sit on the PR until then, or ship this as is?

benlife5: I think wait, presuming the linked entity PR will ship in the next day or so

benlife5:
New linked entity utils make this much nicer... going to test in platform and then will re-request review
Edit: tested & confirmed

mkilpatrick (dismissed):

mkilpatrick:
lg, though should the mapping be a normal mappable field with static content/embedding as an option?

jwartofsky-yext:
Are you going to update this to use the new createSlottedItemSource?

packages/visual-editor/src/components/directory/DirectoryWrapper.tsx
```diff
@@ -22,19 +25,101 @@ import {
...
+      type: "entityField",
+      label: msg("fields.name", "Name"),
+      filter: {
+        types: ["type.string"],
+      },
+    },
+  },
+});
+
+const getNormalizedDirectoryGridData = (
```
jwartofsky-yext, line 70: would be good to have the robots generate some comments/explanation for these

I think they might also be able to trim a lot of this code

packages/visual-editor/src/components/migrations/0077_directory_card_title_field.ts
```diff
@@ -0,0 +1,198 @@
+import { Migration } from "../../utils/migrate.ts";
+import {
+  defaultDirectoryCardSlotData,
+  DirectoryCardProps,
```
jwartofsky-yext: Should we be importing these structs? In a later state (ex: migration 100+), these might have changed again which would make this migration out of date.

packages/visual-editor/src/components/directory/DirectoryWrapper.tsx
```diff
@@ -22,19 +25,101 @@ import {
...
+  manualSlots?: {
+    CardSlot: Slot;
+  };
   slots: {
     CardSlot: Slot;
   };
 };
 
+const directoryCardsSource = createSlottedItemSource<
+  {
```
jwartofsky-yext: Could this be a type somewhere instead of a combination?

packages/visual-editor/src/components/directory/DirectoryCard.tsx
jwartofsky-yext, line 1: This might be getting a bit out of scope, but I wonder if we should make all of the fields on the directory cards mappable and have Codex rewrite the component using the createSlottedItemSource

That would likely result in simpler code with most of the complication being left to the migration

Feel free to ack

packages/visual-editor/src/components/directory/DirectoryWrapper.tsx
benlife5: Done

packages/visual-editor/src/components/migrations/0077_directory_card_title_field.ts
benlife5: Good call, done

packages/visual-editor/src/components/directory/DirectoryWrapper.tsx
benlife5: Updated

packages/visual-editor/src/components/directory/DirectoryCard.tsx
benlife5: I think it would simplify the code but since we only include the basic address/phone/hours on the directory stream, there wouldn't be any actual customization possible, so the UI wouldn't be helpful

mkilpatrick (approved):

packages/visual-editor/src/components/directory/DirectoryCard.tsx
jwartofsky-yext: Ack

jwartofsky-yext (approved):

PR #1209, opened 2026-05-07
fix: add MainContent to custom template config

briantstephan:
This adds MainContent to custom template configs so that migrated layout data works correctly, and newly created layout with MainContent work as well.

Tested in the local editor and verified that an existing custom template that was previously breaking due to the MainContent migration was now working with this change.


jwartofsky-yext (commented):
Do we also need the Header/Footer sections?

benlife5 (approved):

briantstephan:
> Do we also need the Header/Footer sections?

No, we only need `MainContent` because it's provided by visual-editor, while header and footer are generated and registered as template-specific custom components when the skill runs.

jwartofsky-yext (approved):

mkilpatrick (approved):

PR #1211, opened 2026-05-11
chore: make toPuckFields a runtime adapter instead of a TS cast

mkilpatrick:
This turns toPuckFields into a true adapter instead of just a TS cast because actual usage was causing console errors. Also add some exports that should've been exported.

Context:
- We export Yext field types like `basicSelector`, `ctaSelector`, `entityField`, `multiSelector`, etc.
- We also export `YextPuckFieldOverrides` and `YextAutoField`.
- Today `toPuckFields()` is basically:

```
  export const toPuckFields = <Props extends DefaultComponentProps = DefaultComponentProps>(
    fields: YextFields<Props> | YextFieldMap<Props>
  ): Fields<Props> => fields as unknown as Fields<Props>;
```

Problem:
- Puck’s `overrides.fieldTypes.basicSelector` is not enough.
- Puck’s `AutoFieldInternal` still tries to render a default child renderer from `defaultFields[field.type]`.
- For `basicSelector`, `defaultFields.basicSelector` is undefined, causing:

  React.jsx: type is invalid -- expected a string ... but got undefined
  Check the render method of AutoFieldInternal


packages/visual-editor/src/fields/fields.ts
```diff
@@ -142,29 +106,20 @@ export type YextFieldMap<
   [PropName in keyof Omit<T, "editMode">]: YextFieldDefinition<T[PropName]>;
 };
 
-// For things like resolveFields, Puck does not currently let you override the return type
-// so we need this function to satisfy the typing.
 export const toPuckFields = <
```
jwartofsky-yext, line 120: Can you add a jsdoc explaining what this does and why?

packages/visual-editor/src/fields/yextFieldAdapter.ts
```diff
@@ -0,0 +1,41 @@
+import type { Field } from "@puckeditor/core";
+import type { YextFieldDefinition, YextPuckField } from "./fields.ts";
+import { isYextOverrideType } from "./fieldOverrides.ts";
+
+type YextFieldRender = (field: YextPuckField) => Field<any>;
+
+export const adaptYextField = (
```
jwartofsky-yext, line 7: Add a jsdoc that explains what this does and why we want to use it

packages/visual-editor/src/fields/yextFieldAdapter.ts
```diff
@@ -0,0 +1,41 @@
...
+    return {
+      ...field,
+      arrayFields: adaptYextFieldMap(field.arrayFields, renderYextField),
+    } as Field<any>;
+  }
+
+  return field as Field<any>;
+};
+
+export const adaptYextFieldMap = (
```
jwartofsky-yext, line 41: Add jsdoc explaining what this does

jwartofsky-yext (commented):
lgtm, just need jsdocs for more clarity for people looking at this code in the future

packages/visual-editor/src/fields/yextFieldAdapter.ts
mkilpatrick: Covered by adaptYextFieldMap

jwartofsky-yext (approved):

briantstephan (approved):

PR #1215, opened 2026-05-17
fix: embedded field pickers in mapping are not scoped

jwartofsky-yext:
Entity Field pickers under Mappings in itemSource and slottedItemSource are scoped to the items selected in the parent field.

However embedded field pickers did not have this scoping. This change fixes the embedded field pickers to respect the scope.

Also includes passing the sourceField to the Rich Text editor so that it can scope to subfields properly.


packages/visual-editor/src/fields/repeatedSourceFieldContext.ts
```diff
@@ -0,0 +1,5 @@
+import React from "react";
+
+export const RepeatedSourceFieldContext = React.createContext<
```
mkilpatrick, line 3: Should this live in the utils/itemSource folder?

mkilpatrick (approved):

benlife5 (approved):

packages/visual-editor/src/fields/repeatedSourceFieldContext.ts
jwartofsky-yext: I think this location is appropriate

It's not directly used by any of the helpers under utils/itemSource

This is on the consuming side that uses the itemSource

PR #1221, opened 2026-05-26
feat: add comprehensive CTA

benlife5:
Our CTA props, resolution, and rendering are split across many components and slots. This attempts to create a single field type and React component with all the major CTA behavior. This should simplify it for the skill while leaving the existing complex OOTB behavior unaffected.

https://github.com/user-attachments/assets/0bc56f90-eacc-4fe4-b7ef-a1b619d9fbb1


packages/visual-editor/src/fields/styledFields/ComprehensiveCTAField.tsx
```diff
@@ -0,0 +1,310 @@
...
+  const showColorField =
+    isCtaVariantWithColor(currentValue.styles.variant) && !showPresetImageField;
+  const showButtonStyleFields =
+    !showPresetImageField && currentValue.styles.variant !== "link";
+  const showLinkStyleFields =
+    !showPresetImageField && currentValue.styles.variant === "link";
+
+  const dataField = React.useMemo(
+    () => ({
+      type: "object" as const,
```
asanehisa: do we need these `as const`?

asanehisa (dismissed):
Awesome!

packages/visual-editor/src/fields/styledFields/ComprehensiveCTAField.tsx
benlife5: good call

asanehisa (approved):

briantstephan (approved):

packages/visual-editor/src/components/contentBlocks/ComprehensiveCTA.tsx
```diff
@@ -0,0 +1,325 @@
...
+import { type ComprehensiveCTAValue } from "../../fields/styledFields/ComprehensiveCTAField.tsx";
+import {
+  FOOD_DELIVERY_SERVICES,
+  type EnhancedTranslatableCTA,
+} from "../../types/types.ts";
+import { type StyledButtonValue } from "../../fields/styledFields/StyledButtonField.tsx";
+import { type StyledLinkValue } from "../../fields/styledFields/StyledLinkField.tsx";
+import { defaultBaseTextStyles } from "../../fields/styledFields/baseText.tsx";
+
+export type ComprehensiveCTARenderProps = {
```
mkilpatrick, line 17: How does this file work with ComprehensiveCTAField? I didn't see this called in that one.

packages/visual-editor/src/fields/styledFields/ComprehensiveCTAField.test.tsx
```diff
@@ -0,0 +1,323 @@
...
+import { YextAutoField } from "../YextAutoField.tsx";
+import {
+  type ComprehensiveCTAField,
+  type ComprehensiveCTAValue,
+} from "./ComprehensiveCTAField.tsx";
+import { type StyledButtonValue } from "./StyledButtonField.tsx";
+import { type StyledLinkValue } from "./StyledLinkField.tsx";
+
+const field: ComprehensiveCTAField = {
+  type: "comprehensiveCTA",
```
mkilpatrick, line 41: Should this completely overtake CTASelectorField?

packages/visual-editor/src/components/contentBlocks/ComprehensiveCTA.tsx
benlife5: This is an atom for the skill to use that accepts all the props that ComprehensiveCTAField spits out, the idea being that we don't want the skill to need to do a new implementation of all of the cta options every time (think preset images, open in new tab, get directions link, etc.)

packages/visual-editor/src/fields/styledFields/ComprehensiveCTAField.test.tsx
benlife5: I think we need to keep it for now unless we want to do a refactor of the OOTB components:
In the OOTB components, we have the Data / Style separation and so CTA options are usually split between those section (CTASelectorField is the data fields) using resolveFields to connect them. Data and styling is very intertwined for CTAs since there's different styling options depending on what type of data has been selected. My change here bundles it all together so that it is simpler but I don't think it's a drop in replacement for the usages in the OOTB components

mkilpatrick (approved):

briantstephan (approved):

PR #1225, opened 2026-06-02
fix: comingSoon in locator results

asanehisa:
Before:
Screenshot 2026-06-02 at 11 56 38 AM

Now the accordion is hidden when comingSoon is true.


mkilpatrick (dismissed):

briantstephan (dismissed):

packages/visual-editor/src/components/LocatorResultCard.tsx
```diff
@@ -1249,58 +1249,76 @@ const HoursSection = (props: {
...
-                  boldCurrentStatus={true}
-                />
-              </div>
-            </AccordionTrigger>
-            <AccordionContent
-              id={contentId}
-              aria-labelledby={triggerId}
-              role="region"
+        {comingSoon ? (
+          <div className="flex flex-row items-center gap-2">
```
jwartofsky-yext: nit, but you could use a function so that this div can be used both here and in the "not coming soon" case without having to duplicate it

asanehisa: Yup will do!

jwartofsky-yext (approved):
lgtm!

mkilpatrick (approved):

PR #1227, opened 2026-06-02
feat: mapbox uses YEXT_MAPBOX_API_KEY

jwartofsky-yext:
Removes the "Api Key" prop from the Static Map Section and Mapbox Static Map components.

They now use the key from the environment variable.

If the key fails to load, the component will not appear on the live page.

Tested using local starter.


packages/visual-editor/src/components/contentBlocks/MapboxStaticMap.tsx
```diff
@@ -81,8 +75,21 @@ export const MapboxStaticMapComponent: PuckComponent<MapboxStaticProps> = ({
...
+    typeof document === "undefined"
+      ? undefined
+      : (document.getElementById("preview-frame") as HTMLIFrameElement);
+  let mapboxApiKey = streamDocument._env?.YEXT_MAPBOX_API_KEY;
+  if (
+    iframe?.contentDocument &&
+    streamDocument._env?.YEXT_EDIT_LAYOUT_MODE_MAPBOX_API_KEY
+  ) {
+    mapboxApiKey = streamDocument._env.YEXT_EDIT_LAYOUT_MODE_MAPBOX_API_KEY;
+  }
```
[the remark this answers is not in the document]
jwartofsky-yext: This is the same logic as used by the Locator component which uses a similar mechanism for fallback

See: packages/visual-editor/src/components/Locator.tsx line 1194

Manual testing confirms that this works as expected, as does the Locator component

The YEXT_EDIT_LAYOUT_MODE_MAPBOX_API_KEY shouldn't ever be used on a real, live page. It is only for the Editor.

packages/visual-editor/locales/platform/en/visual-editor.json
```diff
@@ -678,7 +677,7 @@
   },
   "spacingDefaultLabel": "Default",
   "staticContent": "Static content",
-  "staticMapEmptyStateAddApiKey": "Add an API key to preview your map",
+  "staticMapEmptyStateMissingApiKey": "Failed to load YEXT_MAPBOX_API_KEY. Please check your Site Settings.",
```
benlife5: Maybe `your site's advanced configuration` since that's the label for the Storm page

benlife5 (dismissed):

packages/visual-editor/locales/platform/en/visual-editor.json
jwartofsky-yext: Done

mkilpatrick (commented):
Migrate to delete the unused prop now

jwartofsky-yext:
> Migrate to delete the unused prop now

Done

briantstephan (approved):

asanehisa (commented):
is there a chance someone is actively using a diff API key and this messes things up for them?

jwartofsky-yext:
> is there a chance someone is actively using a diff API key and this messes things up for them?

It's possible, but I doubt anything would break. The new key would work and they wouldn't need their old key. I started a thread in #team-data-syndication.

From some quick research, there isn't really much to do with a mapbox API key besides render mapbox and track how much your API key has been used, so I don't think any customers would have a strong need for their own key.

https://yext.slack.com/archives/C0A2SGK32GJ/p1780509524438649

asanehisa (approved):
lgtm assuming ppl in data-sync are ok with it

PR #1233, opened 2026-06-09
fix: apply template name directly to config

briantstephan:
Sometimes, the minified js for the template puts the `editTemplateName` const below the `config` const, which makes the editor fail when `config` tries to find a variable that hasn't been assigned yet. This seems to be a bit sporadic, as some templates have had no issues while others break entirely. Assigning the `__YEXT_VISUAL_EDITOR_TEMPLATE_NAME__` directly in the two places it is used should be a safer way to handle this.

This screenshot shows a snippet from the minified `edit-[templateName]` template from a recent attempt at a template push. `Q` appears to be `config` ,and `H` appears to be `editTemplateName`, which has not yet been assigned when `Q` uses it.

Screenshot 2026-06-09 at 4 44 26 PM

### DIFF
--- a/packages/visual-editor/src/vite-plugin/routing/editorRoute.test.ts
+++ b/packages/visual-editor/src/vite-plugin/routing/editorRoute.test.ts
@@ -122,13 +122,16 @@ describe("injectEditorTemplateInfo", () => {
     expect(
       injectEditorTemplateInfo(
         `const editPath = "${EDIT_PATH_PLACEHOLDER}";
-const editTemplateName = "${EDIT_TEMPLATE_NAME_PLACEHOLDER}";`,
+export const config = { name: "${EDIT_TEMPLATE_NAME_PLACEHOLDER}" };
+export const getHeadConfig = () => ({ title: "${EDIT_TEMPLATE_NAME_PLACEHOLDER}" });`,
         {
           path: "edit",
           configName: "edit",
         }
       )
-    ).toBe('const editPath = "edit";\nconst editTemplateName = "edit";');
+    ).toBe(`const editPath = "edit";
+export const config = { name: "edit" };
+export const getHeadConfig = () => ({ title: "edit" });`);
   });
 
   it("replaces all occurrences of both placeholders", () => {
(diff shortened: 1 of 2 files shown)


packages/visual-editor/src/vite-plugin/routing/editorRoute.test.ts
```diff
@@ -122,31 +122,36 @@ describe("injectEditorTemplateInfo", () => {
...
+export const getHeadConfig = () => ({ title: "${EDIT_TEMPLATE_NAME_PLACEHOLDER}" });`,
         {
           path: "edit",
           configName: "edit",
         }
       )
-    ).toBe('const editPath = "edit";\nconst editTemplateName = "edit";');
+    ).toBe(`const editPath = "edit";
+export const config = { name: "edit" };
+export const getHeadConfig = () => ({ title: "edit" });`);
```
jwartofsky-yext, line 134: Do these need to be exported?

packages/visual-editor/src/vite-plugin/templates/edit.tsx
```diff
@@ -33,7 +33,6 @@ const componentRegistry: Record<string, Config<any>> = {
 };
 
 const editPath = "__YEXT_VISUAL_EDITOR_PATH__";
-const editTemplateName = "__YEXT_VISUAL_EDITOR_TEMPLATE_NAME__";
```
jwartofsky-yext, line 36: It's strange that we have to make this change. I would think MinifiedJS wouldn't introduce errors like variables being out of order?

mkilpatrick (approved):

benlife5 (approved):
🙌 I saw this issue yesterday but hadn't had a change to look into it yet

packages/visual-editor/src/vite-plugin/routing/editorRoute.test.ts
briantstephan: They're exported in `edit.tsx` which is where this comes from, so this updates it to match.

packages/visual-editor/src/vite-plugin/templates/edit.tsx
briantstephan: I agree. I'm still not quite clear on why this is only an issue now, but the minified js clearly shows the issue as coming from here. I'm not sure what else we could try, and this should be a safe change either way.

asanehisa (approved):

PR #1240, opened 2026-06-17
chore: update esbuild

asanehisa:
For this [vuln](https://yext.atlassian.net/browse/VULN-44969?atlOrigin=eyJpIjoiMWY4NmMxOGU2ZDE3NDgyMjgxYjAyMTYzOTA2NjUwNGYiLCJwIjoiaiJ9), have esbuild be 0.28.1 everywhere.

Additionally removed some dependencies that weren't used in packages/visual-editor and moved others only used in dev to devDependencies. 

And aligned all the nodes to `"node": "^20.6.0 || ^22 || ^24"` (already was done in other package.json files)


packages/visual-editor/vite.config.ts
```diff
@@ -15,7 +15,13 @@ export default defineConfig(() => ({
       "@yext/visual-editor": path.resolve(__dirname, "src"),
     },
   },
+  optimizeDeps: {
+    esbuildOptions: {
+      target: "es2022",
```
mkilpatrick, line 20: Why do we need to target an older version after upgrading esbuild?

asanehisa: Oh it makes the upgraded esbuild compile to the JS version we're using or else it breaks when testing

mkilpatrick: What is "JS version we're using?" Like Node version? And is this only for running tests when you say "testing" or do you mean testing the full flow/working for real? If this is specific to running tests only then we shouldn't have it do this for the prod output.

asanehisa: Hmm so [target](https://esbuild.github.io/api/#target) by default assumes all of the latest JavaScript and CSS features are supported, but for us it isn't as we are on Node 20, 22, 24. Google says Node version + browser determines which targets are supported, so it is Node dependent. 

And sorry it is for running tests (including testing in fake starter with `pnpm run dev`) and building (`pnpm build`). Or else I see
`│ Transforming destructuring to the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14"…` errors when testing and/or building.

mkilpatrick: Ok, just weird we didn't have to do this before

mkilpatrick (approved):

packages/visual-editor/vite.config.ts
asanehisa: Yup was just because esbuild got upgraded to newest while the rest didn't.

PR #1241, opened 2026-06-18
fix: comprehensive cta qa

benlife5:
1. Don't show the "open in new tab" icon when using a presetImage
2. Don't use the link from the Puck fields when using getDirections

### DIFF
--- a/packages/visual-editor/src/components/contentBlocks/ComprehensiveCTA.tsx
+++ b/packages/visual-editor/src/components/contentBlocks/ComprehensiveCTA.tsx
@@ -302,7 +302,7 @@ export const ComprehensiveCTA = ({
       id={actionType === "button" ? currentValue.data.customId : undefined}
       label={effectiveLabel}
       link={
-        actionType === "link" && resolvedCta
+        actionType === "link" && ctaType !== "getDirections" && resolvedCta
           ? resolveComponentData(resolvedCta.link, locale, streamDocument)
           : undefined
       }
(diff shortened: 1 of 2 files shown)


asanehisa (approved):

packages/visual-editor/src/components/contentBlocks/ComprehensiveCTA.tsx
```diff
@@ -302,7 +302,7 @@ export const ComprehensiveCTA = ({
       id={actionType === "button" ? currentValue.data.customId : undefined}
       label={effectiveLabel}
       link={
-        actionType === "link" && resolvedCta
+        actionType === "link" && ctaType !== "getDirections" && resolvedCta
```
mkilpatrick, line 305: Are we sure we want this? CTAs getDirections has this, which is contradictory already (prefer hardcoded link...user setting link props should not be used). Was it maybe to handle some old layouts?
```
// Prefer hardcoded link, then listings link, then coordinate link
// User settable link props should not be used for get directions
link: resolvedLink || listingsLink || coordinateLink || "#",
```

mkilpatrick (approved):

packages/visual-editor/src/components/contentBlocks/ComprehensiveCTA.tsx
benlife5: Yeah the CTA code is kinda a mess. I think this behavior is because the Address can have a get directions link, which can be used in nearby locations & directory, and in that case we don't want the listingsLink/coordinateLink for the current entity but instead want to pass in a preconstructed link. I think the CTA slot component sets the link to undefined for getDirections, like this

mkilpatrick (approved):

PR #1242, opened 2026-06-18
feat: image url support for image fields in local editor

briantstephan:
This allows users of the local-editor to update image fields using a publicly-hosted image URL, like in the fake starter.

Demo:

https://github.com/user-attachments/assets/b3562003-9163-441a-a0eb-8c1dd01d3b67

### DIFF
--- a/packages/visual-editor/src/utils/isFakeStarterLocalDev.ts
+++ b/packages/visual-editor/src/utils/isFakeStarterLocalDev.ts
@@ -1,29 +1,26 @@
 type LocationLike = string | URL | { pathname: string };
 
-const getPathname = (locationLike: LocationLike) => {
+export const isFakeStarterLocalDevRoute = (locationLike: LocationLike) => {
   if (typeof locationLike === "string") {
-    return new URL(locationLike, "http://localhost").pathname;
+    return new URL(locationLike, "http://localhost").pathname.startsWith(
+      "/dev-"
+    );
   }
 
   if (locationLike instanceof URL) {
-    return locationLike.pathname;
+    return locationLike.pathname.startsWith("/dev-");
   }
 
-  return locationLike.pathname;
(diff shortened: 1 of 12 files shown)


packages/visual-editor/src/utils/isFakeStarterLocalDev.ts
```diff
@@ -16,6 +16,10 @@ export const isFakeStarterLocalDevRoute = (locationLike: LocationLike) => {
   return getPathname(locationLike).startsWith("/dev-");
 };
 
+export const isLocalEditorRoute = (locationLike: LocationLike) => {
```
mkilpatrick: Should we just update isFakeStarterLocalDev? If we want both, I'd make it a callable func without having to pass in variables like we do for isFakeStarterLocalDev. It's weird that that one handles window undefined and window.location but you make those calls manually in usages for this func.

briantstephan: I've replaced `isFakeStarterLocalDev` with `isLocalDev` which covers both the fake starter and local-editor and can be used in multiple places.

packages/visual-editor/src/utils/isLocalDev.ts
```diff
@@ -1,29 +1,26 @@
 type LocationLike = string | URL | { pathname: string };
 
-const getPathname = (locationLike: LocationLike) => {
+export const isFakeStarterLocalDevRoute = (locationLike: LocationLike) => {
```
mkilpatrick: Can you just do this directly in isLocalDev and comment the difference? If you do update the filename/test

briantstephan: Done

mkilpatrick (approved):

benlife5 (approved):

PR #1243, opened 2026-06-24
feat: support fonts in layout editor

benlife5:
Our templates have font selectors in the layout editor, however this has some issues:
1. On the live page, we only load the fonts that are published in the theme (not the layout)
2. In the platform editor, only the published fonts load, so the font dropdown is unstyled
3. In the local editor, no fonts load, so you cannot test this locally

This PR addresses these issues by
1. When running applyTheme, recursively scanning the layoutData for `fontFamily` and injecting any fonts found (in addition to fonts published in the theme
2. Loading all the fonts whenever the FontSelector is mounted, regardless of whether the editor is platform/local, layout/theme

Tested by generating a template artifact with a custom pack. Confirmed the fonts loaded as expected in the editors and live page. https://qa.yext.com/s/4116399/yextsites/273779/pagesets

### DIFF
--- a/packages/visual-editor/src/utils/fonts/visualEditorFonts.ts
+++ b/packages/visual-editor/src/utils/fonts/visualEditorFonts.ts
@@ -201,7 +201,6 @@ export const createFontLinkElements = (
  * to avoid appending duplicate font tags across rerenders.
  */
 export const loadFontsIntoDOM = (
-  document: Document,
   googleFonts: FontRegistry,
   customFonts: FontRegistry,
   idPrefix: string = "visual-editor-fonts"
@@ -528,10 +527,12 @@ export const extractFontFamiliesFromTheme = (data: ThemeData): string[] => {
 /**
  * Extracts and parses the font family names from the LayoutData.
  */
-export const extractFontFamiliesFromLayout = (data: unknown): string[] => {
+export const extractFontFamiliesFromLayout = (
+  data: Record<string, any> | undefined
+): string[] => {
   const fontFamilies = new Set<string>();
 
-  const visit = (value: unknown) => {
+  const visit = (value: Record<string, any> | undefined) => {
     if (!value || typeof value !== "object") {
       return;
     }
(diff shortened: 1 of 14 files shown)


packages/visual-editor/src/utils/fonts/visualEditorFonts.ts
```diff
@@ -525,17 +526,43 @@ export const extractFontFamiliesFromTheme = (data: ThemeData): string[] => {
...
- * @param data the ThemeData
- * @param availableFonts built-in Google fonts from the repository
- * @param customFonts custom fonts from FontAPIServer
- * @return An object containing two font registries: inUseGoogleFonts and inUseCustomFonts.
+ * Extracts and parses the font family names from the LayoutData.
  */
-export const filterInUseFontRegistries = (
-  data: ThemeData,
-  availableFonts: FontRegistry,
+export const extractFontFamiliesFromLayout = (data: unknown): string[] => {
```
mkilpatrick: Record<string, any>?

packages/visual-editor/src/utils/fonts/visualEditorFonts.ts
```diff
@@ -175,13 +175,14 @@ export const googleFontLinkTags = fontLinkDataToHTML(
  * they can be appended directly into a document.
  */
 export const createFontLinkElements = (
+  document: Document,
```
mkilpatrick, line 178: Is this used?

benlife5: It's called in updateFontLinksInDocument, which is called for both the Puck window and the Puck preview iframe window. I was able to clean up another function that was passing it along though

mkilpatrick: Oh wow I definitely read this as the streamDocument.

mkilpatrick (approved):

asanehisa (approved):

PR #1244, opened 2026-06-25
feat: hide entity field toggles

asanehisa:
https://yext.slack.com/archives/C0ANYA9ME0Z/p1782413867542499


benlife5 (dismissed):

packages/visual-editor/src/internal/puck/components/LayoutHeader.tsx
mkilpatrick, line 233: Per the bots, move the separator into the localDev check now

asanehisa: Ah I hid it with EntityFieldsToggle so we don't forget it as only RBC is really using those local dev buttons

benlife5 (approved):
Strange that the tests are taking forever

mkilpatrick (approved):

PR #1246, opened 2026-06-29
feat: add color helpers for generated templates

jwartofsky-yext:
Adds helpers for resolving colors, hex colors, and contrast colors to be used by generated templates

Has helped to eliminate over 1000 lines from the yext-quick-service template

Robot guidance for new templates/components:
```
For generated templates/components, the main new things to use are the shared color/default-contrast helpers and the now-standardized CTA default behavior.
They should look to use:

getDefaultForegroundColor(surfaceColor, streamDocument?)
Use this when a component needs the semantic default text/icon color for a surface.
It follows this pattern:
light surface => black
dark surface => white


getSurfaceColorStyle(surfaceColor, streamDocument?)
Use this when a component owns a surface and wants both:backgroundColor
readable default foreground color

Good for wrappers, cards, banners, panels, etc.

It can be used to simplify this:
const backgroundColor = getThemeColorCssValue(surfaceColor);
const foregroundColor = getThemeColorCssValue(
  getDefaultForegroundColor(surfaceColor, streamDocument)
);

const style = {
  ...(backgroundColor ? { backgroundColor } : {}),
  ...(foregroundColor ? { color: foregroundColor } : {}),
};

into this:
const style = getSurfaceColorStyle(surfaceColor, streamDocument);

getThemeColorCssValue(color)
Use this to turn a ThemeColor or token/custom value into a CSS color value.

getThemeColorHexValue(color, streamDocument?)
Use this when the component needs a concrete hex for runtime decisions like contrast checks.

normalizeThemeColorToken(color)
Use this when the component needs to know whether a color is truly explicit or still "default".

isDarkColor(color, streamDocument?)
Use this for dark/light surface decisions when that’s the only thing the component needs to know.

For CTAs specifically:
External/generated components using [`CTA`](/Users/jwartofsky/pages/visual-editor/packages/visual-editor/src/components/atoms/cta.tsx) should usually stop re-implementing default CTA contrast logic.
The built-in CTA runtime now handles the common default paths:Outline/default CTA color adapts from the surface
Link/default CTA color adapts from the surface
explicit CTA colors still work
dark-surface outline explicit colors still go through the existing safeguard

For image fields/config:
aspectRatioOptions
ThemeOptions.ASPECT_RATIO
So the practical guidance for generated code is:
If you are building a custom surface: use getSurfaceColorStyle(...)
If you are building custom text/icons inside a surface: use getDefaultForegroundColor(...)
If you are rendering CTAs: prefer the shared CTA component and let it handle default contrast
If you need reusable aspect-ratio selectors: use aspectRatioOptions or ThemeOptions.ASPECT_RATIO
```


packages/visual-editor/src/components/atoms/cta.tsx
```diff
@@ -270,6 +271,11 @@ export const CTA = (props: CTAProps) => {
     background,
   } = resolvedProps;
   const isDarkBackground = background?.isDarkColor;
+  const hasExplicitColor = !!normalizeThemeColorToken(color);
```
mkilpatrick, line 274: Do the changes here affect the build-in components in a negative way or it'll be the same?

packages/visual-editor/src/utils/colors.ts
```diff
@@ -371,6 +388,7 @@ const getDerivedPaletteHexColor = (
   const baseColorHex = normalizeHexColor(
     getThemeValue(`--colors-${baseColorToken}`, streamDocument)
   );
+  console.log("base color hex is", baseColorHex);
```
mkilpatrick: rm

packages/visual-editor/src/components/atoms/cta.tsx
jwartofsky-yext: They appear the same

The only difference is they use some shared logic now

I had codex do an analysis to confirm

Have also tested with the fake starter before/after to confirm no change in behavior

Will verify with the screenshot tests aswell

benlife5 (commented):
Generally LGTM! Could you explain `getSurfaceColorStyle` a little more or show an example in the PR description?

jwartofsky-yext:
> Generally LGTM! Could you explain `getSurfaceColorStyle` a little more or show an example in the PR description?

Added an abbreviated example in the description. Here's the larger example:

getSurfaceColorStyle(surfaceColor, streamDocument?) is a convenience helper for components that “own a surface” and want both:
the surface background color
the matching default foreground color for content inside that surface

It returns an inline style object like:
{
  backgroundColor: ...,
  color: ...
}

using the shared Visual Editor defaults:
light surface => foreground becomes palette-primary-dark
dark surface => foreground becomes white

So instead of doing this manually:
const backgroundColor = getThemeColorCssValue(surfaceColor);
const foregroundColor = getThemeColorCssValue(
  getDefaultForegroundColor(surfaceColor, streamDocument)
);

const style = {
  ...(backgroundColor ? { backgroundColor } : {}),
  ...(foregroundColor ? { color: foregroundColor } : {}),
};
you can do:
const style = getSurfaceColorStyle(surfaceColor, streamDocument);


A good example use case:
a generated custom card/banner/panel component that takes a ThemeColor background and wants text/icons/links inside to inherit the correct default foreground automatically.

Example:
import {
  getSurfaceColorStyle,
  type ThemeColor,
  type StreamDocument,
} from "@yext/visual-editor";

type PromoCardProps = {
  backgroundColor?: ThemeColor;
  streamDocument: StreamDocument;
  title: string;
  description: string;
};
```
export const PromoCard = ({
  backgroundColor,
  streamDocument,
  title,
  description,
}: PromoCardProps) => {
  return (
    <section
      style={getSurfaceColorStyle(backgroundColor, streamDocument)}
      className="rounded-image-borderRadius p-6"
    >
      <h3>{title}</h3>
      <p>{description}</p>
    </section>
  );
};
```

Why that helps:
the card background is applied
the card’s default text color is applied
any children that inherit color will follow the right surface contrast
generated templates don’t need to re-implement “if dark use white, if light use primary-dark”

It’s most useful when:
the component owns a background
the component wants descendants to inherit the default text color
you want a single standard VE surface/default-foreground behavior


It’s less useful when:

you only need the foreground decision, not the backgrounduse getDefaultForegroundColor(...)

you only need one resolved color valueuse getThemeColorCssValue(...) directly

benlife5:
ok that makes sense, thanks.
Only other question is regarding the foreground using palette-primary-dark -- that's for text & icons? I think we might want to go with black for our AI templates based on [this message](https://yext.slack.com/archives/C0ANYA9ME0Z/p1781721561898329)

packages/visual-editor/src/components/atoms/cta.test.tsx
asanehisa, line 1: curious if we think this is helpful to have? We have CTAs tested in Grid Screenshot tests already

packages/visual-editor/src/utils/colors.ts
```diff
@@ -422,6 +446,50 @@ export const getThemeColorHexValue = (
...
+  if (!normalizedSurfaceColor) {
+    return undefined;
+  }
+
+  const luminance = getColorLuminance(surfaceColor, streamDocument);
+  if (luminance !== undefined) {
+    return prefersWhiteForeground(luminance)
+      ? { selectedColor: "white", contrastingColor: "black" }
+      : {
+          selectedColor: "palette-primary-dark",
```
asanehisa: agree with ben's comment on white / black

packages/visual-editor/src/components/atoms/cta.test.tsx
jwartofsky-yext: The screenshot tests are currently having some issues, and I think having these contrast tests is helpful, so I'm  OK with leaving them in for now

benlife5 (approved):

asanehisa (approved):

PR #1247, opened 2026-06-29
fix: infer image, cta, and rtf types in local-editor

benlife5:
We infer the types of entity document fields in the local-editor, but RTF, Image, and CTA fields were just being picked up as generic objects. This means that they wouldn't show up in the correct dropdowns (particularly for createItemSource). This checks for their subfields and narrows the type if there's a match.

Also removes the `c_` prefix when generating the display name.

### DIFF
--- a/packages/visual-editor/src/vite-plugin/local-editor/entityFields.ts
+++ b/packages/visual-editor/src/vite-plugin/local-editor/entityFields.ts
@@ -18,7 +18,7 @@ type MutableFieldNode = {
  * data.
  */
 export const inferEntityFields = (
-  document: Record<string, unknown>,
+  document: Record<string, unknown> | Record<string, unknown>[],
   stream?: LocalEditorStreamDefinition
 ): {
   fields: YextSchemaField[];
@@ -27,12 +27,14 @@ export const inferEntityFields = (
   const configuredDisplayNames = buildConfiguredDisplayNames(stream);
   const rootNodes = new Map<string, MutableFieldNode>();
 
-  for (const [key, value] of Object.entries(document)) {
-    if (shouldSkipField(key)) {
-      continue;
-    }
+  for (const documentEntry of Array.isArray(document) ? document : [document]) {
+    for (const [key, value] of Object.entries(documentEntry)) {
+      if (shouldSkipField(key)) {
+        continue;
+      }
 
-    mergeFieldNode(rootNodes, inferFieldNode(key, value));
+      mergeFieldNode(rootNodes, inferFieldNode(key, value));
+    }
   }
 
   const fields = Array.from(rootNodes.values())
@@ -82,6 +84,15 @@ const inferFieldNode = (name: string, value: unknown): MutableFieldNode => {
   }
 
   if (isPlainObject(value)) {
+    const inferredStructuredType = inferStructuredObjectType(value);
+    if (inferredStructuredType) {
+      return {
+        name,
+        typeRegistryId: inferredStructuredType,
+        children: new Map(),
+      };
+    }
+
     return {
       name,
       typeRegistryId: "type.object",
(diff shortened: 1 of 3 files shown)


packages/visual-editor/src/vite-plugin/local-editor/entityFields.ts
```diff
@@ -189,6 +200,69 @@ const shouldSkipField = (fieldName: string): boolean => {
...
+    typeof value.url === "string" &&
+    (typeof value.alternateText === "string" ||
+      typeof value.width === "number" ||
+      typeof value.height === "number")
+  );
+};
+
+const isCtaValue = (value: Record<string, unknown>): boolean => {
+  return (
+    typeof value.label === "string" &&
```
mkilpatrick: This is looking at the document and not the schema, correct? Idk if all 3 fields are required so if someone doesn't fill out all 3 in KG this would return false even though it's still a cta.

packages/visual-editor/src/vite-plugin/local-editor/entityFields.ts
```diff
@@ -189,6 +200,69 @@ const shouldSkipField = (fieldName: string): boolean => {
...
+  if (isImageValue(value)) {
+    return "type.image";
+  }
+  if (isCtaValue(value)) {
+    return "type.cta";
+  }
+  return undefined;
+};
+
+const isRichTextValue = (value: Record<string, unknown>): boolean => {
```
asanehisa: don't we have a isRichText already?

packages/visual-editor/src/vite-plugin/local-editor/entityFields.ts
benlife5: Yeah, it doesn't currently pull the schema so we just guess from the document shape. I think I can remove linkType, but just checking for only label or link would be too wide

packages/visual-editor/src/vite-plugin/local-editor/entityFields.ts
benlife5: Oh good call

packages/visual-editor/src/vite-plugin/local-editor/entityFields.ts
```diff
@@ -256,11 +231,7 @@ const isImageValue = (value: Record<string, unknown>): boolean => {
 };
 
 const isCtaValue = (value: Record<string, unknown>): boolean => {
-  return (
-    typeof value.label === "string" &&
-    typeof value.link === "string" &&
-    typeof value.linkType === "string"
-  );
+  return typeof value.label === "string" && typeof value.link === "string";
```
mkilpatrick: linkType is actually not unselectable in KG so there's always a value for it

packages/visual-editor/src/vite-plugin/local-editor/entityFields.ts
```diff
@@ -231,7 +231,11 @@ const isImageValue = (value: Record<string, unknown>): boolean => {
 };
 
 const isCtaValue = (value: Record<string, unknown>): boolean => {
-  return typeof value.label === "string" && typeof value.link === "string";
+  return (
```
mkilpatrick, line 234: What happens when this returns false? Anything detrimental?

Would it be better to do linkType && (label || link)? Or even JUST linkType?

mkilpatrick (dismissed):

packages/visual-editor/src/vite-plugin/local-editor/entityFields.ts
benlife5: ooh yeah `linkType && (label || link)` is probably good. If it doesn't match, then it doesn't show up in type.cta dropdowns... We might eventually need to pull the full schema so that the local-editor is more functional with all data types, but this will get us the main types for now

asanehisa (approved):

mkilpatrick (approved):

briantstephan (approved):

PR #1251, opened 2026-06-29
fix: locator mobile test

k-gerner:
The test was hanging due to `searchLoading` being `undefined` and constantly re-triggering the `useEffect`. Added explicit false check to allow early exit

TEST=auto


jwartofsky-yext (dismissed):

packages/visual-editor/src/components/locator/LocatorWrapper.tsx
```diff
@@ -723,7 +723,7 @@ const LocatorInternal = ({
   }, [currentOffset, isMobile, searchActions, searchLoading]);
 
   React.useEffect(() => {
-    if (!isMobile || searchLoading) {
+    if (!isMobile || searchLoading !== false) {
```
mkilpatrick, line 732: It's used on 714 too. Should the setting of searchLoading set to false instead of undefined instead?

k-gerner: My understanding is that `undefined` for this field means we haven't searched yet. Updated the usage on 714

benlife5 (approved):

asanehisa (approved):

mkilpatrick (approved):

PR #1252, opened 2026-06-30
fix: local editor rtf editor improvements

benlife5:
I tried some more dialogs/editors but it always got more complicated than it was worth. Settled on the small improvements of loading the existing value and not clearing the field on Cancel.

https://github.com/user-attachments/assets/9c0c1280-179f-467d-af6a-15dfb0fc62ee


asanehisa (approved):

briantstephan (approved):
Huge improvement!

PR #1258, opened 2026-07-07
fix: favicons tags

benlife5:
1. We were injecting an empty favicon tag before the real favicon tag
2. The type was for .ico, but the system sends pngs

This seems to have caused Google to not display the favicon in Google results for some (but not all, not sure why) sites.


asanehisa (approved):

jhelenek (approved):
👍

briantstephan (approved):

PR #1260, opened 2026-07-07
feat: store local-editor api keys in local storage

briantstephan:
This stores the mapbox and nearby locations keys in local storage when using the local-editor so we don't need to re-enter them on reload.

Tested manually and confirmed that the keys persisted after restarting `npm run dev`.


asanehisa (approved):
nice!

jwartofsky-yext (approved):

PR #1261, opened 2026-07-07
fix: loosen item source mapping requirements for CTAs

jwartofsky-yext:
Allows itemSource to select a parent field without CTAs, even if the cards require a CTA, provided that there is a text field available.

This allows a user to user a constant value to map to the CTA instead of requiring a CTA field.

Impetus: https://yext.slack.com/archives/C02UVSE7P6W/p1783433102868479

This change would allow them to use their c_productReference linked field without having to add CTAs to the Products and link them in the pageset.

They can then set the CTAs to constant value and use text to populate the fields.


jwartofsky-yext:
I think just adding a tooltip is enough

This seems excessive IMO

PR #1262, opened 2026-07-09
feat: support analytics for custom code section

jhelenek:
This PR supports analytics for custom code section. It adds the Analytics track method to the window so that inner-components are able to reference it. Each analytics is registered by componentId so multiple custom code components can coexist without overwriting each other's handlers. 

TEST=manual
Created a dev-release with this change and set it up with a site. For that site's layout, create 3 separate custom code components, each with different analytic calls. Deployed changes and saw analytic events come in both on the network tab and in Snowflake.

Snowflake query:
`select * from prod_analytics2.public.analytics_events where original_business_id = 4146665 order by event_timestamp desc;`

Example Javascript used in custom code component:
```
function trackCustomClick(button) {
  button.textContent = "Clickedddddd";

  window.YextCustomCodeAnalytics.track({
    scope: "customCodeComponent",
    eventName: "Custom Button Clicked",
    action: "LINK",
    buttonText: "Track Click",
  });
}
```

Video: 
https://drive.google.com/file/d/1SVUUV4njl_fKxF-ZfO3aDrCqPrm3PgAx/view?usp=drive_link

### DIFF
--- a/packages/visual-editor/src/components/customCode/CustomCodeSection.tsx
+++ b/packages/visual-editor/src/components/customCode/CustomCodeSection.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { CodeXml } from "lucide-react";
-import { AnalyticsScopeProvider, useAnalytics } from "@yext/pages-components";
+import { AnalyticsScopeProvider } from "@yext/pages-components";
 import { VisibilityWrapper } from "../atoms/visibilityWrapper.tsx";
 import { msg, pt } from "../../utils/i18n/platform.ts";
 import { useDocument } from "../../hooks/useDocument.tsx";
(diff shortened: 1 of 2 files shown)


packages/visual-editor/src/components/customCode/CustomCodeSection.tsx
```diff
@@ -9,6 +9,77 @@ import { resolveEmbeddedFieldsInString } from "../../utils/resolveYextEntityFiel
 import { processHandlebarsTemplate } from "./customCodeHandlebars.ts";
 import { YextComponentConfig, YextFields } from "../../fields/fields.ts";
 
+type AnalyticsTrackProps = Parameters<
```
asanehisa: I think `export type TrackProps` exists in pages-components if that is easier to use?

packages/visual-editor/src/components/customCode/CustomCodeSection.tsx
```diff
@@ -139,6 +256,12 @@ const CustomCodeSectionWrapper = ({
         ref={containerRef}
         dangerouslySetInnerHTML={{ __html: processedHtml }}
       />
+      <CustomCodeAnalyticsBridgeAndScriptRunner
```
mkilpatrick: Does it not work to simply add track to the window like this?

```
const pagesAnalytics = useAnalytics();
  useEffect(() => {
    (window as any).YextCustomCodeAnalytics.track = (request) => {
      pagesAnalytics?.track(request);
    };

    return () => {
      delete (window as any).YextCustomCodeAnalytics.track;
    };
  }, [pagesAnalytics]);
```

jhelenek: Wouldn't window.YextCustomCodeAnalytics.track get overwritten if we had multiple custom code components?

mkilpatrick: True. To avoid creating a separate instance per component you can probably add it in VisualEditorProvider which wraps everything and is used in every template.

jhelenek: if we move `const pagesAnalytics = useAnalytics();` into `VisualEditorProvider`, custom-code events would not automatically be tagged with the `customCodeSection` scope, right?

Example tree:
```
<AnalyticsProvider>
  <VisualEditorProvider>
    <AnalyticsScopeProvider name="customCodeSection">
      <CustomCodeSection />
    </AnalyticsScopeProvider>
  </VisualEditorProvider>
</AnalyticsProvider>
```

mkilpatrick: You can add <AnalyticsScopeProvider name="customCodeSection"> inside the CC component. I guess one issue is that they won't be uniquely named. Alternatively, scope could be a param to the exposed track call and we'd need to add it probably in the real call to track.

packages/visual-editor/src/components/customCode/CustomCodeSection.tsx
```diff
@@ -89,7 +160,72 @@ const EmptyCustomCodeSection = () => {
...
+{
+  const yextAnalytics = {
+    track: (eventName, data) =>
+      window.YextCustomCodeAnalytics.track(${JSON.stringify(componentId)}, eventName, data),
+  };
+
+${processedJavascript}
+}
+`;
+    containerRef.current.appendChild(script);
```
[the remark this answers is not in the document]
jhelenek: I believe this was a valid comment, but no longer applicable since implementation changed

packages/visual-editor/src/components/customCode/CustomCodeSection.tsx
jhelenek: I also saw that in pages-components, but when I asked the robots (since I'm unfamiliar with the code) they said it's not exported from the package root so I wasn't able to import it

packages/visual-editor/src/components/customCode/CustomCodeSection.tsx
jhelenek: Done!

packages/visual-editor/src/utils/VisualEditorProvider.tsx
```diff
@@ -30,12 +31,32 @@ type EditorProps = {
...
+const useBrowserLayoutEffect =
+  typeof window === "undefined" ? React.useEffect : React.useLayoutEffect;
+
 const VisualEditorProvider = <T extends Record<string, any>>({
   templateProps,
   entityFields,
   tailwindConfig,
   children,
 }: VisualEditorProviderProps<T>) => {
+  const pagesAnalytics = useAnalytics();
```
mkilpatrick, line 59: add empty line after

mkilpatrick (commented):
I was thinking how to make it easy to use but there's nothing that's going to work since it's raw JS (no TS) anyways. We'll have to rely on documentation.

packages/visual-editor/src/utils/VisualEditorProvider.tsx
jhelenek: Done!

asanehisa (approved):

mkilpatrick (approved):

PR #1263, opened 2026-07-09
fix: use Radix popover for ColorSelector

briantstephan:
This updates the ColorSelector to use the shared Radix popover instead of a custom solution. It fixes an issue where, if the color field was the last field in the props menu, the color picker would open below the visible area of the screen. Now it opens below the color preview if there is room, or above it if necessary.

This screenshot shows the color picker opening above the color preview because the field is at the bottom of the page:
Screenshot 2026-07-09 at 6 20 15 PM


asanehisa (approved):
nice!

mkilpatrick (approved):

PR #1265, opened 2026-07-15
chore: cleanup CTA contrast logic

jwartofsky-yext:
Cleans up the color contrast logic in CTAs so that explicit colors are not overwritten.

"Default" values will stil change to match the background, but if the user sets a CTA to an explicit color, that color will be respected.


packages/visual-editor/src/components/atoms/cta.test.tsx
```diff
@@ -61,10 +60,9 @@ describe("CTA default contrast", () => {
       }
     );
 
-    expect(screen.getByRole("button", { name: "Learn more" })).toHaveStyle({
```
asanehisa, line 64: any reason to change these?

packages/visual-editor/src/components/testing/screenshots/Grid/[desktop] version 74 - CTA Group custom colors on dark background.png
Comment (asanehisa) [inline packages/visual-editor/src/components/testing/screenshots/Grid/[desktop] version 74 - CTA Group custom colors on dark background.png:1]: these seem to have regressed

packages/visual-editor/src/components/atoms/cta.test.tsx
jwartofsky-yext: toHaveStyle comes from @testing-library/jest-dom which is not actually installed in the repo

packages/visual-editor/src/components/testing/screenshots/Grid/[desktop] version 74 - CTA Group custom colors on dark background.png
Comment (jwartofsky-yext) [reply packages/visual-editor/src/components/testing/screenshots/Grid/[desktop] version 74 - CTA Group custom colors on dark background.png:1]: that is because this third CTA is explicitly set to a color that does not contrast, so this case is working as expected

I have changed it to use a "default" value so that it will adapt for this test

mkilpatrick (approved):

packages/visual-editor/src/components/testing/screenshots/Grid/[desktop] version 74 - CTA Group custom colors on dark background.png
Comment (asanehisa) [reply packages/visual-editor/src/components/testing/screenshots/Grid/[desktop] version 74 - CTA Group custom colors on dark background.png:1]: I thought explicit colors aren't overwritten? it seems like it was overwritten by becoming white

Comment (jwartofsky-yext) [reply packages/visual-editor/src/components/testing/screenshots/Grid/[desktop] version 74 - CTA Group custom colors on dark background.png:1]: Talked offline

Explicit colors are now no longer overridden

They were overridden before, and the test was using that behavior. That is why I removed the "color" values from the test case, so it will use the fallback logic properly.

PR #1268, opened 2026-07-16
feat: use breadcrumbPrefix as slug for DM Root entities 

asanehisa:
Was able to test with some srv deployed [changes](https://gerrit.yext.com/c/alpha/+/340824) in us3dev and confirm it'll work. 

If there is a breadcrumb prefix, for DM Root Breadcrumbs, use that instead of `<prefix>/slug`


packages/visual-editor/src/utils/urls/resolveBreadcrumbsFromPathInfo.ts
```diff
@@ -47,9 +47,14 @@ export const resolveBreadcrumbsFromPathInfo = (
       continue;
     }
 
-    const normalizedSlug = normalizedPrefix
-      ? `${normalizedPrefix}/${directoryLevelSlug}`
-      : directoryLevelSlug;
+    // The breadcrumb prefix already represents the root directory, so avoid
+    // appending the root's own slug to it.
+    const isRootDirectory = parent.meta?.entityType?.id === "dm_root";
+    const normalizedSlug =
```
mkilpatrick: Not a fan of multi level ternaries

packages/visual-editor/src/utils/urls/resolveBreadcrumbsFromPathInfo.ts
```diff
@@ -47,9 +47,14 @@ export const resolveBreadcrumbsFromPathInfo = (
       continue;
     }
 
-    const normalizedSlug = normalizedPrefix
-      ? `${normalizedPrefix}/${directoryLevelSlug}`
-      : directoryLevelSlug;
+    // The breadcrumb prefix already represents the root directory, so avoid
```
mkilpatrick, line 52: At this point we should just do it all in VE instead of half and half and require version checking. That means you need to update resolveUrlFromPathInfo, but unfortunately I don't see pageset info about the dir coming through. The only thing now that could maybe work is document.meta.entityType.id == 'dm_root'. I'm surprised the page group dir config isn't already set, but that could be something we pass to key off of that instead.

asanehisa: We'll still need half and half with version checking as the "View Live Page" button is populated in alpha in gocode/src/yext/publish/yextsites/storm/js/visualeditor/utils/resolveUrlTemplateForEntity.ts

mkilpatrick: So that means still gating on the version?

asanehisa: Yes. But I can either do this current PR with https://gerrit.yext.com/c/alpha/+/340824's changes or change `resolveUrlFromPathInfo` in visual-editor as well as `gocode/src/yext/publish/yextsites/storm/js/visualeditor/utils/resolveUrlTemplateForEntity.ts` in alpha

Plus gating in either option

mkilpatrick: I guess there's no ideal solution then. I'll leave it to you to decide what you think is best. What I care about most is trying to keep any specialized logic in one place as much as possible instead of having half in VE and half in alpha, and the two need to align. I know that's not 100% possible because of how it is and we need to gate, but think about a future where everyone is on the latest or newer version that would allow to clean up some legacy code. At that point can it all be contained to one place?

asanehisa: I think Option 1 does feel the "best" to me in terms of use editing the `urlTemplatePrimary` in alpha used in other places. Will be going forward with that! 

This PR won't need any changes

benlife5 (approved):

jwartofsky-yext:
> Was able to test with some srv deployed changes in us3dev and confirm it'll work.

Can you add more details that explain this PR?

If there is a breadcrumb prefix form pathInfo.breadcrumbPrefix, we now use that for the DM slugs?

A jam or just a list of bullets explaining the testing flow would be helpful

mkilpatrick (approved):

PR #1277, opened 2026-07-28
chore: set js-yaml override to avoid vulnerability

briantstephan:
This avoids the following vulnerability:

https://nvd.nist.gov/vuln/detail/CVE-2026-59869

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,8 @@
   "pnpm": {
     "overrides": {
       "happy-dom": "20.8.8",
-      "esbuild": "0.28.1"
+      "esbuild": "0.28.1",
+      "js-yaml": "4.3.0"
     },
     "peerDependencyRules": {
       "allowedVersions": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -60,7 +60,8 @@
   "pnpm": {
     "overrides": {
       "happy-dom": "20.8.8",
-      "esbuild": "0.28.1"
+      "esbuild": "0.28.1",
+      "js-yaml": "4.3.0"
```
asanehisa, line 64: ah do we have to do it via this rather than direct upgrades of other packages?

briantstephan: I think this is the safest way. The dependency is coming from generate-license-file -> cosmicconfig -> js-yaml, but upgrading to the latest version of generate-license-file (4.2.1) gives us cosmiconfig@^9.0.0; the latest cosmiconfig is 9.0.2 permits js-yaml@^4.1.0, which still doesn't guarantee that we get 4.3.0.

benlife5 (approved):

mkilpatrick (approved):

asanehisa (approved):

PR #1280, opened 2026-07-31
feat: add image fill prop

nanhu95:
Add a prop for component images to size
the image to either "fill" or "fit".

https://github.com/user-attachments/assets/48f84ec7-d85d-4547-b81c-a2e501836eaf


packages/visual-editor/src/components/atoms/image.test.tsx
benlife5, line 1: These components were written pre-heavy use of Codex and so we haven't really set a post-Codex testing strategy. I have mixed feelings on whether tests like this are useful -- more coverage is good, but also they tend to be a lot of test code for testing small functionality & are prone to constant changes. So, I'd tend to say that we shouldn't commit tests like this and keep relying on the screenshot tests, but I'm open to other opinions

packages/visual-editor/src/components/migrations/migrationRegistry.ts
benlife5, line 1: Just a note bc this is fairly common - both your PR and my PR has a new migration, so whoever merges second will need to merge main and renumber the migration

packages/visual-editor/src/components/pageSections/PhotoGallerySection/PhotoGalleryWrapper.tsx
```diff
@@ -85,6 +86,12 @@ export interface PhotoGalleryWrapperProps {
   };
 }
 
+const photoGalleryImageStylingFields = {
```
benlife5: nit - just put this object inline on 117

packages/visual-editor/src/types/images.ts
```diff
@@ -1,6 +1,8 @@
 import { ImageType } from "@yext/pages-components";
 import { TranslatableString } from "./types.ts";
 
+export type ImageFillType = "fill" | "fit";
```
benlife5, line 5: Add a brief tsdoc

packages/visual-editor/src/utils/migrate.test.ts
```diff
@@ -7,6 +7,10 @@ import { mainContentWrapperMigration } from "../components/migrations/0073_main_
 import { normalizeFooterLogoImageMigration } from "../components/migrations/0075_normalize_footer_logo_image.ts";
 import { slotMappedCardsMigration } from "../components/migrations/0076_slot_mapped_cards.ts";
 import { removeMapboxApiKeyPropsMigration } from "../components/migrations/0078_remove_mapbox_api_key_props.ts";
+import { imageFillTypeMigration } from "../components/migrations/0079_image_fill_type.ts";
+import { ProductCard } from "../components/pageSections/ProductSection/ProductCard.tsx";
+import { ImageWrapper } from "../components/contentBlocks/image/Image.tsx";
```
benlife5: Don't import the components here -- if there's future changes, they will drift. I don't think migrate actually needs the components for this sort of test

packages/visual-editor/src/components/atoms/image.test.tsx
jwartofsky-yext: Agreed with this

For the other components we generally don't have tests like this

packages/visual-editor/src/components/contentBlocks/image/styling.test.ts
```diff
@@ -0,0 +1,24 @@
+import { describe, expect, it } from "vitest";
```
jwartofsky-yext: I don't think this test is super necessary

Can remove it

packages/visual-editor/src/components/atoms/image.test.tsx
nanhu95: Noted, and thanks for calling it out. Will delete.

packages/visual-editor/src/components/contentBlocks/image/styling.test.ts
nanhu95: Done

benlife5 (approved):

jwartofsky-yext (approved):

PR #1287, opened 2026-08-11
feat: support section library repo structure

briantstephan:
This adds basic support for the new Section Library starter repo structure. This includes some of the following updates:

- Added Section Library discovery and validation from `src/library`
- Added public `SectionConfig` and `MainContent` support for library section configs
- Generated a Puck config with only library sections, not the built-in Visual Editor components
- Generated Pages render/editor templates, including temporary `main` and `edit` aliases to make this work without platform changes
- Generated the Section Library artifact manifest (which is not yet used) and the temporary legacy template manifest

Currently, this only supports a single ENTITY layout (DIRECTORY, LOCATOR, and multiple ENTITY layout support will be added later).

Tested in platform on this site and confirmed it works without any platform changes.
https://www.yext.com/s/4520471/yextsites/168240/branch/142634/deploys/recent


packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,685 @@
...
+    }
+  }
+};
+
+const writeGeneratedFile = (filePath: string, source: string): void => {
+  fs.ensureDirSync(path.dirname(filePath));
+  fs.writeFileSync(filePath, source);
+};
+
+const isGeneratedFile = (filePath: string): boolean => {
```
benlife5: In other parts of the vite plugin, we've tracked which file paths are generated. This seems more prone to break, but also would allow you to override it if you wanted... so I'm good either way

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,685 @@
...
+  if (!value) {
+    return {};
+  }
+  if (!ts.isStringLiteral(value) || !value.text) {
+    throw new Error(`${sourcePath} config ${name} must be a non-empty string`);
+  }
+  return { category: value.text };
+};
+
+const isSafeId = (value: string): boolean => /^[A-Za-z0-9_-]+$/.test(value);
```
benlife5, line 472: tsdoc for what this is testing for

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,685 @@
...
+          },
+        ],
+      },
+      null,
+      2
+    )
+  );
+};
+
+const buildConfigSource = (rootDir: string, sections: Section[]): string => {
```
benlife5, line 358: Nit preference, but I think it might be nice to store the config/template as separate files with placeholder strings and then read the files in here, replace the strings, and write the result. A little easier to follow/edit

benlife5 (commented):
Generally LGTM

packages/visual-editor/src/vite-plugin/plugin.ts
```diff
@@ -82,9 +90,13 @@ const virtualFiles: VirtualFile[] = [
   },
 ];
 
-export const yextVisualEditorPlugin = (): Plugin => {
+export const yextVisualEditorPlugin = (
+  options: VisualEditorPluginOptions = {}
+): Plugin => {
   let isBuildMode = false;
   const filesToCleanup: string[] = [];
+  let sectionLibraryFiles: string[] = [];
```
asanehisa, line 98: nit: can be `const sectionLibraryFiles: string[] = [];` ?

packages/visual-editor/src/sectionLibrary.ts
```diff
@@ -0,0 +1,7 @@
+/** Metadata that describes a user-selectable Section Library section. */
+export type SectionConfig = {
+  displayName: string;
+  description: string;
+  pageSetTypes: "ENTITY"[];
```
asanehisa: is there a reason we are only allowing "ENTITY"? would doing

```
export type PageSetType = "ENTITY" | "DIRECTORY" | "LOCATOR";

export type SectionConfig = {
  displayName: string;
  description: string;
  pageSetTypes: PageSetType[];
  category?: string;
};
```
be better?

packages/visual-editor/src/vite-plugin/plugin.ts
```diff
@@ -148,12 +162,35 @@ export const yextVisualEditorPlugin = (): Plugin => {
...
     },
     buildStart() {
+      if (options.sectionLibrary) {
+        const generatedLibrary = generateSectionLibraryFiles(process.cwd());
+        sectionLibraryFiles = generatedLibrary.generatedFiles;
+        sectionLibraryManifest = generatedLibrary.manifestSource;
+        return;
+      }
       generateFiles();
     },
```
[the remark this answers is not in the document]
briantstephan: This is what we want, at least for now. We want things to build properly when no library is added to the repo in case we remove the existing test library.

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,685 @@
...
+    configPath,
+    path.join(rootDir, "src", "templates", `${layoutId}.tsx`),
+    path.join(rootDir, "src", "templates", "main.tsx"),
+    path.join(rootDir, "src", "templates", `edit-${layoutId}.tsx`),
+    path.join(rootDir, "src", "templates", "edit.tsx"),
+  ];
+
+  writeGeneratedFile(configPath, buildConfigSource(rootDir, library.sections));
+  writeGeneratedFile(templatePaths[1], buildRenderTemplateSource(layoutId));
+  writeGeneratedFile(templatePaths[2], buildRenderTemplateSource(layoutId));
```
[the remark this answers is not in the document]
briantstephan: `main` and the real layout ID are compatibility aliases for the same page type (`ENTITY`), so they should resolve to the same URL while the existing platform code selects `main`, and the updated platform code in the future will select the real layout ID.

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,685 @@
...
+  id: string;
+  displayName: string;
+  description: string;
+  category?: string;
+  sourcePath: string;
+};
+
+type Layout = {
+  metadata: LayoutMetadata;
+  defaultLayout: unknown;
```
mkilpatrick: `Record<string, any>`?

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,685 @@
...
+  description: string;
+};
+
+type LayoutMetadata = {
+  id: string;
+  displayName: string;
+  previewImageUrl: string;
+  vertical?: string;
+  purpose?: string;
+  pageSetType: "ENTITY";
```
mkilpatrick: We should have a type for the different ones

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,685 @@
...
+  id: string;
+  displayName: string;
+  description: string;
+};
+
+type LayoutMetadata = {
+  id: string;
+  displayName: string;
+  previewImageUrl: string;
+  vertical?: string;
```
mkilpatrick: type the vertical/purpose, also should be lists right?

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,685 @@
...
+  purpose?: string;
+  pageSetType: "ENTITY";
+};
+
+type Section = {
+  id: string;
+  displayName: string;
+  description: string;
+  category?: string;
```
mkilpatrick: What's this represent? The Puck left bar section it's put in? Document it.

packages/visual-editor/src/vite-plugin/plugin.ts
```diff
@@ -129,6 +141,8 @@ export const yextVisualEditorPlugin = (): Plugin => {
     filesToCleanup.forEach((filePath) => {
       fs.rmSync(filePath, { force: true });
     });
+    cleanupGeneratedSectionLibraryFiles(sectionLibraryFiles);
```
mkilpatrick: Shouldn't this be guarded by the option?

packages/visual-editor/src/vite-plugin/plugin.ts
```diff
@@ -148,12 +162,35 @@ export const yextVisualEditorPlugin = (): Plugin => {
...
+      if (options.sectionLibrary) {
+        const generatedLibrary = generateSectionLibraryFiles(process.cwd());
+        sectionLibraryFiles = generatedLibrary.generatedFiles;
+        sectionLibraryManifest = generatedLibrary.manifestSource;
+        return;
+      }
       generateFiles();
     },
+    generateBundle() {
+      if (isBuildMode && sectionLibraryManifest) {
```
mkilpatrick: && options.sectionLibrary?

packages/visual-editor/src/vite-plugin/plugin.ts
```diff
@@ -148,12 +162,35 @@ export const yextVisualEditorPlugin = (): Plugin => {
...
+        });
+      }
+    },
     buildEnd() {
-      if (isBuildMode) {
+      if (isBuildMode && !options.sectionLibrary) {
         cleanupFiles();
       }
     },
+    closeBundle: {
```
mkilpatrick: Can't buildEnd just handle both if you condition inside cleanupFiles?

packages/visual-editor/src/sectionLibrary.ts
mkilpatrick: Yes - and include an ALL option

packages/visual-editor/src/vite-plugin/plugin.ts
briantstephan: I don't think so; the plugin updates them during `buildStart` and resets the generated file paths during cleanup.

packages/visual-editor/src/sectionLibrary.ts
briantstephan: Might as well update to that now, yeah. I had just `ENTITY` before since directory and locator aren't implemented yet but it does seem a little more confusing leaving them out when we know we're going to have them later.

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Good call, updated

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: That's valid, there are benefits to both. To maintain consistency I'll update this to use the file path tracking approach.

packages/visual-editor/package.json
```diff
@@ -31,7 +32,7 @@
   "types": "dist/index.d.ts",
   "scripts": {
     "build:components": "tsc && vite build",
-    "build:plugin": "vite build --config vite.config.plugin.ts",
+    "build:plugin": "tsc -p tsconfig.plugin.json --emitDeclarationOnly && vite build --config vite.config.plugin.ts",
```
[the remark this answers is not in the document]
mkilpatrick: Why was this change needed?

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,685 @@
...
+    if (!activeFiles.has(filePath) && isGeneratedFile(filePath)) {
+      fs.removeSync(filePath);
+    }
+  }
+};
+
+const writeGeneratedFile = (filePath: string, source: string): void => {
+  fs.ensureDirSync(path.dirname(filePath));
+  fs.writeFileSync(filePath, source);
+};
```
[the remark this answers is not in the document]
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.test.ts
```diff
@@ -0,0 +1,249 @@
...
+    fs.writeFileSync(handWrittenTemplate, "export default null;");
+
+    cleanupGeneratedSectionLibraryFiles(generatedFiles);
+
+    expect(fs.existsSync(handWrittenTemplate)).toBe(true);
+    expect(fs.existsSync(generatedFiles[0])).toBe(false);
+    expect(fs.existsSync(path.join(rootDir, ".template-manifest.json"))).toBe(
+      true
+    );
+  });
```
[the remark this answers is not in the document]
briantstephan: Done

packages/visual-editor/package.json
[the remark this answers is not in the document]
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,685 @@
...
+type GeneratedSectionLibrary = {
+  generatedFiles: string[];
+  manifestSource?: string;
+};
+
+/**
+ * Reads one Entity-only Section Library and writes the Pages source that it needs.
+ *
+ * 1. Validate the metadata, sections, and default layout under `src/library`.
+ * 2. Generate the Puck config and render and editor templates.
```
mkilpatrick: commas instead of multiple ands?

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,685 @@
...
+  layout: Layout;
+};
+
+type GeneratedSectionLibrary = {
+  generatedFiles: string[];
+  manifestSource?: string;
+};
+
+/**
+ * Reads one Entity-only Section Library and writes the Pages source that it needs.
```
mkilpatrick: This doesn't really help me understand anything

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,685 @@
...
+      null,
+      2
+    )}\n`,
+  };
+};
+
+/** Reads and validates the Entity-only Section Library source. */
+const readSectionLibrary = (rootDir: string): SectionLibrary | undefined => {
+  const libraryDirectory = path.join(rootDir, "src", "library");
+  const libraryPath = path.join(libraryDirectory, "library.json");
```
mkilpatrick: libraryJsonPath?

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,685 @@
...
+  if (!fs.existsSync(libraryPath)) {
+    return undefined;
```
mkilpatrick, line 111: I think you can just return

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,685 @@
...
+  const sectionIds = new Set<string>();
+  return fs
+    .readdirSync(sectionsDirectory, { withFileTypes: true })
+    .sort((left, right) => left.name.localeCompare(right.name))
+    .map((entry) => {
+      const sourcePath = path.join(sectionsDirectory, entry.name);
+      if (entry.isDirectory()) {
+        throw new Error(`Section directories are not supported: ${sourcePath}`);
+      }
+      if (!entry.isFile() || path.extname(entry.name) !== ".tsx") {
```
mkilpatrick: jsx should be allowed

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,685 @@
...
+        throw new Error(`Sections must be .tsx files: ${sourcePath}`);
+      }
+
+      const id = path.basename(entry.name, ".tsx");
```
mkilpatrick: I forget, did we decide to get the id from the filename? It's probably better to have the user define it in the frontmatter.

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,685 @@
...
+  }
+  if (
+    !declaration?.initializer ||
+    !ts.isObjectLiteralExpression(declaration.initializer)
+  ) {
+    throw new Error(`${sourcePath} config must be an object literal`);
+  }
+
+  const config = declaration.initializer;
+  const pageSetTypes = getProperty(config, "pageSetTypes");
```
mkilpatrick: I think we should use ts-morph to extract the frontmatter, put it into the Section object, then validate that, rather than validating as it's read (except for things that would cause reading the code to fail).

I also think we should have the frontmatter parsing in its own file to keep separation of concerns clear.

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
mkilpatrick: Agree. Why are we doing it this way even though the current plugin has template files it uses?

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Yeah I think we decided that specifying it in the config was unnecessary since the filenames will always be unique, and it simplifies the config a bit. In what cases would users want to change it?

mkilpatrick: What if someone wanted to rename the file (small refactor) but treat it as the same component (section)?

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Updated them to be lists. When you say to "type" them, are you saying that we should list out what all of the vertical and purpose string options are here?

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Yeah it's the group in the Puck left pane. Added documentation for it.

packages/visual-editor/src/vite-plugin/plugin.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/plugin.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Improved clarity

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Improved clarity

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/plugin.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: That's valid, I guess it's better to give people the option. Updated the section config to include an id field which we use for the layout.

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,598 @@
...
+      const sourcePath = path.join(sectionsDirectory, entry.name);
+      if (entry.isDirectory()) {
+        throw new Error(`Section directories are not supported: ${sourcePath}`);
+      }
+      const extension = path.extname(entry.name);
+      if (!entry.isFile() || ![".tsx", ".jsx"].includes(extension)) {
+        throw new Error(`Sections must be .tsx or .jsx files: ${sourcePath}`);
+      }
+
+      const id = path.basename(entry.name, extension);
```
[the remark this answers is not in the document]
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
```diff
@@ -0,0 +1,83 @@
...
+    const component =
+      sourceFile.getFunction(componentName) ??
+      sourceFile.getVariableDeclaration(componentName);
+    if (!component?.isExported()) {
+      throw new Error(`${sourcePath} must named-export ${componentName}`);
+    }
+
+    const config = sourceFile.getVariableDeclaration("config");
+    if (!config?.isExported()) {
+      throw new Error(`${sourcePath} must named-export config`);
```
mkilpatrick: must export 'config'

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
```diff
@@ -0,0 +1,83 @@
...
+    sourcePath,
+    fs.readFileSync(sourcePath, "utf8"),
+    { overwrite: true }
+  );
+  try {
+    const component =
+      sourceFile.getFunction(componentName) ??
+      sourceFile.getVariableDeclaration(componentName);
+    if (!component?.isExported()) {
+      throw new Error(`${sourcePath} must named-export ${componentName}`);
```
mkilpatrick: must export '$componentName'

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
```diff
@@ -0,0 +1,83 @@
...
+): SectionFrontmatter => {
+  const sourceFile = project.createSourceFile(
+    sourcePath,
+    fs.readFileSync(sourcePath, "utf8"),
+    { overwrite: true }
+  );
+  try {
+    const component =
+      sourceFile.getFunction(componentName) ??
+      sourceFile.getVariableDeclaration(componentName);
```
mkilpatrick: What's the function vs variableDeclaration? Is that just if you export the component directly or if you alias it in a different export call?

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
```diff
@@ -0,0 +1,83 @@
...
+    const config = sourceFile.getVariableDeclaration("config");
+    if (!config?.isExported()) {
+      throw new Error(`${sourcePath} must named-export config`);
+    }
+    const object = config
+      .getInitializer()
+      ?.asKind(SyntaxKind.ObjectLiteralExpression);
+    if (!object) {
+      return {
+        hasSectionConfigType:
```
mkilpatrick: Shouldn't we just return the error if the thing can't be parsed instead of returning validation metadata on the Frontmatter object?

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
```diff
@@ -0,0 +1,83 @@
...
+import { Project, SyntaxKind } from "ts-morph";
+
+export type SectionFrontmatter = {
+  hasSectionConfigType: boolean;
+  isConfigObject: boolean;
+  id?: string | null;
+  displayName?: string | null;
+  description?: string | null;
+  category?: string | null;
+  pageSetTypes?: string[];
```
mkilpatrick: This should be typed

packages/visual-editor/src/sectionLibrary.ts
```diff
@@ -1,7 +1,9 @@
-/** Metadata that describes a user-selectable Section Library section. */
+export type PageSetType = "ENTITY" | "DIRECTORY" | "LOCATOR";
```
mkilpatrick: Missing ALL

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
mkilpatrick: Yea, so they match what can be filtered in the gallery

packages/visual-editor/src/sectionLibrary.ts
briantstephan: I'm not sure it's correct to add `ALL` as a page set type, since it isn't actually the identifier for one of our page set types. If anything, we could do something like the following, but it seems a bit unnecessarily complicated, and I don't think it's inconvenient for the user to just list out all of the types to which a component applies.

```
export type PageSetAvailability = PageSetType | "ALL";
```

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
briantstephan: It was to handle the cases where a component is exported as either a function or variable. I've updated to handle this in a cleaner way now.

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
```diff
@@ -0,0 +1,87 @@
...
+        return element.asKind(SyntaxKind.StringLiteral)?.getLiteralValue();
+      });
+
+    return {
+      id: getString("id"),
+      displayName: getString("displayName"),
+      description: getString("description"),
+      category: getString("category"),
+      ...(pageSetTypes?.every((pageSetType) => pageSetType !== undefined)
+        ? { pageSetTypes: pageSetTypes as PageSetType[] }
```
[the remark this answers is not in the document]
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,607 @@
...
+  const entries = sections.map((section, index) => {
+    return `  { id: ${JSON.stringify(section.id)}, component: Section${index}, config: sectionConfig${index} },`;
+  });
+  return sectionLibraryConfigTemplate
+    .replace("/* SECTION_LIBRARY_GENERATED_FILE */", GENERATED_FILE_PREFIX)
+    .replace("/* SECTION_LIBRARY_IMPORTS */", imports.join("\n"))
+    .replace("/* SECTION_LIBRARY_ENTRIES */", entries.join("\n"));
+};
+
+const buildRenderTemplateSource = (layoutId: string): string => {
```
benlife5, line 389: Can we do this and the editor like section-library-config too?

packages/visual-editor/src/sectionLibrary.ts
mkilpatrick: We'd have to map/treat it as working for any pageset type. The reason for ALL is to support someone making something generic that works for all page group types and future proofs if a new page group type comes into existence and they can still use their component on it without requiring an update to the code. Though maybe that's not great either. I guess we can start with it this way for now.

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,607 @@
...
+  id: string;
+  displayName: string;
+  description: string;
+};
+
+type EntityLayoutMetadata = {
+  id: string;
+  displayName: string;
+  previewImageUrl: string;
+  vertical?: string[];
```
mkilpatrick: Still haven't typed these

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,607 @@
...
+
+type LayoutMetadata =
+  | EntityLayoutMetadata
+  | DirectoryLayoutMetadata
+  | LocatorLayoutMetadata;
+
+/**
+ * A Section Library section and its static config metadata.
+ *
+ * `id` is the stable config ID used in layouts and generated Puck configs.
```
mkilpatrick: You should document the types themselves instead of documenting the child type here. Do it as a comment above each property.

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,607 @@
...
+ * `category` is the sidebar group for the section. When it is omitted, the
+ * section appears in the "Sections" group.
+ * `sourcePath` is the path used in generated imports.
+ */
+type Section = SectionFrontmatter & {
+  componentName: string;
+  sourcePath: string;
+};
+
+type ValidatedSection = Section & {
```
mkilpatrick: Is this to handle the fact that SectionFrontmatter has all optional properties? Can you just not make them optional?

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
```diff
@@ -0,0 +1,106 @@
+import fs from "fs-extra";
+import { Node, Project, SyntaxKind } from "ts-morph";
+import {
+  supportedPageSetTypes,
+  type PageSetType,
+} from "../../sectionLibrary.ts";
+
+export type SectionFrontmatter = {
```
mkilpatrick: Isn't this basically the same as SectionConfig in sectionLibrary.ts?

packages/visual-editor/vite.config.plugin.ts
```diff
@@ -15,7 +24,15 @@ export default defineConfig(() => ({
     target: "node18",
     tsconfig: path.resolve(__dirname, "tsconfig.plugin.json"),
     rollupOptions: {
-      external: ["node:path", "fs-extra"],
+      external: (source) => {
+        return (
+          nodeBuiltins.has(source) ||
+          ["fs-extra", "ts-morph", "typescript"].some(
```
mkilpatrick, line 30: I don't think we can externalize these. That means any users of the plugin in the starter must now import these libs in their package.json.

briantstephan: `fs-extra` is a runtime dependency of the plug-in, so I moved it to Visual Editor’s `dependencies`. I think we should keep these Node-only packages externalized; `npm` will visual-editor's dependencies transitively, so the starter(s) shouldn't need to list them directly.

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
briantstephan: True, I've removed it

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
```diff
@@ -92,13 +85,41 @@ export const readSectionFrontmatter = (
...
-      id: getString("id"),
-      displayName: getString("displayName"),
-      description: getString("description"),
-      category: getString("category"),
-      ...(pageSetTypes ? { pageSetTypes } : {}),
+      id,
+      displayName,
+      description,
+      pageSetTypes,
+      ...(category === undefined ? {} : { category }),
```
mkilpatrick, line 122: I recall reading somewhere that it defaults to "Sections" or something. Should that be handled here instead of wherever that is happening? I assume it's in the template wrapper.

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: I've simplified this a bit

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
```diff
@@ -3,23 +3,16 @@ import { Node, Project, SyntaxKind } from "ts-morph";
...
-  displayName?: string | null;
-  description?: string | null;
-  category?: string | null;
-  pageSetTypes?: PageSetType[];
-};
-
 const project = new Project({ compilerOptions: { allowJs: true } });
 
 /** Reads the static SectionConfig metadata from one section source file. */
 export const readSectionFrontmatter = (
```
mkilpatrick: Maybe extractSectionConfigFrontmatter

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -42,41 +57,28 @@ type LayoutMetadata =
...
- * `displayName` is the name shown to the user in the editor.
- * `description` is used by AI agents when generating layouts using existing components.
- * `category` is the sidebar group for the section. When it is omitted, the
- * section appears in the "Sections" group.
- * `sourcePath` is the path used in generated imports.
- */
-type Section = SectionFrontmatter & {
+/** A Section Library section and its static config metadata. */
+type Section = SectionConfig & {
+  /** The file stem for the required component export. */
```
mkilpatrick, line 21: What's a file stem?

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
briantstephan: I think it's better to handle that in the generated config where grouping occurs, so `SectionConfig.category` stays optional and the parser preserves the developer's provided config metadata.

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: It's the filename without the extension (e.g. `Hero` when the filename is `Hero.tsx`)

packages/visual-editor/src/vite-plugin/section-library/sectionFrontmatter.ts
briantstephan: Done

benlife5 (approved):

mkilpatrick (approved):

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,513 @@
...
+import sectionLibraryConfigTemplate from "../templates/section-library-config.tsx?raw";
+import sectionLibraryEditorTemplate from "../templates/section-library-editor.tsx?raw";
+import sectionLibraryRenderTemplate from "../templates/section-library-render.tsx?raw";
+import type { SectionConfig } from "../../sectionLibrary.ts";
+import { extractSectionConfigFrontmatter } from "./sectionFrontmatter.ts";
+
+const GENERATED_FILE_PREFIX =
+  "/** THIS FILE IS GENERATED BY THE SECTION LIBRARY PLUGIN */";
+
+type LibraryMetadata = {
```
jwartofsky-yext: Should all of our types go in sectionLibrary.ts?

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
```diff
@@ -0,0 +1,513 @@
...
+  schemaVersion: 1;
+  id: string;
+  displayName: string;
+  description: string;
+};
+
+const verticals = [
```
jwartofsky-yext: same with enums

packages/visual-editor/src/vite-plugin/templates/section-library-render.tsx
```diff
@@ -0,0 +1,137 @@
...
+    translations: await injectTranslations(props.document),
+  };
+};
+
+const SectionLibraryLayout: Template<TemplateRenderProps> = (props) => {
+  let data = {};
+  try {
+    data = JSON.parse(props.document.__?.layout ?? "{}");
+  } catch {
+    // Render an empty page when Platform has no layout.
```
jwartofsky-yext: if the data is malformed, there should be an error

packages/visual-editor/src/vite-plugin/templates/section-library-render.tsx
```diff
@@ -0,0 +1,137 @@
...
+      applyCertifiedFacts(data.document),
+    ].join("\n"),
+  };
+};
+
+export const getPath: GetPath<TemplateProps> = ({ document }) => {
+  try {
+    return resolveUrlTemplate(document, "");
+  } catch {
+    return layoutId;
```
jwartofsky-yext: this seems weird. Why is layoutId the path if resolve errors?

briantstephan: Updated

packages/visual-editor/src/vite-plugin/templates/section-library-render.tsx
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Done

packages/visual-editor/src/vite-plugin/section-library/sectionLibraryGenerator.ts
briantstephan: Done

